PR #1, opened 2021-11-04
[ci] add jobs to run the acceptance tests

olksdr:
Adding CI jobs to run the acceptance testing:
* create the default config
* create a simple pipeline to run the tests

### DIFF
--- a/.ci/pipelines/acceptance.groovy
+++ b/.ci/pipelines/acceptance.groovy
@@ -2,12 +2,13 @@
 
 pipeline {
   agent { label('linux && immutable && docker') }
-    environment {
-      HOME = "${env.JENKINS_HOME}" 
-      REPOSITORY = "terraform-provider-elasticstack"
-      GIT_REFERENCE_REPO = "/var/lib/jenkins/.git-references/terraform-provider-elasticstack.git"
-      GIT_CREDENTIALS = "f6c7695a-671e-4f4f-a331-acdce44ff9ba"
-    }
+
+  environment {
+    HOME = "${env.JENKINS_HOME}" 
+    REPOSITORY = "terraform-provider-elasticstack"
+    GIT_REFERENCE_REPO = "/var/lib/jenkins/.git-references/terraform-provider-elasticstack.git"
+    GIT_CREDENTIALS = "f6c7695a-671e-4f4f-a331-acdce44ff9ba"
+  }
   
   stages {
     stage('Checkout') {


.ci/jobs/defaults.yml
```diff
@@ -0,0 +1,39 @@
...
+        name: branch_specifier
+        default: refs/heads/main
+        description: "the Git branch specifier to build (&lt;branchName&gt;, &lt;tagName&gt;,&lt;commitId&gt;, etc.)"
+  properties:
+    - github:
+        url: https://github.com/elastic/terraform-provider-elasticstack
+    - inject:
+        properties-content: HOME=$JENKINS_HOME
+  pipeline-scm:
+    script-path: .ci/pipelines/acceptance.groovy
```
JoshMock: Are you putting this in the defaults because you don't expect to add more jobs to this repo? It makes reading the individual jobs slightly more confusing.

olksdr: So, right now there will be only acceptance testing for `main` branch and PRs and nothing else, that's why this is in the defaults. 
Maybe later we will want to automate the release process, or some kind of signing jobs, then we can come back to this and refactor when we need it to change. 

My thinking was to make this first work and refactor later if/when we have to

.ci/jobs/defaults.yml
```diff
@@ -0,0 +1,39 @@
...
+##### GLOBAL METADATA
+
+- meta:
+  cluster: devops-ci
+
+##### JOB DEFAULTS
+
+- job:
+  project-type: pipeline
+  prune-dead-branches: true
```
jonahbull: I think this is only valid for multibranch pipeline jobs.

.ci/jobs/defaults.yml
```diff
@@ -0,0 +1,39 @@
+---
+##### GLOBAL METADATA
+
+- meta:
+  cluster: devops-ci
```
jonahbull: For these top-level properties, the child items need to be indented or the YAML won't get parsed correctly.
```suggestion
- meta:
    cluster: devops-ci
```

.ci/jobs/defaults.yml
```diff
@@ -0,0 +1,39 @@
...
+
+##### JOB DEFAULTS
+
+- job:
```
jonahbull, line 9: Same issue as above, the other properties that are part of the job map need to be indented.

.ci/jobs/elastic+terraform-provider-elasticstack+main.yml
```diff
@@ -0,0 +1,13 @@
+---
+- job:
+  name: elastic+terraform-provider-elasticstack+main
```
jonahbull: Same indentation comment as elsewhere.

.ci/jobs/elastic+terraform-provider-elasticstack+pull-request.yml
```diff
@@ -0,0 +1,21 @@
+---
+- job:
+  name: elastic+terraform-provider-elasticstack+pull-request
```
jonahbull: Same indentation comment as elsewhere.

.ci/jobs/elastic+terraform-provider-elasticstack+main.yml
```diff
@@ -0,0 +1,13 @@
...
+  display-name: elastic / terraform-provider-elasticstack - main
+  description: Main branch testing
+  project-type: pipeline
```
jonahbull: If you have this set in the defaults you could get away with not setting it in the jobs.

.ci/jobs/elastic+terraform-provider-elasticstack+pull-request.yml
```diff
@@ -0,0 +1,21 @@
...
+  display-name: elastic / terraform-provider-elasticstack - pull-request
+  description: Pull request testing
+  project-type: pipeline
```
jonahbull: If you have this set in the defaults you could get away with not setting it in the jobs.

PR #6, opened 2021-11-08
Suppress index template diff

olksdr:
Make sure to normalize and check the user's supplied settings.

Index templates support configuring the index settings (https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#index-modules-settings)
and ES API flexible enough so the following formats if are sent by user will be valid:

```json
{ "index.number_of_routing_shards": "2" }
```

```json
{ "number_of_routing_shards": "2" }
```

```json
{ "index": { "number_of_routing_shards": 2 } }
```

This PR adds special suppression function, which makes sure to normalize
the configured settings and the settings which are returned from ES API
and only then compare them and decides if there is a diff drift.

Also add few tests for utils package and small refactoring of the project
structure.

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -45,9 +45,37 @@ func MapsEqual(m1, m2 interface{}) bool {
 	return reflect.DeepEqual(m2, m1)
 }
 
-func DiffJsonSuppress(k, old, new string, d *schema.ResourceData) bool {
-	result, _ := JSONBytesEqual([]byte(old), []byte(new))
-	return result
+// Flattens the multilevel map, and concatenates keys together with dot "."
+// # Exmaples
+// map of form:
+//     map := map[string]interface{}{
+//             "index": map[string]interface{}{
+//                     "key": 1
+//             }
+//     }
+// becomes:
+//     map := map[string]interface{}{
+//             "index.key": 1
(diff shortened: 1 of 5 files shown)


internal/utils/utils.go
```diff
@@ -45,9 +45,37 @@ func MapsEqual(m1, m2 interface{}) bool {
...
+			switch inner := v.(type) {
+			case map[string]interface{}:
+				flattener(k+key, inner, dst)
+			default:
+				dst[k+key] = v
+			}
+		}
+	}
+	flattener("", m, out)
+	return out
```
JoshMock, line 78: Is it safe to assume that none of the nested values are arrays? This solution looks like it would not flatten values inside an array, but maybe there are no arrays in index settings, or array equality would handle that safely.

olksdr: There is only 1 setting in index settings, which I could find can accept either string or array of strings `index.query.default_field`. But it depends a lot on the way how the API will return the data back if the order preserved the equality will work out perfectly fine, if the order every time will be random, then we will have diff drift, so flattening them won't help either - since they will have different indices after flattening as well, meaning they will have different key names. 

We can come back to this once again after some extensive testing  again.

PR #7, opened 2021-11-08
CI: minor changes in the sh step, use environment variable for parameters, and default git checkout credentials

v1v:
### What

Cosmetic changes in the `sh` step to use the label argument to highlight whats the step about in the OpenTelemetry distributed traces.
Set environment variable with the parameter value to avoid issues with the very first build.
Use default git checkout credentials defined in `estcGithubCheckout`

### DIFF
--- a/.ci/pipelines/acceptance.groovy
+++ b/.ci/pipelines/acceptance.groovy
@@ -6,7 +6,6 @@ pipeline {
     HOME = "${env.JENKINS_HOME}"
     REPOSITORY = "terraform-provider-elasticstack"
     GIT_REFERENCE_REPO = "/var/lib/jenkins/.git-references/terraform-provider-elasticstack.git"
-    GIT_CREDENTIALS = "f6c7695a-671e-4f4f-a331-acdce44ff9ba"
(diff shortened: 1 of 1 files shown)


.ci/pipelines/acceptance.groovy
```diff
@@ -2,37 +2,31 @@
...
-        script {
-          dir('.') {
-            sh '''#!/bin/bash
-
-              make docker-testacc
-            '''
-          }
-        }
-      } 
+        sh(label: 'Run ATs in docker', script: 'make docker-testacc')
```
v1v:
```suggestion
        dir('.') {
          sh(label: 'Run ATs in docker', script: 'make docker-testacc')
        }
```

.ci/pipelines/acceptance.groovy
```diff
@@ -2,37 +2,33 @@
...
     stage('Acceptance Tests') {
       steps {
-        script {
-          dir('.') {
-            sh '''#!/bin/bash
-
-              make docker-testacc
-            '''
-          }
+        dir('.') {
```
v1v, line 26: I'm not 100% sure this is required

PR #8, opened 2021-11-08
[ci] change the url in CI configuration

olksdr:
If the URL contains HTTP(s) the checkout would be made, looks like,
using password auth, which is not supported by GitHub.
We must use SSH to get our source.

### DIFF
--- a/.ci/jobs/defaults.yml
+++ b/.ci/jobs/defaults.yml
@@ -20,7 +20,7 @@
           description: "the Git branch specifier to build (&lt;branchName&gt;, &lt;tagName&gt;,&lt;commitId&gt;, etc.)"
     properties:
       - github:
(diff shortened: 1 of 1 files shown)


.ci/jobs/defaults.yml
```diff
@@ -20,7 +20,7 @@
           description: "the Git branch specifier to build (&lt;branchName&gt;, &lt;tagName&gt;,&lt;commitId&gt;, etc.)"
     properties:
       - github:
-          url: https://github.com/elastic/terraform-provider-elasticstack
+          url: git@github.com:elastic/terraform-provider-elasticstack.git
```
v1v, line 23: For the record, this is exactly what we have done in other places, see https://github.com/elastic/apm-pipeline-library/blob/master/.ci/jobs/apm-schedule-daily.yml#L17

PR #9, opened 2021-11-08
Align provider auth configuration

### LINKED ISSUE #2: Align "elasticsearch_connection" and "provider" blocks
When configuring a single stack using the _provider_ for the connection details, users need to specify 
```url``` , ```username```, ```password```

When configuring the ```elasticsearch_connection``` block on a specific resource, users specify 
```endpoints = ["url1","url2"..]``` , ```username```,```password```

We should probably align both to use the latter "endpoints" for easier UX, allowing users to use the same set of config params.

Also, in the provider block, it may be worthwhile thinking how we depict these are only "elasticsearch" connection details.
potentially - the user can, later on, connect to other resources as well, like kibana, ent_search, etc...

no preference here, let's discuss.
a random proposal (without tech-feasiblity knowhow) from me would be something like:

```
provider elasticstack {
   elasticsearch {
      endpoints = ["url1","url2"...]
      username = "elastic"
      password = "changeme"
   }
  
  # for future
  kibana {
     endpoints = ["url1","url2"...]
     username = "elastic"
     password = "changeme"
  }

}
```

### PR DESCRIPTION
**Pull request by olksdr**
Add per product auth configuration blocks, which allow to align the way
the connections to APIs are made.

/cc @Kushmaro  please, also have a look.  It includes also an example docs changes, in `templates/index.md.tmpl` which used to generate everything in the `docs/`

fix #2

### CODE REVIEW

File: docs/index.md [hunk 1]
```diff
@@ -14,17 +14,72 @@ It is recommended to setup at least minimum security, https://www.elastic.co/gui
...
 ### Optional
 
+- **elasticsearch** (Block List, Max: 1) Default Elasticsearch connection configuration block. (see [below for nested schema](#nestedblock--elasticsearch))
+
+<a id="nestedblock--elasticsearch"></a>
+### Nested Schema for `elasticsearch`
+
+Optional:
+
+- **endpoints** (List of String, Sensitive) A comma-separated list of endpoints where the terraform provider will point to, this must include the http(s) schema and port number.
```
Crazybus, line 83:
```suggestion
- **endpoints** (List of Strings, Sensitive) A comma-separated list of endpoints where the terraform provider will point to, this must include the http(s) schema and port number.
```

Not sure what the convention is but this sounds better in my head

olksdr: Yeah, but this schema is generated from `templates/index.md.tmpl` file by https://github.com/hashicorp/terraform-plugin-docs which is a standard I would assume

Crazybus: Makes sense! Thanks

PR #11, opened 2021-11-15
Add resource to manage Index Lifecycle (ILM) policies 

### LINKED ISSUE #3: [feature] Support ILM policies
Details TBD, will be edited once hashed out

olksdr:
This resource creates or updates lifecycle policy.
An index’s lifecycle policy specifies which phases
are applicable, what actions are performed in each phase,
and when it transitions between phases.

In this PR:
* add the TF dsl to configure the ILM policies
* make sure the validate the edge cases
* make the configuration uniform for all the phases and actions

closes #3

### DIFF
--- a/internal/elasticsearch/index/ilm.go
+++ b/internal/elasticsearch/index/ilm.go
@@ -123,18 +123,21 @@ var suportedActions = map[string]*schema.Schema{
 					Description:      "Assigns an index to nodes that have at least one of the specified custom attributes.",
 					Type:             schema.TypeString,
 					Optional:         true,
+					ValidateFunc:     validation.StringIsJSON,
 					DiffSuppressFunc: utils.DiffJsonSuppress,
 				},
 				"exclude": {
 					Description:      "Assigns an index to nodes that have none of the specified custom attributes.",
 					Type:             schema.TypeString,
 					Optional:         true,
+					ValidateFunc:     validation.StringIsJSON,
 					DiffSuppressFunc: utils.DiffJsonSuppress,
 				},
 				"require": {
 					Description:      "Assigns an index to nodes that have all of the specified custom attributes.",
 					Type:             schema.TypeString,
 					Optional:         true,
+					ValidateFunc:     validation.StringIsJSON,
 					DiffSuppressFunc: utils.DiffJsonSuppress,
 				},
 			},
@@ -374,7 +377,7 @@ func resourceIlmPut(ctx context.Context, d *schema.ResourceData, meta interface{
 		return diags
 	}
 
-	policy, diags := explandIlmPolicy(d)
+	policy, diags := expandIlmPolicy(d)
 	if diags.HasError() {
 		return diags
 	}
@@ -399,7 +402,7 @@ func resourceIlmPut(ctx context.Context, d *schema.ResourceData, meta interface{
 	return resourceIlmRead(ctx, d, meta)
 }
 
-func explandIlmPolicy(d *schema.ResourceData) (*models.Policy, diag.Diagnostics) {
+func expandIlmPolicy(d *schema.ResourceData) (*models.Policy, diag.Diagnostics) {
 	var diags diag.Diagnostics
 	var policy models.Policy
 	phases := make(map[string]models.Phase)


internal/elasticsearch/index/ilm_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+    rollover {
+      max_age = "2d"
+    }
+  }
+}
+ `, name)
+}
+
+func checkResourceILMDestroy(s *terraform.State) error {
```
gtback, line 112: Is there a test that uses this function?

internal/elasticsearch/index/ilm.go
```diff
@@ -0,0 +1,697 @@
...
+			Description: "The DateTime of the last modification.",
+			Type:        schema.TypeString,
+			Computed:    true,
+		},
+	}
+
+	utils.AddConnectionSchema(ilmSchema)
+
+	return &schema.Resource{
+		Description: "Creates or updates lifecycle policy. See,https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-put-lifecycle.html and  https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-index-lifecycle.html",
```
gtback:
```suggestion
		Description: "Creates or updates lifecycle policy. See: https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-put-lifecycle.html and  https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-index-lifecycle.html",
```

internal/elasticsearch/index/ilm_test.go
olksdr: yes, see line 20 in the same file.

internal/elasticsearch/index/ilm.go
olksdr: this is fixed in c547a16

internal/elasticsearch/index/ilm_test.go
gtback: Ah, thanks. I was only looking at the `Config` steps. TIL.

internal/elasticsearch/index/ilm.go
```diff
@@ -0,0 +1,700 @@
...
+			DiffSuppressFunc: utils.DiffJsonSuppress,
+		},
+		"hot": {
+			Description:  "The index is actively being updated and queried.",
+			Type:         schema.TypeList,
+			Optional:     true,
+			MaxItems:     1,
+			AtLeastOneOf: []string{"hot", "warm", "cold", "frozen", "delete"},
+			Elem: &schema.Resource{
+				Schema: getSchema("set_priority", "unfollow", "rollover", "readonly", "shrink", "forcemerge", "searchable_snapshot"),
```
Crazybus, line 41: This is really cool. Is this a Terraform pattern or did you come up with it? In either case me like.

internal/elasticsearch/index/ilm.go
```diff
@@ -0,0 +1,700 @@
...
+		},
+
+		Schema: ilmSchema,
+	}
+}
+
+var suportedActions = map[string]*schema.Schema{
+	"allocate": {
+		Description: "Updates the index settings to change which nodes are allowed to host the index shards and change the number of replicas.",
+		Type:        schema.TypeList,
```
Crazybus, line 112: How come these are all TypeList with MaxItems 1? Just genuinely curious because I can't figure it out. Is it because each of these "blocks" could in theory being defined multiple times so this is how you define it?

internal/elasticsearch/index/ilm.go
```diff
@@ -0,0 +1,700 @@
...
+		Optional:    true,
+		MaxItems:    1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"number_of_shards": {
+					Description: "Number of shards to shrink to.",
+					Type:        schema.TypeInt,
+					Optional:    true,
+				},
+				"max_primary_shard_size": {
```
Crazybus, line 310: This one conflicts with `number_of_shards` if that is something we wanted to add into the schema.

internal/elasticsearch/index/ilm.go
```diff
@@ -0,0 +1,700 @@
...
+}
+
+func getSchema(actions ...string) map[string]*schema.Schema {
+	sch := make(map[string]*schema.Schema)
+	for _, a := range actions {
+		if action, ok := suportedActions[a]; ok {
+			sch[a] = action
+		}
+	}
+	// min age can be set for all the phases
```
Crazybus, line 358: Is this true? I don't think you can set it for the hot phase but I'm also not sure if it matters.

internal/elasticsearch/index/ilm.go
```diff
@@ -0,0 +1,700 @@
...
+
+	if v, ok := d.GetOk("metadata"); ok {
+		metadata := make(map[string]interface{})
+		if err := json.NewDecoder(strings.NewReader(v.(string))).Decode(&metadata); err != nil {
+			return nil, diag.FromErr(err)
+		}
+		policy.Metadata = metadata
+	}
+
+	if v, ok := d.GetOk("hot"); ok {
```
Crazybus, line 420: Is there anything stopping this from just being a loop over all phase strings like this? (not tested):

```go
	phases := []string{"hot", "warm"}
	for _, p := range phases {
		if v, ok := d.GetOk(p); ok {
			// if defined it must contain only 1 entry map
			phase, diags := expandPhase(v.([]interface{})[0].(map[string]interface{}))
			if diags.HasError() {
				return nil, diags
			}
			phases[p] = *phase
		}
	}
```

internal/elasticsearch/index/ilm.go
```diff
@@ -0,0 +1,700 @@
...
+	if ilmDef.Policy.Metadata != nil {
+		metadata, err := json.Marshal(ilmDef.Policy.Metadata)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		if err := d.Set("metadata", string(metadata)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+	if v, ok := ilmDef.Policy.Phases["hot"]; ok {
```
Crazybus, line 603: Same question as above about whether or not this could just be a for loop.

internal/elasticsearch/index/ilm.go
olksdr: It's possible to add additional check and error out if the user provides both settings. Right now we just get the API error if the both are set. 
I can make adjustments in the followup PR

internal/elasticsearch/index/ilm.go
olksdr: yeah, to make sure we have only 1 block defined , we must have it `MaxItems: 1` this way TF will check the settings while evaluating the defined configuration

internal/elasticsearch/index/ilm.go
olksdr: it's just me trying to follow DRY 😄 and generate schema for each block without copy-pasting much

internal/elasticsearch/index/ilm.go
olksdr: Hm, from what I could find yes. But I will still double check it. 
Tests are passing and manual testing also showed that this option can be set here

internal/elasticsearch/index/ilm.go
olksdr: yeah, this could be done simpler. I do not remember why I decided to go with explicit definitions, but I can follow up on this and refactor in followup PR

internal/elasticsearch/index/ilm.go
olksdr: will refactor in followup PR

PR #16, opened 2021-11-18
A guide for using the elasticstack provider alongside the elasticcloud one

Kushmaro:
Create a "guide" according to hashicorp terraform registry guidelines.

### DIFF
--- a/docs/guides/elasticstack-and-cloud.md
+++ b/docs/guides/elasticstack-and-cloud.md
@@ -7,8 +7,8 @@ description: |-
 
 # 
 
-A common scenario for using the Elastic Stack provider, is to manage & configure Elastic Cloud deploymemnts.
-in order to do that, we'll use both the Elastic Cloud provider, as well as the Elastic Stack provider.
+A common scenario for using the Elastic Stack provider, is to manage & configure Elastic Cloud deployments.
+In order to do that, we'll use both the Elastic Cloud provider, as well as the Elastic Stack provider.
 Start off by configuring the two providers in a `provider.tf` file for example:
 
 ```terraform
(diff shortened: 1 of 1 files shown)


templates/guides/elasticstack-and-cloud.md.tmpl
```diff
@@ -0,0 +1,22 @@
...
+
+# 
+
+A common scenario for using the Elastic Stack provider, is to manage & configure Elastic Cloud deploymemnts.
+in order to do that, we'll use both the Elastic Cloud provider, as well as the Elastic Stack provider.
+Start off by configuring the two providers in a `provider.tf` file for example:
+
+{{ tffile "examples/cloud/provider.tf" }}
+
+Notice that the elasticstack provider is not configured, since we'll be using an `elasticsearch_connection` block
```
olksdr: In some places you use `Elastic Stack provider` and sometime `elasticstack provider`. It would be great to have consistency in the naming
```suggestion
Notice that the Elastic Stack  provider is not configured, since we'll be using an `elasticsearch_connection` block
```

templates/guides/elasticstack-and-cloud.md.tmpl
```diff
@@ -0,0 +1,22 @@
...
+for each of our resources, to point to the elastic cloud deployment. This is because `terraform` can not configure providers that are dependent on one anoher.
```
olksdr: > This is because `terraform` can not configure providers that are dependent on one anoher.
I'm pretty sure it should be possible to provide credentials to the provider from the resource of the another provider. Have you tested this? What error are you getting? 

```suggestion
for each of our resources, to point to the Elastic Cloud deployment. This is because `terraform` can not configure providers that are dependent on one another.
```

templates/guides/elasticstack-and-cloud.md.tmpl
```diff
@@ -0,0 +1,22 @@
...
+
+Next, we'll set up an Elastic Cloud `ec_deployment` resource, which represents an Elastic stack deployment on Elastic Cloud.
```
olksdr: should the `stack` be capitalised here?
```suggestion
Next, we'll set up an Elastic Cloud `ec_deployment` resource, which represents an Elastic Stack deployment on Elastic Cloud.
```

templates/guides/elasticstack-and-cloud.md.tmpl
```diff
@@ -0,0 +1,22 @@
+---
+subcategory: ""
+page_title: "Using the Elastic Stack provider with Elastic Cloud"
+description: |-
+    An example of how to spin up a deployment and configure it in a single plan.
+---
+
+# 
+
+A common scenario for using the Elastic Stack provider, is to manage & configure Elastic Cloud deploymemnts.
```
olksdr:
```suggestion
A common scenario for using the Elastic Stack provider, is to manage & configure Elastic Cloud deployments.
```

templates/guides/elasticstack-and-cloud.md.tmpl
```diff
@@ -0,0 +1,22 @@
...
+in order to do that, we'll use both the Elastic Cloud provider, as well as the Elastic Stack provider.
```
olksdr:
```suggestion
In order to do that, we'll use both the Elastic Cloud provider, as well as the Elastic Stack provider.
```

PR #25, opened 2021-12-09
Refactor: move requests handling into `clients` module

olksdr:
There are looks like lots of changes, but this is only the same code was moved around.

Moving all the requests handling into the `clients` module we make sure to have only single place where we have to take care of the clients related issues/upgrades etc.
All the requests just wrapped into `ApiClient`, which handles Marshaling/Unmarshaling of the data and returns proper types to the caller to process them.

Also as part of this PR fixed examples, added `elasticsearch` configuration block and regenerated the docs.

### DIFF
--- a/docs/data-sources/elasticsearch_security_user.md
+++ b/docs/data-sources/elasticsearch_security_user.md
@@ -13,7 +13,9 @@ Use this data source to get information about existing Elasticsearch user.
 ## Example Usage
 
 ```terraform
-provider "elasticstack" {}
+provider "elasticstack" {
+  elasticsearch {}
+}
 
 data "elasticstack_elasticsearch_security_user" "user" {
   username = "elastic"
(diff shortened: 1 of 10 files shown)


internal/clients/api_client.go
```diff
@@ -155,3 +161,413 @@ func (a *ApiClient) ClusterID() (string, error) {
 
 	return "", fmt.Errorf("Unable to get cluster uuid")
 }
+
```
Crazybus, line 164: :bulb: Wonder if this has hit the point where breaking these out into separate files based on the API would make sense.

olksdr: This is a good point! As of now I just moved the code into the `ApiClient` and for next steps, to make this prettier, would be to remove some code duplication and move some functionality into separate files or even modules.

PR #35, opened 2022-01-11
add "ca_file" option to the elasticstack provider

### LINKED ISSUE #34: Support self-signed / internal Certificate Authorities
My Elasticsearch cluster uses an internal Certificate Authority to secure all network communication.  When I try and use this provider with my cluster, I get:  Error: x509: certificate signed by unknown authority

I see the "ec" provider supports  "insecure = true", but when I try adding setting to this provider, I get:  An argument named "insecure" is not expected here.

Is there any way to use this provider with self signed / internal  certificates?

utilitynerd:
allows connecting to an elasticsearch cluster using a custom certificate authority.

I know just enough go to get myself in trouble, but figured I would try to solve my own problem anyways. resolves #34 

It adds the "ca_file" parameter to the elasticstack provider.  If ca_file is set, the certificate is read and its contents are added to the CACert field in the elasticsearch.Config struct.

### DIFF
--- a/internal/clients/api_client.go
+++ b/internal/clients/api_client.go
@@ -90,6 +90,7 @@ func NewApiClientFunc(version string, p *schema.Provider) func(context.Context, 
 					tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
 					config.Transport = tr
 				}
+
 				if caFile, ok := esConfig["ca_file"]; ok && caFile.(string) != "" {
 					caCert, err := ioutil.ReadFile(caFile.(string))
 					if err != nil {
--- a/internal/clients/api_client.go
+++ b/internal/clients/api_client.go
@@ -98,6 +98,7 @@ func NewApiClientFunc(version string, p *schema.Provider) func(context.Context, 
 							Summary:  "Unable to read CA File",
 							Detail:   err.Error(),
 						})
+						return nil, diags
 					}
 					config.CACert = caCert
 				}
--- a/internal/clients/api_client.go
+++ b/internal/clients/api_client.go
@@ -146,6 +146,13 @@ func NewApiClient(d *schema.ResourceData, meta interface{}) (*ApiClient, error) 
 			tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
 			config.Transport = tr
 		}
+		if caFile, ok := conn["ca_file"]; ok && caFile.(string) != "" {
+			caCert, err := ioutil.ReadFile(caFile.(string))
+			if err != nil {
+				return nil, fmt.Errorf("Unable to read ca_file: %w", err)
+			}
+			config.CACert = caCert
+		}
 
 		es, err := elasticsearch.NewClient(config)
 		if err != nil {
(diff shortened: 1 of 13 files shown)


Kushmaro:
Thanks so much for putting up the first community PR @utilitynerd , I have to ask though - is there a specific reason for not implementing "insecure=true" like mentioned in the issue? that would be the direction we'd like to take to resolve this one.

Specifying a root_ca file is often quite cumbersome, and doesn't add much value in TF as it does on browsers for example (where browsers will show if a cert is trusted or not)

internal/clients/api_client.go
```diff
@@ -82,6 +83,17 @@ func NewApiClientFunc(version string, p *schema.Provider) func(context.Context,
...
 					config.Addresses = endpoints
 				}
+				if caFile, ok := esConfig["ca_file"]; ok && caFile.(string) != "" {
+					caCert, err := ioutil.ReadFile(caFile.(string))
+					if err != nil {
+						diags = append(diags, diag.Diagnostic{
+							Severity: diag.Error,
+							Summary:  "Unable to read CA File",
+							Detail:   err.Error(),
+						})
```
olksdr, line 101:
```suggestion
						})
						return nil, diags
```
It would be great if we immediately return the `diags` here to report the error sooner.

olksdr (changes requested):
Hey @utilitynerd, 
Thanks for your contribution! 

There are few more steps to get this over and get merged:
* things we're also support the per elasticsearch-resource `elasticsearch_connection` block, we will need to add `ca_file` in there as well, see https://github.com/elastic/terraform-provider-elasticstack/blob/f0c6a1fe04b0883dbe3386087c58f20530a86d70/internal/utils/utils.go#L124-L149 so this way we will stay consistent in all the places, and then you can update code in  https://github.com/elastic/terraform-provider-elasticstack/blob/f0c6a1fe04b0883dbe3386087c58f20530a86d70/internal/clients/api_client.go#L123-L135 to read the newly added option
* after changes are maybe, the docs must be updates as well, you can run `make gen` command which will re-generate the docs and adds this options as well
* and one more step is you will have you sign the [Contributor Agreement](https://www.elastic.co/contributor-agreement) for the account you're pushing your commits from,  in this case it looks like the commit is coming from https://github.com/mikejones-ucb

Once this is done we can once more go over the PR and see if there are any missing things to be added.

Thanks again for your contribution!

utilitynerd:
Thanks for the feedback!

  I'm not opposed to the "insecure=true" method instead of validating against a custom cert.  Although maybe it would be nice to eventually have both options available.

Since it sounds like the project would prefer the insecure route, why don't I take your suggestions and see if I can  put together a separate PR implementing it.

utilitynerd:
I believe I've addressed the issues above:
- return early loading the ca_file results in an error
- added the ca_file parameter to the elasticsearch_connection block 
- generated updated documentation
- signed the contributor agreement with my other username 


Let me know if there is anything else to address

olksdr:
jenkins test this please

olksdr:
@utilitynerd please, merge `main` into your branch and resolve the conflicts which are coming from #36

olksdr:
ok to test

olksdr (approved):
lgtm

PR #38, opened 2022-01-20
Add Index resource

olksdr:
This PR adds new `index` resource which will handle the index creation with all the aliases, mappings and settings. Also it's possible to use this resource to further manage and update different parts of the ES index, like settings, aliases without destroying the index. 

closes https://github.com/elastic/terraform-provider-elasticstack/issues/33

PR #39, opened 2022-01-20
Add component template resource

xunleii:
This PR adds new `component_template` resource to manage [components template](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-component-template.html) through TF resources. 
This is a "naive" implementation of this resource, mainly based on the `index_template` one (many part are copy/paste).  

Of course, if needed, I can spend more time to see how can I simplify by sharing code between `index_template` and `component_template`, but could decrease the readability.

_NOTE: this PR is "in conflict" with the https://github.com/elastic/terraform-provider-elasticstack/pull/38/files on how we flatten struct; I use a method directly on models and @olksdr uses a function that take the object as parameter. Because its PRs has been already reviewed, I can adapt my code for its PR to avoid conflict (keeping its work or using method on models, whichever you and @olksdr prefer)._
_NOTE²: https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/index/template.go#L220 will panic if no `aliases` are defined on `index_template`. In order to avoid any conflict with https://github.com/elastic/terraform-provider-elasticstack/pull/38/files, I didn't made any changes to fix that. But if my PR is be based on the later one, I will fix it._
_NOTE³: My changes need more tests, so I'll add them tomorrow_


olksdr:
Hey @xunleii! Thanks for your contribution! 

> NOTE²: https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/index/template.go#L220 will panic if no aliases are defined on index_template. 

Could you, please, also provide the error you get?  It would be great if you could open an issue with error and possibly traces, your TF version, the resource definition you're using to test this and system you're running this on.

xunleii:
Hi @olksdr, thanks for your response (and thanks to the team for this provider).
For the panic, I probably missed something when I ran the acceptance tests yesterday; I can no longer reproduce the panic on `index_template` on a clean environment (nor on `component_template` if I delete the last commit)... so should probably be a PEBCAK :sweat_smile:
Also, is there anything to do to validate the CLA verification? I signed the agreement yesterday but this step remains failed (probably missed something).

olksdr:
Hey @xunleii ,
could you, please, merge the `main` in and resolve the conflicts in your branch? So we could review your changes. 


> NOTE: this PR is "in conflict" with the https://github.com/elastic/terraform-provider-elasticstack/pull/38/files on how we flatten struct; I use a method directly on models and @olksdr uses a function that take the object as parameter. Because its PRs has been already reviewed, I can adapt my code for its PR to avoid conflict (keeping its work or using method on models, whichever you and @olksdr prefer).

I would prefer to use existing function, which is already used in the `index_template` and `index` resources. But proposed changes also make sense. 

It would be better to keep this PR simpler and just introduce the new resource, with tests and docs and move any refactoring into separate PR. With refactoring it would be great to move many of free standing functions into the models methods.

docs/resources/elasticsearch_component_template.md
```diff
@@ -0,0 +1,101 @@
+---
```
olksdr, line 1: All the documentation in the `docs/` folder is generated and should not be changed manually. See how it's done for `template` resource: https://github.com/elastic/terraform-provider-elasticstack/blob/0beb7d40a0f168e3e6b83ed6c579061b798e1ad4/templates/resources/elasticsearch_index_template.md.tmpl#L1-L12 

you could copy this template and change the specific parts in names and the paths to the examples and run `make gen` to generate the required files, which you then can add into this PR as well.

internal/elasticsearch/cluster/component_template.go
```diff
@@ -0,0 +1,290 @@
+package cluster
```
olksdr: Could you, please, move this file into `internal/elasticsearch/index/` folder (`index` module), where the `template` resource is located, just to keep them all together?

internal/elasticsearch/cluster/component_template.go
```diff
@@ -0,0 +1,290 @@
...
+	"github.com/elastic/terraform-provider-elasticstack/internal/models"
+	"github.com/elastic/terraform-provider-elasticstack/internal/utils"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+func ResourceComponentTemplate() *schema.Resource {
+	// NOTE: component_template and index_template uses the same schema
+	componentTemplateSchema := map[string]*schema.Schema{
```
olksdr: Please, also add the ID as `Computed` value:

```suggestion
    componentTemplateSchema := map[string]*schema.Schema{
			"id": {
			Description: "Internal identifier of the resource",
			Type:        schema.TypeString,
			Computed:    true,
		},
```

olksdr (changes requested):
Looks nice!

I've added few comments about documentation generation, and the placement of the new resource.

olksdr:
ok to test

olksdr:
And one more thing: 
please add the record in `CHANGELOG.md`, like here https://github.com/elastic/terraform-provider-elasticstack/blob/0beb7d40a0f168e3e6b83ed6c579061b798e1ad4/CHANGELOG.md#L3

docs/resources/elasticsearch_component_template.md
xunleii: I used `make docs-generate` to generate this file, but without adding the template file. I'll add it and regenete it the new documentation just after

xunleii:
Hi @olksdr, thanks for your review. I made all changes that you asked me and I made them inside two `fixup` commits if you want to check these changes only.
When all will be good, I will squash them to clean the git history.

internal/clients/index.go
```diff
@@ -78,6 +78,69 @@ func (a *ApiClient) DeleteElasticsearchIlm(policyName string) diag.Diagnostics {
...
+}
+
+func (a *ApiClient) GetElasticsearchComponentTemplate(templateName string) (*models.ComponentTemplateResponse, diag.Diagnostics) {
+	var diags diag.Diagnostics
+	req := a.es.Cluster.GetComponentTemplate.WithName(templateName)
+	res, err := a.es.Cluster.GetComponentTemplate(req)
+	if err != nil {
+		return nil, diag.FromErr(err)
+	}
+	defer res.Body.Close()
```
olksdr, line 109:
```suggestion
	defer res.Body.Close()
	if res.StatusCode == http.StatusNotFound {
		return nil, nil
	}
```

We also need to handle situation when the component was deleted manually, but still has record in the TF state and in the TF config files. 

See also template, when the removed template was handled same way: https://github.com/elastic/terraform-provider-elasticstack/blob/68ba59435c4166b6f929345d7f506811e24f8fa8/internal/clients/index.go#L172-L174

Also see my other suggestion, when the check is done and ID of the resource set to empty string (`""`) to force the  re-creation of the resource

internal/elasticsearch/index/component_template.go
```diff
@@ -0,0 +1,294 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	compId, diags := clients.CompositeIdFromStr(d.Id())
+	if diags.HasError() {
+		return diags
+	}
+	templateId := compId.ResourceId
+
+	tpl, diags := client.GetElasticsearchComponentTemplate(templateId)
```
olksdr, line 239:
```suggestion
	tpl, diags := client.GetElasticsearchComponentTemplate(templateId)
	if tpl == nil && diags == nil {
		d.SetId("")
		return diags
	}
```

This is another part of the suggestion when we need to handle the deleted resource outside of the TF

see also how it's done in template https://github.com/elastic/terraform-provider-elasticstack/blob/68ba59435c4166b6f929345d7f506811e24f8fa8/internal/elasticsearch/index/template.go#L285-L289

olksdr (changes requested):
Thanks for addressing my review comments. 

Only one small thing left to do, see my 2 proposed suggestion, to handle the situation when the resource was deleted outside of the TF. Those 2 pieces of. code also need to be added in, and we are good to merge this new resource.

And thanks again for working on this!

xunleii:
Hi, I applied your suggestion and resolved the conflict on CHANGELOG. Thanks

olksdr:
jenkins test this

Issue #37, opened 2022-01-20
Error deleting snapshot resources manually


cmanzur:
This is a **bug** report:

1- Create this resource and `terraform apply`
```
resource "elasticstack_elasticsearch_snapshot_repository" "s3" {
  name = "s3"

  s3 {
    bucket = "backups"
    compress  = true
  }
}
```

2- Manually delete the repository from KIBANA UI.

3- Run `terraform apply` again

```
Error: Unable to get the information about snapshot repository: s3

Failed with:
{"error":{"root_cause":[{"type":"repository_missing_exception","reason":"[s3]
missing"}],"type":"repository_missing_exception","reason":"[s3]
missing"},"status":404}
```

Same applies for: **elasticstack_elasticsearch_snapshot_lifecycle**

**Notes:**
To use s3, you must install the plugin: `bin/elasticsearch-plugin install --batch repository-s3`

ES version: **docker.elastic.co/elasticsearch/elasticsearch:7.16.3**

olksdr:
Hey @cmanzur, thanks for reporting this issue!  I'll be looking into this in the beginning of the next week.

olksdr:
New version of the provider was published, see: [v0.2.0](https://github.com/elastic/terraform-provider-elasticstack/releases/tag/v0.2.0)

PR #40, opened 2022-01-25
Identify the missing/deleted resources in ES cluster

olksdr:
For TF to understand that some resources were deleted manually (maybe by accident) we must make sure to set the resource id to empty string (""). In our case if ES reports back 404 we know that requested resource is missing and we can let TF restore it based on the configuration or update the TF state if the deletion was done appropriately. 

As part of this PR I decided to split the ES client file into few different files by the cluster functionality. 

fix: https://github.com/elastic/terraform-provider-elasticstack/issues/37


olksdr:
For review please see commit 69353a3, which has actual change.

Crazybus (approved):
LGTM!

JoshMock (approved):

PR #44, opened 2022-01-27
Added title in the cloud guide

olksdr:
Currently this title is missing which leaves empty `#` in the beginning of the file, see https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/guides/elasticstack-and-cloud


Crazybus (approved):
LGTM!

PR #45, opened 2022-01-28
Add data stream resource

olksdr:
This PR adds `elasticstack_elasticsearch_data_stream` resource which makes it possible to create data stream based on index templates.


Crazybus (approved):
LGTM!

PR #51, opened 2022-01-31
Make sure to define ID as rad-only/computed attribute

olksdr:
We must explicitly define ID in the schema, to make sure it's rendered as read-only in the docs, since it should not be possible to change or set the id in the TF resource definition.


Crazybus (approved):
LGTM!

Issue #53, opened 2022-01-31
[Bug] Index settings not being imported to terraform 


fdvieira-td:
**Describe the bug**
Hey guys 👋 We've found a bug in the provider when using `resource "elasticstack_elasticsearch_index"` where this resource does not import the existing index settings to terraform. 

**To Reproduce**
1. TF configuration used. Create a new index with alias, mappings and settings manually:
```
{
    "dummy": {
        "aliases": {
            "dummy1": {}
        },
        "mappings": {
            "properties": {
                "field1": {
                    "type": "text"
                }
            }
        },
        "settings": {
            "index": {
                "creation_date": "1643651976221",
                "number_of_shards": "3",
                "number_of_replicas": "2",
                "uuid": "cluster-uuid",
                "version": {
                    "created": "7090199"
                },
                "provided_name": "dummy"
            }
        }
    }
}
```
2. Add the configuration of this index to `terraform`. 
```
  # module.elasticsearch_indices.elasticstack_elasticsearch_index.this["dummy"] will be created
  + resource "elasticstack_elasticsearch_index" "this" {
      + id           = (known after apply)
      + mappings     = jsonencode(
            {
              + properties = {
                  + field1 = {
                      + type = "text"
                    }
                }
            }
        )
      + name         = "dummy"
      + settings_raw = (known after apply)

      + alias {
          + is_hidden      = false
          + is_write_index = false
          + name           = "dummy1"
        }

      + settings {
          + setting {
              + name  = "index.number_of_replicas"
              + value = "2"
            }
          + setting {
              + name  = "index.number_of_shards"
              + value = "3"
            }
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.
```
2. Import the index to terraform by running 
``` 
 terraform import 'module.elasticsearch_indices.elasticstack_elasticsearch_index.this["dummy"]' 'cluster-uuid/dummy' 
```
3. The output of the command is as follows:
```
module.elasticsearch_indices.elasticstack_elasticsearch_index.this["dummy"]: Importing from ID "cluster-uuid/dummy"...
module.elasticsearch_indices.elasticstack_elasticsearch_index.this["dummy"]: Import prepared!
  Prepared elasticstack_elasticsearch_index for import
module.elasticsearch_indices.elasticstack_elasticsearch_index.this["dummy"]: Refreshing state... [id=cluster-uuid/dummy]

Import successful!

The resources that were imported are shown above. These resources are now in
[...]

PR #55, opened 2022-02-01
Support import of the index settings

olksdr:
Make sure we are able to import all the index settings. Also, after index is created we support only the settings, which are defined in the TF files. 

fix https://github.com/elastic/terraform-provider-elasticstack/issues/53


gtback (commented):
I don't feel qualified to give this a +1, but it all looks good to me.

Crazybus (approved):
LGTM!

Issue #59, opened 2022-02-03
[Bug] Error when building `elasticstack_elasticsearch_index_lifecycle` resource


officialgawney:
Hello we're using the elastic provider in terraform to automate the build of some of our elastic resources. 

However when trying to build an ILM policy to our cluster via the **elasticstack_elasticsearch_index_lifecycle** resource the resource plans by fails to apply. 

Specifically we believe it's related the the use of the allocate block - we're able to run a plan / apply successfully without it. 

allocate {
    number_of_replicas = 0
}

**Please see our implementation below followed by the error message we're seeing on running terraform apply** 

**Implementation**

resource "elasticstack_elasticsearch_index_lifecycle" "index_lifecycle_default" {
  name = "logs-${module.account_insights.tribe_alias}-default"

  hot {
    min_age = "0ms"
    set_priority {
      priority = 100
    }
    rollover {
      max_primary_shard_size = "10gb"
      max_age = "14d"
    }
  }

  warm {
    min_age = "0ms"
    set_priority {
      priority = 60
    }
    readonly {}
    allocate {
      exclude = "{hot}"
      number_of_replicas = 0
    }
  }

  delete {
    min_age = "31d"
    delete {
      delete_searchable_snapshot = true
    }
    wait_for_snapshot {
      policy = "tvg-s3-policy"
    }
  }
}

**Error message**

Error: Unable to create or update the ILM policy

  with module.elastic_tribe_config.elasticstack_elasticsearch_index_lifecycle.index_lifecycle_default,
  on ../../lifecycle.tf line 1, in resource "elasticstack_elasticsearch_index_lifecycle" "index_lifecycle_default":
   1: resource "elasticstack_elasticsearch_index_lifecycle" "index_lifecycle_default" {

Failed with:
{"error":{"root_cause":[{"type":"x_content_parse_exception","reason":"Failed
to build [allocate] after last required field
arrived"}],"type":"x_content_parse_exception","reason":"[1:329]
[put_lifecycle_request] failed to parse field
[policy]","caused_by":{"type":"x_content_parse_exception","reason":"[1:329]
[lifecycle_policy] failed to parse field
[phases]","caused_by":{"type":"x_content_parse_exception","reason":"[1:329]
[phases] failed to parse field
[warm]","caused_by":{"type":"x_content_parse_exception","reason":"[1:329]
[phase] failed to parse field
[actions]","caused_by":{"type":"x_content_parse_exception","reason":"[1:329]
[...]

PR #60, opened 2022-02-04
Fix issue with `allocate` action for ILM

olksdr:
Allocate action can contain `include`, `exclude` and `require` fields, which are JSON objects and must be handled appropriately.

fix https://github.com/elastic/terraform-provider-elasticstack/issues/59


Crazybus (approved):
LGTM!

Issue #66, opened 2022-02-10
[Bug] Unable to update existing user without having most fields set


faudebert:
**Describe the bug**
For example, if `password` (or `password_hash`) is not known by Terraform (ex. after an import), it's not possible to manage this user with Terraform at all.

**To Reproduce**
1. Import an existing user with `terraform import 'elasticstack_elasticsearch_security_user.coin' abcdefg/coin`
2. Update your local resource to match the imported user. For example:
```tf
resource "elasticstack_elasticsearch_security_user" "coin" {
    enabled  = true
    roles    = [
        "custom_role",
    ]
    username = "coin"
}
```
3. Update a field, for example, `roles = ["custom_role", "custom_role2"]`
4. Run `terraform apply` which fails with the following error:

```
│ Failed with: {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Provided password
│ hash uses [NOOP] but the configured hashing algorithm is
│ [BCRYPT]"}],"type":"illegal_argument_exception","reason":"Provided password hash uses [NOOP] but the
│ configured hashing algorithm is [BCRYPT]"},"status":400}
```

**Expected behavior**
`custom_role2` role is added to user `coin`.

**Debug output**

```
[TRACE] sending request to ES: {"full_name":"","email":"","roles":["custom_role","custom_role2"],"password_hash":"","enabled":true}: timestamp=2022-02-10T17:19:29.352+0100
```

While not set by the resource, we can see `full_name`, `email` and `password_hash` fields are passed to Elasticsearch request with an empty string.

When not set or `null`, those parameters should not be added to Elasticsearch calls to allow to update relevant fields only.

**Desktop:**
 - OS: macOS
 - Terraform Version: v1.1.4
 - Provider version: v0.2.0

**Additional context**

This might be handy being able to manage users with Terraform even if they are initially provisioned by some other means (i.e. we don't want to manage password at all with Terraform).

olksdr:
hey @faudebert, thanks for opening issue and providing all the details! I will put this on my list to look into this asap.

olksdr:
New version of the provider released: [v0.3.0](https://github.com/elastic/terraform-provider-elasticstack/releases/tag/v0.3.0)

PR #68, opened 2022-02-14
Send only set values to ES api

olksdr:
fix https://github.com/elastic/terraform-provider-elasticstack/issues/66


Crazybus (approved):
LGTM!

PR #69, opened 2022-02-14
Add role mapping resource

ottramst:
Trying to add an initial implementation of the [role mapping resource](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-role-mapping.html). Pretty basic implementation where the user can specify `role_templates` and `rules` as a HEREDOC or `jsonencode`.

Go easy on me - first time implementing something in a Terraform provider and not so good at golang, yet :)

One issue which I can't figure out - during `make docker-testacc`, the test errors with `role_mapping_test.go:18: Step 1/2 error: Check failed: Check 1/4 error: elasticstack_elasticsearch_security_role_mapping.test: Attribute 'name' expected "m16rozeuju", got ""`, although everything should be correctly implemented. If someone can help me figure this out, I'll happily fix this.


internal/elasticsearch/security/role_mapping.go
```diff
@@ -0,0 +1,240 @@
...
+		"roles": {
+			Description: " A list of role names that are granted to the users that match the role mapping rule. Default is [].",
+			Type:        schema.TypeSet,
+			Optional:    true,
+			Computed:    true,
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+		},
+		"role_templates": {
```
olksdr, line 42: It would be better to define this `role_templates` as a `Set` of `map[string]*schema.Schema` to make sure it can be defined multiple times and use the proper TF domain specific language instead of just providing `json`

internal/elasticsearch/security/role_mapping.go
```diff
@@ -0,0 +1,240 @@
...
+			Description: "A list of mustache templates that will be evaluated to determine the roles names that should granted to the users that match the role mapping rules. Each record must be a valid JSON document.",
+			Type:        schema.TypeList,
+			Optional:    true,
+			MinItems:    1,
+			Elem: &schema.Schema{
+				Type:         schema.TypeString,
+				ValidateFunc: validation.StringIsJSON,
+			},
+		},
+		"rules": {
```
olksdr, line 52: Same goes about the rules, it might be better to provide the proper schema here for all the rules types which can be supported inside of the rules and in the worst case scenario just use json there.

olksdr (changes requested):
hey @ottramst! Thank you for opening this PR and contributing your code to the project. 

It is a very good start! 

About the provider: I try to eliminate usage of raw json in our provider as much as possible and relay more on TF standard model - this will hide the complexity of writing and checking json by users and also provide better way to validate the schema by TF.

My proposal is to create a proper schema for this resource, trying to describe each element  by using TF defined types.  As mentioned above, this will enforce proper checks and remove users headaches with more json formatting, checking, etc.

internal/elasticsearch/security/role_mapping.go
ottramst: Sure thing, will try to add tomorrow/day after.

internal/elasticsearch/security/role_mapping.go
ottramst: Not sure how this will work... Rules can be pretty complex, with types nested inside types. i.e
```
{
  "all": [
    {
      "field": {
        "username": "*"
      }
    },
    {
      "except": {
        "any": [
          {
            "field": {
              "username": "*"
            }
          },
          {
            "all": [
              {
                "field": {
                  "username": "*"
                }
              }
            ]
          }
        ]
      }
    },
    {
      "any": [
        {
          "field": {
            "username": "*"
          }
        }
      ]
    }
  ]
}
```

Because of my limited knowledge about TF providers, I'm not 100% sure how to write code that can validate this.

olksdr: Yeah, you're right. We still need to make some tradeoffs here. 
At least we should research this way if it's even possible and what it would take to make it work. If there is too much complexity we definitely will need to fall back to plain json.. 

Thanks a lot of working on this!

Andyandpandy:
Any progress on this? @ottramst @olksdr

Andyandpandy:
Just gonna do another poke here, since it's over 2 months ago I last poked. I'm very keen on getting this as a feature.
@ottramst @olksdr

fludo (approved):

fludo:
Any news on this? @olksdr Seems that validation review is missing. @ottramst  ?

ottramst:
Hey, guys!

Sorry for non-responsiveness on this issue. Unfortunately, because of my busy schedule, I can not continue with this pull request. Anyone is more than welcome to take over from me, but I just can't find the time for this right now.

Again, sorry..

PR #72, opened 2022-02-17
Bump provider support to ES version `8.x`

olksdr:
In this PR:
* change cluster settings in examples and tests to make them be correct for both `7.x` and `8.x`
* change `data_stream` section of `index_template`
  * add  a new field `allow_custom_routing`, which appeared only in `8.x`
  * make sure the index template still works with both `7.x` and `8.x`
* bump test ES cluster to `8.0.0` so the all new resources will be defaulting to newest releases of Elasticsearch


internal/elasticsearch/index/template.go
```diff
@@ -195,10 +201,39 @@ func resourceIndexTemplatePut(ctx context.Context, d *schema.ResourceData, meta
...
-		// only one definition of stream allowed
-		stream := v.([]interface{})[0].(map[string]interface{})
-		indexTemplate.DataStream = stream
+	if d.HasChange("data_stream") {
+		old, new := d.GetChange("data_stream")
+		log.Printf("[TRACE] old data stream settings <%v> new settings <%v>", old, new)
+
+		// 8.x workaround
+		hasAllowCustomRouting := false
+		//for _, s := range []interface{}{old} {
```
JoshMock: nit: can we remove this?

internal/elasticsearch/index/template.go
```diff
@@ -195,10 +201,39 @@ func resourceIndexTemplatePut(ctx context.Context, d *schema.ResourceData, meta
...
+		if old != nil && len(old.([]interface{})) == 1 {
+			if old.([]interface{})[0] != nil {
+				setting := old.([]interface{})[0].(map[string]interface{})
+				if _, ok := setting["allow_custom_routing"]; ok {
+					hasAllowCustomRouting = true
+				}
+			}
+		}
+		//	}
```
JoshMock: cleanup

JoshMock (approved):
All seems relatively straightforward :+1:

internal/elasticsearch/index/template.go
olksdr: yep,  done in [ec5e137](https://github.com/elastic/terraform-provider-elasticstack/pull/72/commits/ec5e13731c807656bb3cfed5dc43f96bb2763ea9)

Crazybus (approved):
LGTM!

> make sure the index template still works with both 7.x and 8.x

Is the plan to also add CI tests for 7.x? Right now everything works but I can imagine future changes that might accidentally break 7.x support accidentally.

olksdr:
> Is the plan to also add CI tests for 7.x? Right now everything works but I can imagine future changes that might accidentally break 7.x support accidentally.

Yep, as a followup I want to create something like matrix testing for different ES versions.

PR #76, opened 2022-02-23
Convert CI job to the `matrix` one for acceptance testing

olksdr:
I'm trying to follow docs from https://www.jenkins.io/doc/book/pipeline/syntax/#declarative-matrix to create a matrix job to test this provider with different versions of the ES 

This is the initial PR and just adds 2 ES version `'8.0.0'` and `'7.15.1'`. I just want to make sure the job does what it's supposed to do and later I also plan to run different tests depending on the version major version `8.x` or `7.x` 

Looking great: 
image

### DIFF
--- a/.ci/pipelines/acceptance.groovy
+++ b/.ci/pipelines/acceptance.groovy
@@ -1,30 +1,43 @@
 @Library('estc') _
 
 pipeline {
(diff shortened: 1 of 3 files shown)


Crazybus (approved):
LGTM!

Issue #91, opened 2022-03-16
[Bug] Data Stream settings are not persisting in index template


poiriermike:
**Describe the bug**
`data_stream` setting for `elasticstack_elasticsearch_index_template` is not persisting on subsequent runs

**To Reproduce**
1. Create an index template
2. Modify index template settings (tested by modifying priority
3. Re-run terraform plan (will show changes, but nothing about data stream)
4. Run terraform apply. Will remove the data stream property from the ojbect

**Expected behavior**
Data stream settings are preserved on update

**Debug output**
Run `terraform` command with `TF_LOG=trace` and provide extended information on TF operations.

Logs https://gist.github.com/poiriermike/d34ef8dda376fa245337c0b2c1a2c0c0

**Screenshots**

![Screen Shot 2022-03-16 at 3 35 46 PM](https://user-images.githubusercontent.com/9126871/158702641-078f6342-67a0-418e-bc65-c93f58b1db82.png)

![Screen Shot 2022-03-16 at 3 34 41 PM](https://user-images.githubusercontent.com/9126871/158702663-0324cdff-c849-4d99-bfdd-63d8b720c49c.png)

**Versions (please complete the following information):**
 - OS: [e.g. Linux]
 - Terraform Version [1.1.7]
 - Provider version [0.3.2]
 - Elasticsearch Version [8.0.1]

**Additional context**

olksdr:
New version of the provider is released https://github.com/elastic/terraform-provider-elasticstack/releases/tag/v0.3.3

PR #95, opened 2022-03-21
Use data_stream setting if present

olksdr:
This changes will make sure that we always use data_stream configuration if it's present. 
The options which data_stream object can contain will be set accordingly as well.

fix https://github.com/elastic/terraform-provider-elasticstack/issues/91


Crazybus (approved):
LGTM!

Issue #116, opened 2022-06-09
[Bug] Terraform does not import Snapshot Repository name when importing the resource


ignuki:
When importing an existing `elasticstack_elasticsearch_snapshot_repository` resource, Terraform does not import the name field of the existing repository, which then forces a replacement of the current resource since the name field is required for that resource.

**To Reproduce**
1. Import an existing snapshot repository.
2. Run `terraform apply` without changing anything in the configuration.
3. Terraform will replace the existing resource.

**Expected behavior**
Terraform should import the name of the repository.

**Versions (please complete the following information):**
 - OS: Debian 11
 - Terraform Version 1.2.2
 - Provider version 0.3.3
 - Elasticsearch Version 7.13.4

PR #117, opened 2022-06-09
Fix snapshot repository resource import 

ignuki:
Naive fix for #116


tobio:
Fixed in https://github.com/elastic/terraform-provider-elasticstack/pull/187

PR #118, opened 2022-06-09
[ci] Extends the list of tested versions

thbkrkr:
This commit extends the list of tested versions. I started with the `7.11.x` based on https://www.elastic.co/support/eol.


olksdr:
hey @thbkrkr,
please merge `main` in and let's see if the CI passes this time

olksdr:
The failing test might have to be updated to incorporate this difference and make it pass.

olksdr (approved):

olksdr (changes requested):
Ah, one more thing, could you, please, run `make gen` to re-generate the docs, since you added change to the description into the `media_type` field.  

And also it might be a good idea to add the record to `CHANGELOG` file

Issue #119, opened 2022-06-09
[Bug] unable to create index with analyzer and normalizer in setting sections


sangeethdba:
**Describe the bug**
unable to create index with analyzer and normalizer in setting sections

**Expected behavior**
need an example of index creation which has custom normalizer and analyzer define

**Versions (please complete the following information):**
 - OS: [e.g. Linux]
 - Terraform Version [e.g. 1.0.0]
 - Provider version [e.g. v0.1.0]
 - Elasticsearch Version [e.g. 7.16.0]

**Additional context**

sangeethdba:
settings {
    setting {
      name  = "index.number_of_shards"
      value = "1"
    }
    setting {
      name  = "index.number_of_replicas"
      value = "1"
    }
    setting {
      name  = "index.search.idle.after"
      value = "20s"
    }
     analysis {
normalizer {
   lowercase_normalizer = {
      "type" = "custom",
      "filter" = [
        "lowercase", "asciifolding"
      ]
    }

sangeethdba:
╷
│ Error: Unsupported block type
│
│   on test_1.tf line 712, in resource "elasticstack_elasticsearch_index" "idx_qa_7":
│  712:      analysis {
│
│ Blocks of type "analysis" are not expected here.

tobio:
Fixed as part of https://github.com/elastic/terraform-provider-elasticstack/pull/137

Issue #108, opened 2022-05-06
[Bug] Not able to set array as value in index settings


bruno-s-coelho:
**Describe the bug**
We need to create an index with the following settings:
```
"settings": {
        "index": {
            "number_of_shards": "3",
            "analysis": {
                "normalizer": {
                    "keyword_normalizer": {
                        "type": "custom",
                        "char_filter": [],
                        "filter": [
                            "lowercase",
                            "asciifolding"
                        ]
                    }
                }
            }
        }
    },
```

When we are defining settings, the provider is just accepting string values.

**To Reproduce**
The configuration used was the following one:
```
resource "elasticstack_elasticsearch_index" "my_index" {
  name = "my-index"

  alias {
    name = "my_alias_1"
  }

  alias {
    name = "my_alias_2"
    filter = jsonencode({
      term = { "user.id" = "developer" }
    })
  }

  mappings = jsonencode({
    properties = {
      field1 = { type = "keyword" }
    }
  })

  settings {
    setting {
      name  = "index.number_of_shards"
      value = "1"
    }
    setting {
      name  = "index.number_of_replicas"
      value = "2"
    }
    setting {
      name  = "index.analysis.normalizer.keyword_normalizer.filter"
      value = "['lowercase', 'asciifolding']"
    }
  }
}
```

When terraform apply is executed, the output is:

Failed with:
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Custom
Analyzer [keyword_normalizer] failed to find filter under name
[[\"lowercase\"]"}],"type":"illegal_argument_exception","reason":"Custom
Analyzer [keyword_normalizer] failed to find filter under name
[[\"lowercase\"]"},"status":400}

**Expected behavior**
We were expecting to have a new index with the alias and the settings as well created.

**Debug output**
```
 Error: Unable to create index: my-index
│ 
│   with elasticstack_elasticsearch_index.my_index,
│   on indices.tf line 1, in resource "elasticstack_elasticsearch_index" "my_index":
│    1: resource "elasticstack_elasticsearch_index" "my_index" {
│ 
│ Failed with: {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Custom Analyzer [keyword_normalizer] failed to find
│ filter under name [['lowercase']"}],"type":"illegal_argument_exception","reason":"Custom Analyzer [keyword_normalizer] failed to find
│ filter under name [['lowercase']"},"status":400}
```

**Screenshots**
[...]

PR #137, opened 2022-09-10
Improve index resource usability by adding individual setting fields

### LINKED ISSUE #124: [Feature] Support index setting with json format
This issue is a feature request to support the JSON format setting definition, as the index mapping definition supports.

I'm trying to use Elasticsearch as our search engine and we have massive amount of analysis settings on the index. 
But since the current provider doesn't support the json format settings, I needed to write everything in terraform format, which was very hard to read and struggled to convert json formatted data to this specific format.

And I believe this is causing lots of issues like the below.
- #108 
- #119 

I'm more than happy to contribute if this feature looks good to the community. Thanks!

k-yomo:
Resolves #124 

As mentioned in #124, current `settings` field's usability is low due to the limitation of type(string object with no nest) and causing issues like #108 and #119 .

This PR introduces individual fields to improve usability by allowing json for some fields like `analyzer` and robustness by having strict key and value type for each setting key.

### NOTE
- This PR is backward compatible, so the `settings` field is deprecated but still usable until we remove it (in  next major version update?).
- The schema is pretty much same as `phillbaker/terraform-provider-elasticsearch`'s [index resource](https://github.com/phillbaker/terraform-provider-elasticsearch/blob/master/es/resource_elasticsearch_index.go).  So it also makes it easier to migrate from `phillbaker/terraform-provider-elasticsearch`.

### DIFF
--- a/internal/elasticsearch/index/index.go
+++ b/internal/elasticsearch/index/index.go
@@ -119,10 +119,11 @@ func ResourceIndex() *schema.Resource {
 			Optional:    true,
 		},
 		"codec": {
-			Type:        schema.TypeString,
-			Description: "The `default` value compresses stored data with LZ4 compression, but this can be set to `best_compression` which uses DEFLATE for a higher compression ratio. This can be set only on creation.",
-			ForceNew:    true,
-			Optional:    true,
+			Type:         schema.TypeString,
+			Description:  "The `default` value compresses stored data with LZ4 compression, but this can be set to `best_compression` which uses DEFLATE for a higher compression ratio. This can be set only on creation.",
+			ForceNew:     true,
+			Optional:     true,
+			ValidateFunc: validation.StringInSlice([]string{"best_compression"}, false),
 		},
 		"routing_partition_size": {
 			Type:        schema.TypeInt,
@@ -137,11 +138,11 @@ func ResourceIndex() *schema.Resource {
 			Optional:    true,
 		},
 		"shard_check_on_startup": {
-			Type:        schema.TypeString,
-			Elem:        &schema.Schema{Type: schema.TypeString},
-			Description: "Whether or not shards should be checked for corruption before opening. When corruption is detected, it will prevent the shard from being opened. Accepts `false`, `true`, `checksum`.",
-			ForceNew:    true,
-			Optional:    true,
+			Type:         schema.TypeString,
+			Description:  "Whether or not shards should be checked for corruption before opening. When corruption is detected, it will prevent the shard from being opened. Accepts `false`, `true`, `checksum`.",
+			ForceNew:     true,
+			Optional:     true,
+			ValidateFunc: validation.StringInSlice([]string{"false", "true", "checksum"}, false),
 		},
 		"sort_field": {
 			Type:        schema.TypeSet,
@@ -150,6 +151,7 @@ func ResourceIndex() *schema.Resource {
 			ForceNew:    true,
 			Optional:    true,
 		},
+		// sort_order can't be set type since it can have dup strings like ["asc", "asc"]
 		"sort_order": {
 			Type:        schema.TypeList,
 			Elem:        &schema.Schema{Type: schema.TypeString},
@@ -338,9 +340,10 @@ func ResourceIndex() *schema.Resource {
 			Optional:    true,
 		},
 		"search_slowlog_level": {
-			Type:        schema.TypeString,
-			Description: "Set which logging level to use for the search slow log, can be: `warn`, `info`, `debug`, `trace`",
-			Optional:    true,
+			Type:         schema.TypeString,
+			Description:  "Set which logging level to use for the search slow log, can be: `warn`, `info`, `debug`, `trace`",
+			Optional:     true,
+			ValidateFunc: validation.StringInSlice([]string{"warn", "info", "debug", "trace"}, false),
 		},
 		"indexing_slowlog_threshold_index_warn": {
 			Type:        schema.TypeString,
@@ -363,9 +366,10 @@ func ResourceIndex() *schema.Resource {
 			Optional:    true,
 		},
 		"indexing_slowlog_level": {
-			Type:        schema.TypeString,
-			Description: "Set which logging level to use for the search slow log, can be: `warn`, `info`, `debug`, `trace`",
-			Optional:    true,
+			Type:         schema.TypeString,
+			Description:  "Set which logging level to use for the search slow log, can be: `warn`, `info`, `debug`, `trace`",
+			Optional:     true,
+			ValidateFunc: validation.StringInSlice([]string{"warn", "info", "debug", "trace"}, false),
 		},
 		"indexing_slowlog_source": {
 			Type:        schema.TypeString,
@@ -543,20 +547,20 @@ If specified, this mapping can include: field names, [field data types](https://
 						} else if v, ok := index.Settings["index."+key]; ok {
 							value = v
 						} else {
-							tflog.Warn(ctx, fmt.Sprintf("setting '%s' is not currently managed by terraform provider and inored", key))
+							tflog.Warn(ctx, fmt.Sprintf("setting '%s' is not currently managed by terraform provider and has been ignored", key))
 							continue
 						}
 						switch typ {
 						case schema.TypeInt:
 							v, err := strconv.Atoi(value.(string))
 							if err != nil {
-								return nil, fmt.Errorf("failed to convert setting '%s' value %v to int: %v", key, value, err)
+								return nil, fmt.Errorf("failed to convert setting '%s' value %v to int: %w", key, value, err)
 							}
 							value = v
 						case schema.TypeBool:
 							v, err := strconv.ParseBool(value.(string))
 							if err != nil {
-								return nil, fmt.Errorf("failed to convert setting '%s' value %v to bool: %v", key, value, err)
+								return nil, fmt.Errorf("failed to convert setting '%s' value %v to bool: %w", key, value, err)
 							}
 							value = v
 						}
@@ -720,7 +724,7 @@ func resourceIndexCreate(ctx context.Context, d *schema.ResourceData, meta inter
 			setting := s.(map[string]interface{})
 			name := setting["name"].(string)
 			if _, ok := index.Settings[name]; ok {
-				return diag.FromErr(fmt.Errorf("setting '%s' is already defined by the other field, please remove it from `settings` not to produce unexpected settings", name))
+				return diag.FromErr(fmt.Errorf("setting '%s' is already defined by the other field, please remove it from `settings` to avoid unexpected settings", name))
 			}
 			index.Settings[name] = setting["value"]
 		}
@@ -802,7 +806,7 @@ func resourceIndexUpdate(ctx context.Context, d *schema.ResourceData, meta inter
 		}
 		for k, v := range ns {
 			if _, ok := updatedSettings[k]; ok && v != nil {
-				return diag.FromErr(fmt.Errorf("setting '%s' is already updated by the other field, please remove it from `settings` not to produce unexpected settings", k))
+				return diag.FromErr(fmt.Errorf("setting '%s' is already updated by the other field, please remove it from `settings` to avoid unexpected settings", k))
 			} else {
 				updatedSettings[k] = v
 			}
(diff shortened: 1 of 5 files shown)


k-yomo:
@tobio 
Could I ask you to review if the concept of this PR looks good?

tobio:
@k-yomo yep, I agree this is the right approach to take.

What do you think about keeping the general `settings` field and merging those with the individual settings?

It looks like your code's already doing that, and it would mean the provider was flexible in supporting new index settings before those individual fields were added.

k-yomo:
> What do you think about keeping the general settings field and merging those with the individual settings?
> 
> It looks like your code's already doing that, and it would mean the provider was flexible in supporting new index settings before those individual fields were added.

@tobio 
Totally agree! It would support smooth migration from `settings`. 
And yes, this PR's code is doing merge unless they are conflicting.


Thank you for the review! I'll proceed to complete the PR🙏

k-yomo:
@tobio 
This PR is also ready for review! 

While developing, I noticed it will be problematic to continue to have both individual settings and `settings` fields. Since it's gonna be confusing when importing, refreshing, and updating.   In fact, currently, settings are not really refreshed because of it.

Thus I believe `settings` should be removed in the future, and I put a deprecated flag and comments. But let me know if you or the other maintainers have a different opinion.
We can also deprecate settings later if it's controversial.

internal/elasticsearch/index/index.go
```diff
@@ -37,6 +105,301 @@ func ResourceIndex() *schema.Resource {
...
+		},
+		"load_fixed_bitset_filters_eagerly": {
+			Type:        schema.TypeBool,
+			Description: "Indicates whether cached filters are pre-loaded for nested queries. This can be set only on creation.",
+			ForceNew:    true,
+			Optional:    true,
+		},
+		"shard_check_on_startup": {
+			Type:        schema.TypeString,
+			Elem:        &schema.Schema{Type: schema.TypeString},
```
```suggestion
```

Looks like this can be removed.

internal/elasticsearch/index/index.go
```diff
@@ -37,6 +105,301 @@ func ResourceIndex() *schema.Resource {
...
+		},
+		"sort_field": {
+			Type:        schema.TypeSet,
+			Elem:        &schema.Schema{Type: schema.TypeString},
+			Description: "The field to sort shards in this index by.",
+			ForceNew:    true,
+			Optional:    true,
+		},
+		"sort_order": {
+			Type:        schema.TypeList,
```
tobio, line 156: nit: `sort_field` and `sort_order` should probably have the same iterable type. If this is intentional can we add a comment here about why one needs to be a list and one a set.

internal/elasticsearch/index/index.go
```diff
@@ -37,6 +105,301 @@ func ResourceIndex() *schema.Resource {
...
+			Elem:        &schema.Schema{Type: schema.TypeString},
+			Description: "The direction to sort shards in. Accepts `asc`, `desc`.",
```
tobio, line 158: It looks like a `StringInSlice` validation would make sense here.

internal/elasticsearch/index/index.go
```diff
@@ -37,6 +105,301 @@ func ResourceIndex() *schema.Resource {
...
+		"load_fixed_bitset_filters_eagerly": {
+			Type:        schema.TypeBool,
+			Description: "Indicates whether cached filters are pre-loaded for nested queries. This can be set only on creation.",
+			ForceNew:    true,
+			Optional:    true,
+		},
+		"shard_check_on_startup": {
+			Type:        schema.TypeString,
+			Elem:        &schema.Schema{Type: schema.TypeString},
+			Description: "Whether or not shards should be checked for corruption before opening. When corruption is detected, it will prevent the shard from being opened. Accepts `false`, `true`, `checksum`.",
```
tobio: `StringInSlice`?

internal/elasticsearch/index/index.go
```diff
@@ -37,6 +105,301 @@ func ResourceIndex() *schema.Resource {
...
+		},
+		"number_of_routing_shards": {
+			Type:        schema.TypeInt,
+			Description: "Value used with number_of_shards to route documents to a primary shard. This can be set only on creation.",
+			ForceNew:    true,
+			Optional:    true,
+		},
+		"codec": {
+			Type:        schema.TypeString,
+			Description: "The `default` value compresses stored data with LZ4 compression, but this can be set to `best_compression` which uses DEFLATE for a higher compression ratio. This can be set only on creation.",
```
tobio: `StringInSlice`?

internal/elasticsearch/index/index.go
```diff
@@ -37,6 +105,301 @@ func ResourceIndex() *schema.Resource {
...
+			Optional:    true,
+		},
+		"search_slowlog_threshold_fetch_trace": {
+			Type:        schema.TypeString,
+			Description: "Set the cutoff for shard level slow search logging of slow searches in the fetch phase, in time units, e.g. `500ms`",
+			Optional:    true,
+		},
+		"search_slowlog_level": {
+			Type:        schema.TypeString,
+			Description: "Set which logging level to use for the search slow log, can be: `warn`, `info`, `debug`, `trace`",
```
tobio: `StringInSlice`?

internal/elasticsearch/index/index.go
```diff
@@ -37,6 +105,301 @@ func ResourceIndex() *schema.Resource {
...
+			Optional:    true,
+		},
+		"indexing_slowlog_threshold_index_trace": {
+			Type:        schema.TypeString,
+			Description: "Set the cutoff for shard level slow search logging of slow searches for indexing queries, in time units, e.g. `500ms`",
+			Optional:    true,
+		},
+		"indexing_slowlog_level": {
+			Type:        schema.TypeString,
+			Description: "Set which logging level to use for the search slow log, can be: `warn`, `info`, `debug`, `trace`",
```
tobio: `StringInSlice`

internal/elasticsearch/index/index.go
```diff
@@ -176,23 +533,36 @@ If specified, this mapping can include: field names, [field data types](https://
...
-					for k, v := range index.Settings {
-						if _, ok := ignoredDefaults[k]; ok {
+					for key, typ := range allSettingsKeys {
+						var value interface{}
+						if v, ok := index.Settings[key]; ok {
+							value = v
+						} else if v, ok := index.Settings["index."+key]; ok {
+							value = v
+						} else {
+							tflog.Warn(ctx, fmt.Sprintf("setting '%s' is not currently managed by terraform provider and inored", key))
```
```suggestion
							tflog.Warn(ctx, fmt.Sprintf("setting '%s' is not currently managed by terraform provider and has been ignored", key))
```

internal/elasticsearch/index/index.go
```diff
@@ -176,23 +533,36 @@ If specified, this mapping can include: field names, [field data types](https://
...
-					}
-					settings["setting"] = result
-
-					if err := d.Set("settings", []interface{}{settings}); err != nil {
-						return nil, err
+						switch typ {
+						case schema.TypeInt:
+							v, err := strconv.Atoi(value.(string))
+							if err != nil {
+								return nil, fmt.Errorf("failed to convert setting '%s' value %v to int: %v", key, value, err)
```
```suggestion
								return nil, fmt.Errorf("failed to convert setting '%s' value %v to int: %w", key, value, err)
```

We should probably be wrapping the error here.

internal/elasticsearch/index/index.go
```diff
@@ -176,23 +533,36 @@ If specified, this mapping can include: field names, [field data types](https://
...
+							}
+							value = v
+						case schema.TypeBool:
+							v, err := strconv.ParseBool(value.(string))
+							if err != nil {
+								return nil, fmt.Errorf("failed to convert setting '%s' value %v to bool: %v", key, value, err)
```
```suggestion
								return nil, fmt.Errorf("failed to convert setting '%s' value %v to bool: %w", key, value, err)
```

Similarly

internal/elasticsearch/index/index.go
```diff
@@ -297,15 +667,63 @@ func resourceIndexCreate(ctx context.Context, d *schema.ResourceData, meta inter
...
-		managed_settings := v.([]interface{})[0].(map[string]interface{})["setting"].(*schema.Set)
-		sets := make(map[string]interface{}, managed_settings.Len())
-		for _, s := range managed_settings.List() {
+		managedSettings := v.([]interface{})[0].(map[string]interface{})["setting"].(*schema.Set)
+		for _, s := range managedSettings.List() {
 			setting := s.(map[string]interface{})
-			sets[setting["name"].(string)] = setting["value"]
+			name := setting["name"].(string)
+			if _, ok := index.Settings[name]; ok {
+				return diag.FromErr(fmt.Errorf("setting '%s' is already defined by the other field, please remove it from `settings` not to produce unexpected settings", name))
```
```suggestion
				return diag.FromErr(fmt.Errorf("setting '%s' is already defined by the other field, please remove it from `settings` to avoid unexpected settings", name))
```

internal/elasticsearch/index/index.go
```diff
@@ -375,7 +800,17 @@ func resourceIndexUpdate(ctx context.Context, d *schema.ResourceData, meta inter
 				delete(ns, k)
 			}
 		}
-		if diags := client.UpdateElasticsearchIndexSettings(ctx, indexName, ns); diags.HasError() {
+		for k, v := range ns {
+			if _, ok := updatedSettings[k]; ok && v != nil {
+				return diag.FromErr(fmt.Errorf("setting '%s' is already updated by the other field, please remove it from `settings` not to produce unexpected settings", k))
```
```suggestion
				return diag.FromErr(fmt.Errorf("setting '%s' is already updated by the other field, please remove it from `settings` to avoid unexpected settings", k))
```

internal/elasticsearch/index/index_test.go
```diff
@@ -44,6 +44,99 @@ func TestAccResourceIndex(t *testing.T) {
...
+	indexName := sdkacctest.RandStringFromCharSet(22, sdkacctest.CharSetAlphaNum)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { acctest.PreCheck(t) },
+		CheckDestroy:      checkResourceIndexDestroy,
+		ProviderFactories: acctest.Providers,
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccResourceIndexSettingsConflict(indexName),
+				ExpectError: regexp.MustCompile("setting 'number_of_shards' is already defined by the other field, please remove it from `settings` not to produce unexpected settings"),
```
tobio:
```suggestion
				ExpectError: regexp.MustCompile("setting 'number_of_shards' is already defined by the other field, please remove it from `settings` to avoid unexpected settings"),
```

internal/elasticsearch/index/index.go
```diff
@@ -448,6 +883,8 @@ func resourceIndexRead(ctx context.Context, d *schema.ResourceData, meta interfa
 			return diag.FromErr(err)
 		}
 	}
+	// TODO: We ideally should set read settings to each field to detect changes
+	// But for now, setting it will cause unexpected diff for the existing clients which use `settings`
```
tobio, line 891: I haven't thought this through entirely, but I expect we'd be able to handle this with `DiffSuppressFunc`. So suppress the diff of individual settings when they're unset, but set in the generic `settings` field, we could also do the inverse and suppress the diff of generic `settings` which only contain settings managed by the individual settings fields.

tobio (commented):
Looking good, coupe of comments and some text updates.

internal/elasticsearch/index/index.go
k-yomo: `StringInSlice` is a validation for `schema.TypeString` and also it seems validation can't be used for list...!
https://github.com/hashicorp/terraform-plugin-sdk/blob/44ccfdce81e30da9672c731270773a2510772796/helper/schema/schema.go#L348-L349

internal/elasticsearch/index/index.go
k-yomo: Yes this is intentional because when `sort_field` has multiple fields, `sort_order` can have dup strings like `["asc", "asc"]`.
Let me add a comment!

internal/elasticsearch/index/index.go
k-yomo: Oh I see! Yeah it seems to work.  Let me try

k-yomo: > // This is valid only for attributes of primitive types, because
   // DiffSuppressFunc itself is only compatible with primitive types.

I found that `DiffSuppressFunc` can only be used for primitive types, so it can't be used..!

tobio:
jenkins test this

tobio (approved):

PR #140, opened 2022-09-12
Parallelize acceptance tests

k-yomo:
This PR updates acceptance tests to use `ParallelTest` to be able to run in parallel and reduce time.

### DIFF
--- a/internal/elasticsearch/cluster/settings_test.go
+++ b/internal/elasticsearch/cluster/settings_test.go
@@ -12,7 +12,7 @@ import (
 )
 
 func TestAccResourceClusterSettings(t *testing.T) {
-	resource.UnitTest(t, resource.TestCase{
+	resource.ParallelTest(t, resource.TestCase{
 		PreCheck:          func() { acctest.PreCheck(t) },
 		CheckDestroy:      checkResourceClusterSettingsDestroy,
 		ProviderFactories: acctest.Providers,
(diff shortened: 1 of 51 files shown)


internal/elasticsearch/cluster/settings_test.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 func TestAccResourceClusterSettings(t *testing.T) {
-	resource.UnitTest(t, resource.TestCase{
```
k-yomo, line 15: > // UnitTest is a helper to force the acceptance testing harness to run in the
// normal unit test suite. This should only be used for resource that don't
// have any external dependencies.
https://github.com/hashicorp/terraform-plugin-sdk/blob/44ccfdce81e30da9672c731270773a2510772796/helper/resource/testing.go#L755-L765

`UnitTest` is inappropriate because the resources depend on elasticsearch.

tobio:
jenkins test this

k-yomo:
fmm, the total time on Jenkins is not really improved...
some of the tests are failing with `circuit_breaking_exception`, so the concurrent count(`10`) might be too high.

tobio:
> fmm, the total time on Jenkins is not really improved...

Taking a quick look at the Jenkins build times. It's taking ~3m to build the test image and then ~1m to run the tests. I'm totally happy to accept changes which improve the build time, not trying to throw up a roadblock on improvements here. Given the fairly small gains seen with this PR and the failures that have been introduced I'd strongly err towards stability over potentially increased flakiness though.

k-yomo:
Agree! let me close this PR.

PR #141, opened 2022-09-12
Debug log each req/res of API call

k-yomo:
This PR introduces automatic debug log dump of each req/res instead of manually adding trace logs.
Also `log.Printf` is [legacy](https://www.terraform.io/plugin/log/writing) now, so it's replaced with `tflog`.

It would make it easier to debug an issue.
(e.g. below is the issue in a provider which uses the same debug transport.
https://github.com/k-yomo/terraform-provider-algolia/issues/79


tobio:
jenkins test this

tobio (approved):

PR #143, opened 2022-09-12
Ignore user's metadata with underscore prefix

### LINKED ISSUE #104: Add new resource to manage system user passwords
System users cannot be managed by the existing `elasticstack_elasticsearch_security_user` resource. These users are not editable, except to change their password via the [change password](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-change-password.html). 

I think the simplest solution is to have a dedicated resource to manage the password on system users, updating the existing resource to error when used against a system user. This has a couple of benefits:
* It's clear from the schema what fields are able to be edited
* Any attempt to manage other fields in system users will result in an error which should direct consumers to the correct resource type 
 
---

https://github.com/elastic/terraform-provider-elasticstack/issues/66

**Describe the bug**
The `kibana_system` user has metadata `_reserved` starting with an underscore.  The elastic api does not let us set or modify metadata starting with underscore.

If I import a `kibana_system` user into terraform, it also loads the metadata.  When I then try to update the user, for example changing the password, the request fails because the validation "metadata cannot start with underscore".

We should make sure that if the `metadata` attribute is not set or set to null, terraform won't try to update it.

**To Reproduce**

1. Create a user resource

```
resource "elasticstack_elasticsearch_security_user" "kibana_system" {
    username = "kibana_system"
    password = "changeme"
    roles    = ["kibana_system"]
}
```

and import from your cluster.
2. Change the password and apply.
3. Update fails because of metadata update validation.

**Expected behavior**

Do not try to update metadata if my resource does not ask for it.

**Debug output**

```
Terraform will perform the following actions:

  # elasticstack_elasticsearch_security_user.kibana_system will be updated in-place
  ~ resource "elasticstack_elasticsearch_security_user" "kibana_system" {
      + email     = ""
      + full_name = ""
        id        = "KrzMC_O7QJa0DkIaO7RRYA/kibana_system"
      + password  = (sensitive value)
        # (4 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

elasticstack_elasticsearch_security_user.kibana_system: Modifying... [id=KrzMC_O7QJa0DkIaO7RRYA/kibana_system]
╷
│ Error: Unable to create or update a user
│
│   with elasticstack_elasticsearch_security_user.kibana_system,
│   on main.tf line 52, in resource "elasticstack_elasticsearch_security_user" "kibana_system":
│   52: resource "elasticstack_elasticsearch_security_user" "kibana_system" {
│
│ Failed with: {"error":{"root_cause":[{"type":"action_request_validation_exception","reason":"Validation Failed: 1: metadata keys may not start with
│ [_];"}],"type":"action_request_validation_exception","reason":"Validation Failed: 1: metadata keys may not start with [_];"},"status":400}
╵
```

 - OS: linux
 - Terraform Version 1.1.9
 - Provider version 0.3.3
 - Elasticsearch Version 8.1.3

k-yomo:
Fixes #104 by ignoring metadata starting with `_` with a warning log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 ## [Unreleased]
 ### Added
 - Allow use of `api_key` instead of `username`/`password` for authentication ([#130](https://github.com/elastic/terraform-provider-elasticstack/pull/130))
+- Add debug log of eac req/res ([#141](https://github.com/elastic/terraform-provider-elasticstack/pull/141))
 ### Fixed
-- Upgrade Go version to 1.19 and sdk version to v2.22.0 ([#138](https://github.com/elastic/terraform-provider-elasticstack/pull/139))
+- Ignore user's metadata with underscore prefix ([#143](https://github.com/elastic/terraform-provider-elasticstack/pull/143))
+- Expose provider package ([#142](https://github.com/elastic/terraform-provider-elasticstack/pull/142))
+- Upgrade Go version to 1.19 and sdk version to v2.22.0 ([#139](https://github.com/elastic/terraform-provider-elasticstack/pull/139))
 - Make API calls context aware to be able to handle timeouts ([#138](https://github.com/elastic/terraform-provider-elasticstack/pull/138))
 - Correctly identify a missing security user ([#101](https://github.com/elastic/terraform-provider-elasticstack/issues/101))
 - Support **7.x** Elasticsearch < **7.15** by removing the default `media_type` attribute in the Append processor ([#118](https://github.com/elastic/terraform-provider-elasticstack/pull/118))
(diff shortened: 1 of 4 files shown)


k-yomo:
@tobio 
Could I ask you to review the PR?🙏 (Also thank you for reviewing and kicking CI all the time)

tobio:
jenkins test this

tobio:
https://github.com/elastic/terraform-provider-elasticstack/issues/104#issuecomment-1245891584

I'm going to close this PR out, since it won't address the core issue with managing system users with the existing resource.

k-yomo:
Oh I see, I saw your comment in the issue and it makes sense.

tobio:
@k-yomo thanks for all the contributions here this week! There's some really valuable contributions!

Sorry for the confusion on this issue, we need to better curate the backlog here. There's an ongoing effort to get people more consistently working on the two terraform providers, part of that should involve grooming the existing backlog as well.

k-yomo:
@tobio 
Not at all! That's great to hear!

Issue #115, opened 2022-06-06
[Feature] Need support for Logstash Pipeline Terraform Resource in Elastic Cloud


raswinraaj:
**Is your feature request related to a problem? Please describe.**
Currently, the Elasticstack provider has support only for Ingest Pipelines. Need a similar support for Logstash pipelines

**Describe the resource you would like to have implemented.**
Logstash Pipeline

**Describe the solution you'd like**
Need a terraform resource which will help in creating Logstash Pipeline

**Describe alternatives you've considered**
As of now, i am using API call as described in https://www.elastic.co/guide/en/kibana/master/logstash-configuration-management-api-create.html

**Additional context**

tobio:
This has been released with [0.5.0](https://github.com/elastic/terraform-provider-elasticstack/releases/tag/v0.5.0)

PR #151, opened 2022-10-09
[ISSUE-115] Add Logstash Centralised Pipeline Management

RobsonSutton:
- Add Logstash Centralised Pipeline Management feature, raised under https://github.com/elastic/terraform-provider-elasticstack/issues/115 
- ES Logstash API endpoints: https://www.elastic.co/guide/en/elasticsearch/reference/current/logstash-apis.html


RobsonSutton:
@tobio -Please can you take a look at this one when you have a spare sec? Cheers! 👍 

Note: First full feature PR so please do shout if there is a better approach i should take with anything defined or if any parts don't make sense!

tobio:
jenkins test this

internal/elasticsearch/logstash/pipeline_test.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func testAccResourceLogstashPipelineCreate(pipelineID string) string {
+	return fmt.Sprintf(`
+provider "elasticstack" {
+  elasticsearch {}
+}
+
+resource "elasticstack_elasticsearch_logstash_pipeline" "test" {
+  pipeline_id = "%s"
```
tobio, line 67: Would be nice to format the definitions in this file.

internal/utils/utils.go
```diff
@@ -189,3 +190,8 @@ func StringToHash(s string) (*string, error) {
 	hash := fmt.Sprintf("%x", bs)
 	return &hash, nil
 }
+
+func FormatStrictDateTime(t time.Time) string {
+	strictDateTime := fmt.Sprintf("%d-%02d-%02dT%02d:%02d:%02d.%03dZ", t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(), t.Second(), t.Nanosecond())
```
tobio: Should be able to replace this with `.Format`

```suggestion
	strictDateTime := t.Format("2006-01-02T15:04:05.000Z")
```

internal/elasticsearch/logstash/pipeline.go
```diff
@@ -0,0 +1,174 @@
...
+		"pipeline_metadata": {
+			Description: "Optional metadata about the pipeline.",
+			Type:        schema.TypeMap,
+			Optional:    true,
+			Elem: &schema.Schema{
+				Type:    schema.TypeString,
+				Default: nil,
+			},
+		},
+		"pipeline_settings": {
```
tobio: IMO it would be nicer to have explicit schema items for the handful of `pipeline.*` available in [Logstash](https://www.elastic.co/guide/en/logstash/8.4/logstash-settings-file.html). We'd be able to validate individual settings (both keys and values) within the provider rather than simply allow a free form map. 

It would also be consistent with how we're now treating [index](https://github.com/elastic/terraform-provider-elasticstack/pull/137) settings. 

WDYT?

tobio (commented):
Nice addition! Couple of comments in the code.

internal/utils/utils.go
RobsonSutton: This certainly makes life a little easier! 😆 Will update this

internal/elasticsearch/logstash/pipeline.go
RobsonSutton: I actually started attempting to do this before throwing my toys out of the pram and going with the simpler approach, will take another look at this today though to see how to achieve this!

internal/elasticsearch/logstash/pipeline_test.go
RobsonSutton: done 👍

internal/utils/utils.go
RobsonSutton: done 👍

internal/elasticsearch/logstash/pipeline.go
RobsonSutton: Bit more testing and maybe a few tweaks left, nearly got it working, just some weird behaviour surrounding defaults when not defining the pipeline_settings block in the resource. Will take another look tomorrow 👍

tobio:
jenkins test this

internal/elasticsearch/logstash/pipeline.go
```diff
@@ -0,0 +1,249 @@
...
+	settings := make(map[string]interface{})
+	settings["pipeline_workers"] = pipelineSettings.PipelineWorkers
+	settings["pipeline_batch_size"] = pipelineSettings.PipelineBatchSize
+	settings["pipeline_batch_delay"] = pipelineSettings.PipelineBatchDelay
+	settings["queue_type"] = pipelineSettings.QueueType
+	settings["queue_max_bytes_number"] = pipelineSettings.QueueMaxBytesNumber
+	settings["queue_max_bytes_units"] = pipelineSettings.QueueMaxBytesUnits
+	settings["queue_checkpoint_writes"] = pipelineSettings.QueueCheckpointWrites
+
+	return append(make([]interface{}, 0), settings)
```
tobio:
```suggestion
	return []interface{}{ settings }
```

tobio (commented):
Looking good, one small comment on the code but really I just want to test some specifics out here before 🍏

RobsonSutton:
> Looking good, one small comment on the code but really I just want to test some specifics out here before 🍏

Yeah if you are able to test then that would be massively appreciated! 👍 I have been trying to figure out how to define either an empty default or pass through the defaults defined for each setting when the `pipeline_settings` block is not passed to the provider which has me a bit stumped currently. Seems to populate these defaults for the settings when pipeline_settings = {} but when this block doesn't exist I can't seem to figure out how to populate these with each settings default (as setting this is required on the ES API by the looks of it)!

internal/elasticsearch/logstash/pipeline.go
RobsonSutton: Updated 👍

internal/elasticsearch/logstash/pipeline.go
RobsonSutton: @tobio - My brain clearly wasn't working the other day as I was building based on the deprecated approach in index settings... apologies for this! 

I have just updated the approach to follow the newer one in https://github.com/elastic/terraform-provider-elasticstack/pull/151/commits/1a45bda9bb1e71522b5c3f902f0505e7b38c6d38, this has also solved the issue I was having with the defaults not being applied when `pipeline_settings {}` was not set, so win win!

tobio: Exactly what I was looking to test, I'd missed the update here and noticed the code I pulled down looked different to what I remember :). Thanks for this!

tobio:
jenkins test this

internal/elasticsearch/logstash/pipeline.go
```diff
@@ -0,0 +1,259 @@
...
+			Description: "The maximum number of events an individual worker thread collects before executing filters and outputs.",
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Default:     125,
+		},
+		"pipeline_workers": {
+			Description: "The number of parallel workers used to run the filter and output stages of the pipeline.",
+			Type:        schema.TypeInt,
+			Optional:    true,
+			Default:     1,
```
tobio: Logstash will default to the number of cores on the host. We should drop the default here and only specify this if it's set to something other than 0.

internal/elasticsearch/logstash/pipeline.go
```diff
@@ -0,0 +1,259 @@
...
+			Description: "Optional metadata about the pipeline.",
+			Type:        schema.TypeMap,
+			Optional:    true,
+			Elem: &schema.Schema{
+				Type:    schema.TypeString,
+				Default: nil,
+			},
+		},
+		// Pipeline Settings
+		"pipeline_batch_delay": {
```
tobio, line 77: There's a bunch more [settings](https://www.elastic.co/guide/en/logstash/8.4/logstash-settings-file.html) we should look to support. It looks like:

```
pipeline.unsafe_shutdown
pipeline.plugin_classloaders
pipeline.ordered
pipeline.ecs_compatibility
queue.max_events
queue.page_capacity
queue.checkout.acks
queue.checkout.retry
queue.drain
```

I'm happy to add these in a follow up if you're short of time, it would be good to have a more complete group of options though.

internal/elasticsearch/logstash/pipeline.go
```diff
@@ -0,0 +1,259 @@
...
+		return diags
+	}
+	return nil
+}
+
+func expandIndividuallyDefinedPipelineSettings(ctx context.Context, d *schema.ResourceData, settingsKeys map[string]schema.ValueType) map[string]interface{} {
+	settings := make(map[string]interface{})
+	for key := range settingsKeys {
+		tfFieldKey := convertSettingsKeyToTFFieldKey(key)
+		settings[key] = d.Get(tfFieldKey)
```
tobio: With the `pipeline.workers` comment above this should likely be something like
```suggestion
		if setting, ok := d.GetOk(tfFieldKey); ok { 
			settings[key] = setting
		}
```

I wonder if we should instead extract this method into utils and re-use the index implementation instead.

RobsonSutton: Yeah I have no clue why I copied and simplified this, have extracted to utils now though as suggested 👍

internal/elasticsearch/logstash/pipeline.go
RobsonSutton: I was just focusing on the ones visible via the UI initially, I've just updated to include those mentioned too though 👍 Have removed defaults for all to reduce plan noise since all have defaults being set on the API side, happy to re-add these defaults if you think they are required at all though?

internal/elasticsearch/logstash/pipeline.go
RobsonSutton: Updated to validate >=1 for input, is this what you meant? can't think of a case where a user would want workers set to 0 (unless I'm missing something) and as you said this default will be managed on the elasticsearch side 👍

tobio:
jenkins test this

tobio:
jenkins test this

internal/elasticsearch/logstash/pipeline.go
```diff
@@ -0,0 +1,295 @@
...
+	id, diags := client.ID(ctx, pipelineID)
+	if diags.HasError() {
+		return diags
+	}
+
+	var logstashPipeline models.LogstashPipeline
+
+	logstashPipeline.PipelineID = pipelineID
+	logstashPipeline.Description = d.Get("description").(string)
+	logstashPipeline.LastModified = utils.FormatStrictDateTime(time.Now())
```
tobio:
```suggestion
	logstashPipeline.LastModified = utils.FormatStrictDateTime(time.Now().UTC())
```

The Kibana UI shows this last update 'in 11 hours' for me :)

tobio:
jenkins test this

tobio (approved):
LGTM. Thanks for this! Really appreciate the patience through the review :)

RobsonSutton:
> LGTM. Thanks for this! Really appreciate the patience through the review :)

No probs at all! 👍 Thanks for helping throughout the process, am new to Go and provider dev so has been good getting some feedback on better approaches, usage of syntax etc!

Issue #70, opened 2022-02-14
[Feature] Support for API Key Management


BenB196:
**Is your feature request related to a problem? Please describe.**
Managing Elasticsearch API keys is somewhat cumbersome, especially on large scale deployments where you want fine grain permission controls.

**Describe the resource you would like to have implemented.**
Ability to manage/create API keys via Terraform.

**Describe the solution you'd like**
I'd like to be able to specify an API key resource:
- For creation and be able to get back the information needed for using the API key.
- That exists and delete it via terraform.

I'd like to be able to get a list of existing API keys via a data source.

**Describe alternatives you've considered**
You can use the Elasticsearch API via something like curl:
- Create API Key https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html
  - Grant API Key https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-grant-api-key.html
- Delete API Key https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-invalidate-api-key.html
- API Key Data Source https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-get-api-key.html
- API Key List Data Source https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-query-api-key.html

**Additional context**
Security Resources are already partially supported via [user](https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/elasticsearch_security_user) and [role](https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/elasticsearch_security_role)

jitbasemartin:
I started work here: https://github.com/jitbasemartin/terraform-provider-elasticstack/tree/feature/issue-70 but I think it's too much for me :/ 
I will try to continue later but if someone wants to start something else I won't be angry ^^

vladyslav-sokolov:
+1

strelok899:
+1 on this feature

tobio:
This has been released with [0.5.0](https://github.com/elastic/terraform-provider-elasticstack/releases/tag/v0.5.0)

PR #158, opened 2022-10-22
Add missing `analysis_char_filter` field

k-yomo:
Add missing `analysis_char_filter` field to set `char_filter`.
https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-custom-analyzer.html#_configuration


tobio:
jenkins test this

internal/elasticsearch/index/index.go
```diff
@@ -390,6 +390,13 @@ func ResourceIndex() *schema.Resource {
 			ForceNew:     true, // To add a tokenizer, the index must be closed, updated, and then reopened; we can't handle that here.
 			ValidateFunc: validation.StringIsJSON,
 		},
+		"analysis_char_filter": {
+			Type:         schema.TypeString,
+			Description:  "A JSON string describing the char_filters applied to the index.",
+			Optional:     true,
+			ForceNew:     true, // To add a char_filters, the index must be closed, updated, and then reopened; we can't handle that here.
```
tobio: I'm not a big fan of the `ForceNew` here. 

Best case is someone gets to easily recreate a new index they're playing around with, worst case is they delete a large production index. Comparing that to the behaviour if we left `ForceNew: false`, the worst case is they have to manually `POST <index>/_close` and `_open` around the terraform apply. 

IMO there's a pathway for applying this without destroying the index, and destroying the index may be a pretty cataclysmic event. The error message is reasonably clear what the issue is, we should let the operators deal with it as best suits their use-case. 

```
Can't update non dynamic settings [[index.analysis.char_filter.zero_with_spaces.type, index.analysis.char_filter.zero_with_spaces.mappings]] for open indices [[test/9oUEz4DkTIKswrtbwrEWzg]]
```

tobio (commented):
Looks good, IMO we shouldn't be setting `ForceNew` for this setting though.

internal/elasticsearch/index/index_test.go
```diff
@@ -83,6 +83,7 @@ func TestAccResourceIndexSettings(t *testing.T) {
 					resource.TestCheckResourceAttr("elasticstack_elasticsearch_index.test_settings", "routing_rebalance_enable", "primaries"),
 					resource.TestCheckResourceAttr("elasticstack_elasticsearch_index.test_settings", "gc_deletes", "30s"),
 					resource.TestCheckResourceAttr("elasticstack_elasticsearch_index.test_settings", "analysis_analyzer", `{"text_en":{"filter":["lowercase","minimal_english_stemmer"],"tokenizer":"standard","type":"custom"}}`),
```
tobio: This expectation also needs to be updated:

```
index_test.go:50: Step 1/1 error: Check failed: Check 28/32 error: elasticstack_elasticsearch_index.test_settings: Attribute 'analysis_analyzer' expected "{\"text_en\":{\"filter\":[\"lowercase\",\"minimal_english_stemmer\"],\"tokenizer\":\"standard\",\"type\":\"custom\"}}", got "{\"text_en\":{\"char_filter\":\"zero_width_spaces\",\"filter\":[\"lowercase\",\"minimal_english_stemmer\"],\"tokenizer\":\"standard\",\"type\":\"custom\"}}"
```

internal/elasticsearch/index/index.go
k-yomo: I agree.  Not only for this but I think we should have `deletion_protection` like cloud SQL or the other important resources have to prevent unintended deletion.
The `deletion_protection` is `true` by default, and need to update it to `false` when we truly want to delete the index.  With this, we have the flexibility to allow playing around with indices by setting `deletion_protection = false` and also offering a safer way to manage indices for production use cases.
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance

k-yomo: If we think it still would be better not to set `ForceNew`, let me create a separate PR because it should be also removed from the other fields (like analysis_filter or analysis_normalizer).

internal/elasticsearch/index/index_test.go
k-yomo: Sorry, fixed!
https://github.com/elastic/terraform-provider-elasticstack/pull/158/commits/b67b1a9808a7ac0409a1adcc04683f183cf2d2e8

tobio:
jenkins test this

internal/elasticsearch/index/index.go
tobio: > Not only for this but I think we should have `deletion_protection`

Can you open an issue for this. 

>  let me create a separate PR because it should be also removed from the other fields too

+1 on this, but can we remove it from the new code in this PR first?

k-yomo: Sure, then let me just put the all changes to this PR since I just realized they are not many.
https://github.com/elastic/terraform-provider-elasticstack/pull/158/commits/4d26ec26771df592f56f4cb8799db918ac915e05

Also opened the issue!
https://github.com/elastic/terraform-provider-elasticstack/issues/165

tobio:
jenkins test this

tobio (approved):

tobio:
jenkins test this

yoavshemesh-riskified:
Hi @tobio , 
when is this feature will be available within the TF provider?
we really need it.

tobio:
@yoavshemesh-riskified I don't have a firm date available sorry. Definitely before the end of they year. More realistically, we're working on https://github.com/elastic/terraform-provider-elasticstack/issues/70 and will release the provider after that's merged so hopefully sometime in November.

PR #167, opened 2022-10-26
Add `deletion_protection` field to index

### LINKED ISSUE #165: [Feature] Make index resource safer to manage
We may want to have `deletion_protection` in `index` resource like cloud SQL or the other important resources have to prevent unintended deletion.
The `deletion_protection` should be `true` by default and need to be updated it to `false` when we truly want to delete the index. With this, we will have a safer way to manage indices for production use cases.
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance

k-yomo:
Resolves #165


tobio:
Discussing this one internally as well. 

[`lifecycle.prevent_destroy`](https://developer.hashicorp.com/terraform/tutorials/state/resource-lifecycle?in=terraform%2Fstate#prevent-resource-deletion) performs a similar role and is provided by Terraform on all resources. It's opt-in which may be less effective at preventing accidental destruction, but is more flexible for users who don't want to flip an explicit flag when destroying resources. 

As it stands, this is a breaking change to the provider, adding functionality which is already available to users who require it (i.e `lifecycle.prevent_destroy`).

k-yomo:
@tobio 
Thank you for the discussion!
IMO, safety is more important for this kind of resource than flexibility as you suggested the change on the analyzer related fields not to have `ForceNew`.  
Also, adding `deletion_protection` does not require a change to the existing code unless they try to intentionally destroy/recreate it so not a big breaking change. (It was introduced to cloud SQL after a while since the resource was created https://github.com/hashicorp/terraform-provider-google-beta/pull/2579)

Zawadidone:
@tobio `lifecycle.prevent_destroy` can not use variables which forces a wrapper around the index Terraform resource to (not) delete the resource.

tobio (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [Unreleased]
 ### Added
 - Add 'mapping_coerce' field to index resource ([#229](https://github.com/elastic/terraform-provider-elasticstack/pull/229))
+- Add 'deletion_protection' field to index resource ([#167](https://github.com/elastic/terraform-provider-elasticstack/pull/167))
```
tobio: Can we highlight this as a breaking change, with an example of how the field should look if users want to delete an index.

k-yomo: Sure, updated!
https://github.com/elastic/terraform-provider-elasticstack/pull/167/files?short_path=06572a9#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

tobio (approved):

PR #169, opened 2022-10-27
Add Github actions workflow

### LINKED ISSUE #168: [Feature] Add basic PR checks as Github actions
Lint is run as part of the current acceptance test task, however this isn't run automatically for community sourced PR's which unnecessarily lengthens the feedback cycle for those PR's. Basic CI checks should be automatically triggered for all PRs via GH actions, they should include:

* Lint + Compilation
* Unit tests
* That there are no uncommitted docs changes

k-yomo:
Resolves #168

Add Github action workflow which includes
- build
- lint (misspell golangci-lint fmt-check docs-check)
- acctest
  - matrix is used to replace jenkins, but if we keep jenkins, then just specifying the latest version might be enough.

example run is below
https://github.com/k-yomo/terraform-provider-elasticstack/actions/runs/3342618068

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
         run: go mod download
 
       - name: Build
-        run: go build -v .
+        run: make build
 
   lint:
     name: Lint
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Tests
+name: Build/Lint/Test
 on:
   push:
     paths-ignore:
@@ -37,15 +37,12 @@ jobs:
           cache: true
           cache-dependency-path: tools/go.sum
       - uses: hashicorp/setup-terraform@v2
-        with:
-          terraform_version: '1.2.*'
-          terraform_wrapper: false
 
       - name: Lint
         run: make lint
 
   test:
-    name: Matrix Test
+    name: Matrix Acceptance Test
     needs: build
     runs-on: ubuntu-latest
     services:
@@ -64,8 +61,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        terraform:
-          - '1.3.3'
         elasticsearch:
           - '7.11.2'
           - '7.12.1'
@@ -84,9 +79,6 @@ jobs:
           go-version-file: 'go.mod'
           cache: true
       - uses: hashicorp/setup-terraform@v2
-        with:
-          terraform_version: ${{ matrix.terraform }}
-          terraform_wrapper: false
 
       - name: Get dependencies
         run: go mod download
@@ -96,7 +88,6 @@ jobs:
         run: make testacc
         env:
           TF_ACC: "1"
-          TF_ACC_TERRAFORM_VERSION: ${{ matrix.terraform }}
           ELASTICSEARCH_VERSION: ${{ matrix.elasticsearch }}
           ELASTICSEARCH_ENDPOINTS: "http://localhost:9200"
           ELASTICSEARCH_USERNAME: "elastic"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,6 +41,8 @@ jobs:
           cache: true
           cache-dependency-path: tools/go.sum
       - uses: hashicorp/setup-terraform@v2
+        with:
+          terraform_wrapper: false
 
       - name: Lint
         run: make lint
@@ -83,6 +85,8 @@ jobs:
           go-version-file: 'go.mod'
           cache: true
       - uses: hashicorp/setup-terraform@v2
+        with:
+          terraform_wrapper: false
 
       - name: Get dependencies
         run: go mod download
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,6 +51,8 @@ jobs:
     name: Matrix Acceptance Test
     needs: build
     runs-on: ubuntu-latest
+    env:
+      ELASTIC_PASSWORD: password
     services:
       elasticsearch:
         image: docker.elastic.co/elasticsearch/elasticsearch:${{ matrix.elasticsearch }}
@@ -59,7 +61,6 @@ jobs:
           xpack.security.enabled: true
           repositories.url.allowed_urls: https://example.com/*
           path.repo: /tmp
-          ELASTIC_PASSWORD: password
         ports:
           - 9200:9200
         options: --health-cmd="curl http://localhost:9200/_cluster/health" --health-interval=10s --health-timeout=5s --health-retries=10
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,103 @@
+name: Tests
+on:
+  push:
```
tobio, line 3: Can we add a `pull_request` trigger here as well to have the checks show up in PRs.

.github/workflows/test.yml
```diff
@@ -0,0 +1,103 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: 'tools/go.mod'
+          cache: true
+          cache-dependency-path: tools/go.sum
+      - uses: hashicorp/setup-terraform@v2
+        with:
+          terraform_version: '1.2.*'
```
tobio: Any reason to not just run latest? 

```suggestion
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,103 @@
...
+          ELASTIC_PASSWORD: password
+        ports:
+          - 9200:9200
+        options: --health-cmd="curl http://localhost:9200/_cluster/health" --health-interval=10s --health-timeout=5s --health-retries=10
+    timeout-minutes: 15
+    strategy:
+      fail-fast: false
+      matrix:
+        terraform:
+          - '1.3.3'
```
tobio: I think for now we should just run latest. I know I've got an open issue to look at having a pinned version _as well as_ latest, but I think that's only worthwhile if there's ongoing stability issues with latest (which shouldn't really be the case). 

```suggestion
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,103 @@
...
+
+      - name: Get dependencies
+        run: go mod download
+
+      - name: TF acceptance tests
+        timeout-minutes: 10
+        run: make testacc
+        env:
+          TF_ACC: "1"
+          TF_ACC_TERRAFORM_VERSION: ${{ matrix.terraform }}
```
tobio:
```suggestion
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,103 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+          cache: true
+      - uses: hashicorp/setup-terraform@v2
+        with:
+          terraform_version: ${{ matrix.terraform }}
+          terraform_wrapper: false
```
tobio: This step shouldn't be required since the acceptance test runner would download the relevant TF version

```suggestion
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,103 @@
+name: Tests
```
tobio:
```suggestion
name: Build/Lint/Test
```

Naming is hard, `Tests` doesn't really describe what's happening in this workflow though.

.github/workflows/test.yml
```diff
@@ -0,0 +1,103 @@
...
+      - uses: hashicorp/setup-terraform@v2
+        with:
+          terraform_version: '1.2.*'
+          terraform_wrapper: false
+
+      - name: Lint
+        run: make lint
+
+  test:
+    name: Matrix Test
```
tobio:
```suggestion
    name: Matrix Acceptance Test
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,103 @@
...
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+          cache: true
+
+      - name: Get dependencies
+        run: go mod download
+
+      - name: Build
+        run: go build -v .
```
tobio:
```suggestion
        run: make build
```

tobio (commented):
❤️ can we get these action results reported into the PR as well. 

We should also make sure `make test` is run as part of the `Build` job.

.github/workflows/test.yml
```diff
@@ -0,0 +1,103 @@
...
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+          cache: true
+
+      - name: Get dependencies
+        run: go mod download
```
tobio, line 28: Is this step actually required? Doesn't `setup-go` download the module dependencies already?

k-yomo:
> We should also make sure make test is run as part of the Build job.

Should we?  I think acceptance tests cover the unit test, so it wouldn't be needed.

.github/workflows/test.yml
k-yomo: `setup-go` caches the downloaded module but it doesn't download the module I think!
The official doc also installs modules separately.
https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go#building-and-testing-your-code

k-yomo:
> can we get these action results reported into the PR as well.

I guess once it's merged, it will be shown up on the list.
Once we confirm it works, we also would be better to require the CI to pass for merging a PR!

.github/workflows/test.yml
k-yomo: Oh no, missed to update.  just removed as suggested!

k-yomo:
Oh sorry, CI is failing due to terraform command related issue, let me check
https://github.com/k-yomo/terraform-provider-elasticstack/actions/runs/3343086487/jobs/5535969837

k-yomo:
This is the latest run.  It seems working well!
https://github.com/k-yomo/terraform-provider-elasticstack/actions/runs/3343165215

.github/workflows/test.yml
```diff
@@ -0,0 +1,102 @@
...
+    runs-on: ubuntu-latest
+    services:
+      elasticsearch:
+        image: docker.elastic.co/elasticsearch/elasticsearch:${{ matrix.elasticsearch }}
+        env:
+          discovery.type: single-node
+          xpack.security.enabled: true
+          repositories.url.allowed_urls: https://example.com/*
+          path.repo: /tmp
+          ELASTIC_PASSWORD: password
```
tobio: Might be worth moving this value to `test.env.ELASTIC_PASSWORD` so it can be shared with the `TF acceptance tests` step.

tobio:
> Once we confirm it works, we also would be better to require the CI to pass for merging a PR!

💯, I'll remove the Jenkins jobs too.

tobio (approved):
Thanks so much for this. 

Having lint more visible is awesome and triggering these automatically should make it much nicer for community PR's too!

Happy to move the password to an env variable if you don't get to it. I'll get this merged and update the PR rules.

.github/workflows/test.yml
k-yomo: Sounds good! Is that like this?
https://github.com/elastic/terraform-provider-elasticstack/pull/169/commits/42b38b36fe17a042e9f2ac505fad0ae6aad15ce0

k-yomo: We also needed this
https://github.com/elastic/terraform-provider-elasticstack/pull/169/commits/066be9c139c151964441158305ea8c1e8191ac69

tobio:
jenkins test this

tobio (approved):

PR #172, opened 2022-10-28
Add release workflow

### LINKED ISSUE #157: [Feature] Provider releases should be automated through CI rather than run on a local dev machine.
For the most part, we should be able to copy the pipeline config from the Elastic Cloud provider.

k-yomo:
Resolves #157 

This PR adds release workflow to release new version of the provider to Terraform registry when we cut the new version tag.  The workflow itself is pretty much same as the one provided by scaffolidng.
https://github.com/hashicorp/terraform-provider-scaffolding/blob/main/.github/workflows/release.yml

We need to set below two variables to secret beforehand.
- `GPG_PRIVATE_KEY`
- `PASSPHRASE`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,38 @@
+name: Release
+on:
+  push:
+    tags:
+      - 'v*'
+
+permissions:
+  contents: write
+
+jobs:
+  goreleaser:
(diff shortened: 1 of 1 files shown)


tobio:
I'm going to re-use the jenkins based workflow on this one since it allows us to use the existing secret management for the GPG key.

PR #173, opened 2022-10-29
Add script resource

### LINKED ISSUE #107: [Feature] Support for search templates / stored scripts
We have many search template we deploy per-stack using the API/Dev-tools. This is actually errror-prone and we would like to include it in our Terraform scripts so we can manage it efficiently.

**Available APIs**
Create&Update script: https://www.elastic.co/guide/en/elasticsearch/reference/current/create-stored-script-api.html
Check currently deployed scripts: https://www.elastic.co/guide/en/elasticsearch/reference/current/get-stored-script-api.html

**Solution**
Provide support to create/update/delete scripts & search template

Using API / Dev console with query stored in standalone text file.

k-yomo:
Resolves #107 

This PR adds script resource in reference to the below docs.
https://www.elastic.co/guide/en/elasticsearch/reference/current/create-stored-script-api.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/get-stored-script-api.html

📝 `context` and `params` can't be retrieved by GET endpoint, so they are not importable as of now.


internal/clients/cluster.go
```diff
@@ -180,3 +180,59 @@ func (a *ApiClient) GetElasticsearchSettings(ctx context.Context) (map[string]in
...
+	if res.StatusCode == http.StatusNotFound {
+		return nil, nil
+	}
+	if diags := utils.CheckError(res, fmt.Sprintf("Unable to get the script: %s", id)); diags.HasError() {
+		return nil, diags
+	}
+	type getScriptResponse = struct {
+		Script *models.Script `json:"script"`
+	}
+	var scriptResponse getScriptResponse
```
tobio:
```suggestion
	var scriptResponse struct {
		Script *models.Script `json:"script"`
	}
```

Any benefit to the named type here?

tobio:
```suggestion
	var scriptResponse struct {
		Script *models.Script `json:"script"`
	}
```

Any benefit to the named type here?

docs/resources/elasticsearch_script.md
```diff
@@ -0,0 +1,65 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "elasticstack_elasticsearch_script Resource - terraform-provider-elasticstack"
+subcategory: ""
```
tobio: `subcategory: "Cluster"` (I think, not sure if this warrants it's own category)

Might need to add a [template](https://github.com/elastic/terraform-provider-elasticstack/tree/master/templates/resources)

internal/clients/cluster.go
```diff
@@ -180,3 +180,59 @@ func (a *ApiClient) GetElasticsearchSettings(ctx context.Context) (map[string]in
...
+	}
+
+	return scriptResponse.Script, nil
+}
+
+func (a *ApiClient) PutElasticsearchScript(ctx context.Context, script *models.Script) diag.Diagnostics {
+	req := struct {
+		Script *models.Script `json:"script"`
+	}{
+		Script: script,
```
tobio:
```suggestion
		script,
```

internal/clients/cluster.go
```diff
@@ -180,3 +180,59 @@ func (a *ApiClient) GetElasticsearchSettings(ctx context.Context) (map[string]in
...
+func (a *ApiClient) GetElasticsearchScript(ctx context.Context, id string) (*models.Script, diag.Diagnostics) {
+	res, err := a.es.GetScript(id, a.es.GetScript.WithContext(ctx))
+	if err != nil {
+		return nil, diag.FromErr(err)
+	}
+	defer res.Body.Close()
+	if res.StatusCode == http.StatusNotFound {
+		return nil, nil
+	}
+	if diags := utils.CheckError(res, fmt.Sprintf("Unable to get the script: %s", id)); diags.HasError() {
```
tobio:
```suggestion
	if diags := utils.CheckError(res, fmt.Sprintf("Unable to get saved script: %s", id)); diags.HasError() {
```

internal/clients/cluster.go
```diff
@@ -180,3 +180,59 @@ func (a *ApiClient) GetElasticsearchSettings(ctx context.Context) (map[string]in
...
+	scriptBytes, err := json.Marshal(req)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	res, err := a.es.PutScript(script.ID, bytes.NewReader(scriptBytes), a.es.PutScript.WithContext(ctx), a.es.PutScript.WithScriptContext(script.Context))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	defer res.Body.Close()
+	if diags := utils.CheckError(res, "Unable to put the script"); diags.HasError() {
```
tobio:
```suggestion
	if diags := utils.CheckError(res, "Unable to put saved script"); diags.HasError() {
```

internal/elasticsearch/cluster/script.go
```diff
@@ -0,0 +1,151 @@
...
+			Description: "Identifier for the stored script. Must be unique within the cluster.",
+			Type:        schema.TypeString,
+			Required:    true,
+			ForceNew:    true,
+		},
+		"lang": {
+			Description:  "Script language. For search templates, use `mustache`. Defaults to `painless`.",
+			Type:         schema.TypeString,
+			Optional:     true,
+			Default:      "painless",
```
tobio: This default makes sense in the context of a stored script, but not when used to manage a search template. IMO it would be better to mirror the API and have this as a required field without a default.

internal/elasticsearch/cluster/script.go
```diff
@@ -0,0 +1,151 @@
...
+
+	id := d.Id()
+	compId, diags := clients.CompositeIdFromStr(id)
+	if diags.HasError() {
+		return diags
+	}
+
+	script, diags := client.GetElasticsearchScript(ctx, compId.ResourceId)
+	if !d.IsNewResource() && script == nil && diags == nil {
+		tflog.Warn(ctx, fmt.Sprintf(`Script "%s" not found, removing from state`, compId.ResourceId))
```
tobio, line 82: It doesn't look like we log a message in this case for the other resources. Any reason this should be a warning?

internal/elasticsearch/cluster/script_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"testing"
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/acctest"
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
+	sdkacctest "github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestAccResourceScript(t *testing.T) {
```
tobio, line 14: Can we add a test managing a search template as well?

docs/resources/elasticsearch_script.md
k-yomo: Oh totally missed it, will fix.
I want this😂
https://github.com/hashicorp/terraform-plugin-docs/issues/156

tobio: Nice, yep very much want that one too :)

internal/elasticsearch/cluster/script.go
k-yomo: Yes, the reason is to let the user know that the resource is removed from state when actually not found on API which is sometimes unexpected for users.
I think it's pretty common pattern which the other providers do too.
https://github.com/hashicorp/terraform-provider-aws/search?q=removing+from+state

It it makes sense, I'll add it to the other resources too with `!d.IsNewResource()` check.
https://github.com/hashicorp/terraform-provider-aws/issues/16796

internal/elasticsearch/cluster/script.go
k-yomo: Oh that's true, fixed to make it required field.

internal/clients/cluster.go
k-yomo: Oh no, unnamed type would be fine, fixed so!

docs/resources/elasticsearch_script.md
```diff
@@ -0,0 +1,65 @@
...
+
+```terraform
+provider "elasticstack" {
+  elasticsearch {}
+}
+
+resource "elasticstack_elasticsearch_script" "my_script" {
+  script_id = "my_script"
+  source    = "Math.log(_score * 2) + params['my_modifier']"
+  context   = "score"
```
tobio, line 24: This example is no longer valid, and needs `lang` defined.

docs/resources/elasticsearch_script.md
```diff
@@ -0,0 +1,65 @@
...
+page_title: "Elasticstack: elasticstack_elasticsearch_script Resource"
+description: |-
+  Creates or updates a stored script or search template.
+---
+
+# Resource: elasticstack_elasticsearch_script
+
+Creates or updates a stored script or search template. See https://www.elastic.co/guide/en/elasticsearch/reference/current/create-stored-script-api.html
+
+## Example Usage
```
tobio, line 13: Should have said this earlier sorry, having a search template example would be nice too.

tobio (commented):
Couple of minor docs changes.

docs/resources/elasticsearch_script.md
k-yomo: Oh good catch, fixed🙏
https://github.com/elastic/terraform-provider-elasticstack/pull/173/commits/5bde0d445569e6c3d5c550dd3eaabed0a3c6403e

internal/elasticsearch/cluster/script.go
k-yomo: I just realized that adding `!d.IsNewResource()` is not good here since it can cause null pointer error in the following code.
It might be better to return error when the resource is `d.IsNewResource())` as aws provider does, but I simply removed to align with the other resources for now.

tobio (approved):

PR #185, opened 2022-11-01
Add warn log when to remove resource from state

k-yomo:


tobio (commented):
LGTM, can we add a changelog entry for this one though. It might be nice to call out the behavioural difference for users wondering why they're suddenly getting a warning.

tobio (approved):

PR #188, opened 2022-11-08
Add system-user resource

### LINKED ISSUE #104: Add new resource to manage system user passwords

k-yomo:
Resolves #104 

This PR adds `elasticstack_elasticsearch_security_system_user` resource to manage built-in system user.


internal/elasticsearch/security/system_user.go
```diff
@@ -0,0 +1,165 @@
...
+		Description: "Updates system user's password and enablement. See, https://www.elastic.co/guide/en/elasticsearch/reference/current/built-in-users.html",
+
+		CreateContext: resourceSecuritySystemUserPut,
+		UpdateContext: resourceSecuritySystemUserPut,
+		ReadContext:   resourceSecuritySystemUserRead,
+		DeleteContext: resourceSecuritySystemUserDelete,
+
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
```
tobio: I'm not sure it ever makes sense to import this resource? A user would never be able to import the password/hash and so it's pretty likely there would be changes to be applied. 

Given the 'create' doesn't actually create the user I'd be inclined to just make this not importable. WDYT? 
```suggestion
```

examples/resources/elasticstack_elasticsearch_security_system_user/resource.tf
```diff
@@ -0,0 +1,16 @@
+provider "elasticstack" {
+  elasticsearch {}
+}
+
+resource "elasticstack_elasticsearch_security_system_user" "kibana_system" {
+  username = "kibana_system"
+
+  // use hashed password: see https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-user.html#security-api-put-user-request-body
```
tobio: I initially read this comment as if I couldn't set the plaintext password. 
```suggestion
  // For details on how to generate the hashed password see https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-user.html#security-api-put-user-request-body
```

tobio (commented):
Looks nice, couple of minor comments. We'd need a changelog entry here as well please.

internal/elasticsearch/security/system_user.go
k-yomo: Oh that's true and making the resource not importable makes sense.
Fixed it!
https://github.com/elastic/terraform-provider-elasticstack/pull/188/commits/a641bcd07ecf8a5c55f7eeef95bb52e023ce4379

docs/resources/elasticsearch_security_system_user.md
```diff
@@ -0,0 +1,64 @@
...
+layout: ""
+page_title: "Elasticstack: elasticstack_elasticsearch_security_system_user Resource"
+description: |-
+  Updates system user's password and enablement.
+---
+
+# Resource: elasticstack_elasticsearch_security_system_user
+
+Updates system user's password and enablement. See, https://www.elastic.co/guide/en/elasticsearch/reference/current/built-in-users.html
+Since this resource is to manage undeletable built-in user, destroy will removes it only from terraform state and the user itself won't be deleted.
```
tobio:
```suggestion
Since this resource is to manage built-in users, destroy will not delete the underlying Elasticsearch and will only remove it from Terraform state.
```

templates/resources/elasticsearch_security_system_user.md.tmpl
```diff
@@ -0,0 +1,18 @@
...
+layout: ""
+page_title: "Elasticstack: elasticstack_elasticsearch_security_system_user Resource"
+description: |-
+  Updates system user's password and enablement.
+---
+
+# Resource: elasticstack_elasticsearch_security_system_user
+
+Updates system user's password and enablement. See, https://www.elastic.co/guide/en/elasticsearch/reference/current/built-in-users.html
+Since this resource is to manage undeletable built-in user, destroy will removes it only from terraform state and the user itself won't be deleted.
```
tobio:
```suggestion
Since this resource is to manage built-in users, destroy will not delete the underlying Elasticsearch and will only remove it from Terraform state.
```

tobio (commented):
Small update to the docs here. Looking great!

tobio (approved):

Fuco1:
Can we get a release? Please :heart_eyes:

tobio:
@Fuco1 we'll have a release out in the next 2 weeks

PR #191, opened 2022-11-11
Add authentication to Elasticsearch via client cert

andreaskapfer:
This PR adds the ability to authenticate to Elasticsearch via client certificate.

It also fixes some bugs related to the options `ca_file` and `ca_data`.


internal/clients/api_client.go
```diff
@@ -116,6 +115,37 @@ func NewApiClientFunc(version string, p *schema.Provider) func(context.Context,
...
+						if err != nil {
+							diags = append(diags, diag.Diagnostic{
+								Severity: diag.Error,
+								Summary:  "Unable to read certificate or key file",
+								Detail:   err.Error(),
+							})
+							return nil, diags
+						}
+						ensureTLSClientConfig(&config)
+						config.Transport.(*http.Transport).TLSClientConfig.Certificates = []tls.Certificate{cert}
```
tobio: IMO these blocks would likely read more nicely if `ensureTLSClientConfig` returned a `*http.Transport` so the subsequent code could avoid the casts. WDYT?

To be clear: "You're wrong because..." is totally fine.

internal/clients/api_client.go
```diff
@@ -116,6 +115,37 @@ func NewApiClientFunc(version string, p *schema.Provider) func(context.Context,
 				if caData, ok := esConfig["ca_data"]; ok && caData.(string) != "" {
 					config.CACert = []byte(caData.(string))
 				}
+
+				if certFile, ok := esConfig["cert_file"]; ok && certFile.(string) != "" {
+					if keyFile, ok := esConfig["key_file"]; ok && keyFile.(string) != "" {
```
tobio, line 120: I know this is enforced in the schema, but IMO we should return an error if `cert_file` is ok but `key_file` is not ok.

internal/clients/api_client.go
```diff
@@ -116,6 +115,37 @@ func NewApiClientFunc(version string, p *schema.Provider) func(context.Context,
...
+								Detail:   err.Error(),
+							})
+							return nil, diags
+						}
+						ensureTLSClientConfig(&config)
+						config.Transport.(*http.Transport).TLSClientConfig.Certificates = []tls.Certificate{cert}
+					}
+				}
+				if certData, ok := esConfig["cert_data"]; ok && certData.(string) != "" {
+					if keyData, ok := esConfig["key_data"]; ok && keyData.(string) != "" {
```
tobio, line 142: Similarly, I think this code should error out if `cert_data` is specified without `key_data`

internal/clients/api_client.go
```diff
@@ -199,6 +229,30 @@ func NewApiClient(d *schema.ResourceData, meta interface{}) (*ApiClient, error)
 			}
 			config.CACert = caCert
 		}
+		if caData, ok := conn["ca_data"]; ok && caData.(string) != "" {
+			config.CACert = []byte(caData.(string))
+		}
+
+		if certFile, ok := conn["cert_file"]; ok && certFile.(string) != "" {
+			if keyFile, ok := conn["key_file"]; ok && keyFile.(string) != "" {
```
tobio, line 250: Similarly

internal/clients/api_client.go
```diff
@@ -199,6 +229,30 @@ func NewApiClient(d *schema.ResourceData, meta interface{}) (*ApiClient, error)
...
+				cert, err := tls.LoadX509KeyPair(certFile.(string), keyFile.(string))
+				if err != nil {
+					return nil, fmt.Errorf("Unable to read certificate or key file: %w", err)
+				}
+				ensureTLSClientConfig(&config)
+				config.Transport.(*http.Transport).TLSClientConfig.Certificates = []tls.Certificate{cert}
+			}
+		}
+		if certData, ok := conn["cert_data"]; ok && certData.(string) != "" {
+			if keyData, ok := conn["key_data"]; ok && keyData.(string) != "" {
```
tobio, line 262: Similarly

tobio (commented):
Sorry for the delay here, took a little longer to spin up a cluster with PKI auth to test this. 

Couple of nits on the code.

internal/clients/api_client.go
andreaskapfer: Good idea! I would even go one step further. What about returning `*tls.Config`?
I guess everyone calling `ensureTLSClientConfig` wants to customize `TLSClientConfig` afterwards.

tobio: Sounds great, thanks!

andreaskapfer: done

internal/clients/api_client.go
andreaskapfer: done

internal/clients/api_client.go
andreaskapfer: done

internal/clients/api_client.go
andreaskapfer: done

internal/clients/api_client.go
andreaskapfer: done

tobio (approved):

tobio:
Thanks for this @andreaskapfer we'll have a new version of the provider out with this change in the next few weeks.

andreaskapfer:
Great! Many thanks for the review @tobio

tobio:
This has been released with [0.5.0](https://github.com/elastic/terraform-provider-elasticstack/releases/tag/v0.5.0)

PR #194, opened 2022-11-16
Create an ApiClient directly from the env variables for acceptance testing

tobio:
This makes the client accessible to all test callbacks, and doesn't rely on the Terraform provider being fully configured, i.e the first step having been run


webfella (approved):
This looks great mate! Appreciate you jumping on this quick too

Issue #100, opened 2022-04-25
[Feature] Support Kibana connection in provider & in kibana resources


Kushmaro:
In order for us to be able to manage Kibana resources, the provider connection block should also support kibana.
the kibana connection block should also accept an `insecure = true` parameter, that accept untrusted certificates. (this parameter value defaults to `false`)

### Example1 :
if an Elasticsearch block is configured, the kibana block should automatically pick up credentials as they should be the same.
```
provider "elasticstack" {
  elasticsearch {
    username  = "elastic"
    password  = "changeme"
    endpoints = ["http://localhost:9200"]
  }
  kibana {
    endpoints = ["https://my.kibana.url.com:<port>"]
    insecure = true
  }
}
```

### Example2 :
If only kibana is defined, it should require credentials to be mandatory
```
provider "elasticstack" {
  kibana {
    username  = "elastic"
    password  = "changeme"
    endpoints = ["https://my.kibana.url.com:<port>"]
  }
}
```

Lastly, any kibana resource created in the provider should support the same sort of connection specification like the elasticsearch resources do, so that customers can create deployments using  the Elastic Cloud terraform provider, and configure them on the fly.

tobio:
Is there a plan around interfacing with the Kibana API? AFAIK Elastic doesn't provide a Go (or any other language) client for the Kibana API. If there's an expectation that we build that client as part of adding TF support for those resources then it's going to add significantly to the dev time on these tasks. Is there capacity within the Clients team to at least provide the scaffold for these clients? 

@Kushmaro (@d-carmo for transparency)

Kushmaro:
@tobio Yes, there's a plan to interface with the Kibana API. The provider should interact with all parts of the stack.
The ResponseOps team have rules and alerts for TF on their roadmap. 

/cc @shanisagiv1

tobio:
Sure, but who's building the actual api client? If it's us then 10x any estimate for this task since we'd also have to build all the basics of an API client.

PR #198, opened 2022-11-21
Refactor elasticsearch cluster client methods

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/100

As it stands the ApiClient struct is becoming cluttered and massive. Adding Kibana will only make this worse. First step in breaking up the API access methods.


webfella (approved):
Nice one man!

PR #200, opened 2022-11-21
Add sensitive to api_key and encoded fields in api key resource

### LINKED ISSUE #199: [Feature] `api_key` and `encoded` should be sensitive fields in api_key resource
`api_key` and `encoded` (encoded api key) are the sensitive values, so the fields should be marked as `sensitive` same as the password field in user resource.
https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/security/user.go#L36-L51

The other provider marks api key as sensitive too.
https://github.com/hashicorp/terraform-provider-google/blob/main/google/resource_google_service_account_key.go#L75-L80

k-yomo:
Resolves #199


tobio (approved):
Thanks for this!

Issue #203, opened 2022-11-21
DRY up the elasticsearch connection schema and client construction. 


tobio:
Currently the provider defines an [`elasticsearch`](https://github.com/elastic/terraform-provider-elasticstack/blob/main/provider/provider.go#L24-L80) block which is duplicated by the [`elasticsearch_connection`](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/utils/utils.go#L126-L182) block added to each resource. 

These schemas are expected to be identical, we shouldn't be defining them twice. 

Similarly, but slight more involved is the code which reads these blocks. [`NewApiClientFunc`](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/clients/api_client.go#L61) reads the provider level configuration whilst [`NewApiClient`](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/clients/api_client.go#L164) reads the resource level configuration. At a high level, their behaviour is the same, we should refactor the common code. `NewApiClient` currently returns an `error`, but the consuming code promptly turns this in a `Diagnostics` so we should be able to just return the `Diagnostics` directly.  

Related to https://github.com/elastic/terraform-provider-elasticstack/pull/191

webfella:
Added to `0.5.0`, so we can get this done before adding more duplication via https://github.com/elastic/terraform-provider-elasticstack/issues/100.

Kushmaro:
@tobio @webfella , we did this BY DESIGN, in order to be able to use both the Cloud provider alongside the stack provider. 
please do NOT refactor this.

tobio:
Currently just sharing the schema between the two connection methods. 

That said, we should remove the `elasticsearch_connection` block from resources. It doesn't work during import, so creates a bad initial user experience. The same functionality is already provided by the core Terraform configuration language, so we're just needlessly adding complexity to the provider.

PR #209, opened 2022-11-23
DRY up ES connection

webfella:
First part of [#203](https://github.com/elastic/terraform-provider-elasticstack/issues/203).

### DIFF
--- a/internal/schema/connection.go
+++ b/internal/schema/connection.go
@@ -16,10 +16,17 @@ func GetConnectionSchema(keyName string, useEnvAsDefault bool) *schema.Schema {
 	keyFilePath := makePathRef(keyName, "key_file")
 	keyDataPath := makePathRef(keyName, "key_data")
 
+	usernameRequiredWithValidation := []string{passwordPath}
+	passwordRequiredWithValidation := []string{usernamePath}
+
 	withEnvDefault := func(key string, dv interface{}) schema.SchemaDefaultFunc { return nil }
 
 	if useEnvAsDefault {
 		withEnvDefault = func(key string, dv interface{}) schema.SchemaDefaultFunc { return schema.EnvDefaultFunc(key, dv) }
+
+		// RequireWith validation isn't compatible when used in conjunction with DefaultFunc
+		usernameRequiredWithValidation = nil
+		passwordRequiredWithValidation = nil
 	}
 
 	return &schema.Schema{
@@ -30,17 +37,19 @@ func GetConnectionSchema(keyName string, useEnvAsDefault bool) *schema.Schema {
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
 				"username": {
-					Description: "Username to use for API authentication to Elasticsearch.",
-					Type:        schema.TypeString,
-					Optional:    true,
-					DefaultFunc: withEnvDefault("ELASTICSEARCH_USERNAME", nil),
+					Description:  "Username to use for API authentication to Elasticsearch.",
+					Type:         schema.TypeString,
+					Optional:     true,
+					DefaultFunc:  withEnvDefault("ELASTICSEARCH_USERNAME", nil),
+					RequiredWith: usernameRequiredWithValidation,
 				},
 				"password": {
-					Description: "Password to use for API authentication to Elasticsearch.",
-					Type:        schema.TypeString,
-					Optional:    true,
-					Sensitive:   true,
-					DefaultFunc: withEnvDefault("ELASTICSEARCH_PASSWORD", nil),
+					Description:  "Password to use for API authentication to Elasticsearch.",
+					Type:         schema.TypeString,
+					Optional:     true,
+					Sensitive:    true,
+					DefaultFunc:  withEnvDefault("ELASTICSEARCH_PASSWORD", nil),
+					RequiredWith: passwordRequiredWithValidation,
 				},
 				"api_key": {
 					Description:   "API Key to use for authentication to Elasticsearch",
--- a/internal/schema/connection.go
+++ b/internal/schema/connection.go
@@ -28,7 +28,6 @@ func GetConnectionSchema(keyName string) *schema.Schema {
 					Type:        schema.TypeString,
 					Optional:    true,
 					DefaultFunc: schema.EnvDefaultFunc("ELASTICSEARCH_USERNAME", nil),
-					// RequiredWith: []string{passwordPath},
 				},
 				"password": {
 					Description: "Password to use for API authentication to Elasticsearch.",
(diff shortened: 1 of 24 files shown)


internal/schema/connection.go
```diff
@@ -0,0 +1,112 @@
...
+		MaxItems:    1,
+		Optional:    true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"username": {
+					Description: "Username to use for API authentication to Elasticsearch.",
+					Type:        schema.TypeString,
+					Optional:    true,
+					DefaultFunc: schema.EnvDefaultFunc("ELASTICSEARCH_USERNAME", nil),
+					// RequiredWith: []string{passwordPath},
```
webfella: Everything seems to carry over fine, apart from the `RequiredWith` validation for `username` and `password` fields.

Reading the docs I assumed this would only activate if either of those fields were present, but maybe the `DefaultFunc` is messing with it (`DefaultFunc` was only present on the elasticsearch schema, and not the `elasticsearch_connection` blocks)?

webfella (commented):
I'll be back on later with fresh eyes, but I've captured what I'm stuck on at the moment @tobio if you have any insight.

internal/schema/connection.go
tobio: https://github.com/hashicorp/terraform-plugin-sdk/issues/1099

Taking a look at the TF code there's not much we can do to have both of these attributes at the moment. Seems like this is the best approach for now, are you able to open an issue so if the TF issue is closed off we can improve the validation within the provider?

internal/schema/connection.go
```diff
@@ -0,0 +1,112 @@
...
+		Type:        schema.TypeList,
+		MaxItems:    1,
+		Optional:    true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"username": {
+					Description: "Username to use for API authentication to Elasticsearch.",
+					Type:        schema.TypeString,
+					Optional:    true,
+					DefaultFunc: schema.EnvDefaultFunc("ELASTICSEARCH_USERNAME", nil),
```
tobio: We should only be taking the defaults from the env in the provider schema too. Something like:

```suggestion
func GetConnectionSchema(keyName string, useEnvAsDefault bool) *schema.Schema {
	usernamePath := makePathRef(keyName, "username")
	passwordPath := makePathRef(keyName, "password")
	caFilePath := makePathRef(keyName, "ca_file")
	caDataPath := makePathRef(keyName, "ca_data")
	certFilePath := makePathRef(keyName, "cert_file")
	certDataPath := makePathRef(keyName, "cert_data")
	keyFilePath := makePathRef(keyName, "key_file")
	keyDataPath := makePathRef(keyName, "key_data")
	
	withEnvDefault := func(key string, dv interface{}) schema.SchemaDefaultFunc { return nil }
	if useEnvAsDefault {
		withEnvDefault = func(key string, dv interface{}) schema.SchemaDefaultFunc { return schema.EnvDefaultFunc(key, dv) }
	}

	return &schema.Schema{
		Description: "Elasticsearch connection configuration block.",
		Type:        schema.TypeList,
		MaxItems:    1,
		Optional:    true,
		Elem: &schema.Resource{
			Schema: map[string]*schema.Schema{
				"username": {
					Description:  "Username to use for API authentication to Elasticsearch.",
					Type:         schema.TypeString,
					Optional:     true,
					DefaultFunc:  withEnvDefault("ELASTICSEARCH_USERNAME", nil),
```

internal/schema/connection.go
tobio:
```suggestion
```

internal/schema/connection.go
```diff
@@ -0,0 +1,112 @@
...
+					// RequiredWith: []string{passwordPath},
+				},
+				"password": {
+					Description: "Password to use for API authentication to Elasticsearch.",
+					Type:        schema.TypeString,
+					Optional:    true,
+					Sensitive:   true,
+					DefaultFunc: schema.EnvDefaultFunc("ELASTICSEARCH_PASSWORD", nil),
+					// RequiredWith: []string{usernamePath},
```
tobio:
```suggestion
```

internal/utils/utils.go
```diff
@@ -123,91 +124,9 @@ func IsEmpty(v interface{}) bool {
...
-					Optional:      true,
-					RequiredWith:  []string{"elasticsearch_connection.0.cert_data"},
-					ConflictsWith: []string{"elasticsearch_connection.0.cert_file", "elasticsearch_connection.0.key_file"},
-				},
-			},
-		},
-	}
+	connectionKeyName := "elasticsearch_connection"
+
+	providedSchema[connectionKeyName] = providerSchema.GetConnectionSchema(connectionKeyName)
```
tobio: Would probably move this to a constant.

```suggestion
const resourceEsConnectionSchemaKey = "elasticsearch_connection"

// Returns the common connection schema for all the Elasticsearch resources,
// which defines the fields which can be used to configure the API access
func AddConnectionSchema(providedSchema map[string]*schema.Schema) {
	providedSchema[resourceEsConnectionSchemaKey] = providerSchema.GetConnectionSchema(resourceEsConnectionSchemaKey)
```

provider/provider.go
```diff
@@ -17,96 +18,13 @@ func init() {
 }
 
 func New(version string) func() *schema.Provider {
+	esKeyName := "elasticsearch"
```
tobio: Similarly should probably be a const

internal/schema/connection.go
webfella: Cheers mate, the best way to integrate that was going to be my next ask... so I appreciate the suggestion.

internal/schema/connection.go
webfella: Thanks for confirming mate, i'll create an issue and link it back

webfella: I hooked this into the same logic we use for the env var, so we can still keep the validation for the connection blocks https://github.com/elastic/terraform-provider-elasticstack/pull/209/commits/e0ad01d84bac53fbc1b282be2735db20752fbc4c.

If you forsee any issues there, let me know and i'll back it out and create an issue instead.

tobio (approved):
LGTM

Kushmaro:
I think we need to revert this. This functionality was intently put there so that users can connect independently to each created Elastic Deployment on Elastic Cloud, and configure it as they see fit without being limited by terraform to connection details put at the `provider` level

see such a usage example in the guide section
https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/guides/elasticstack-and-cloud

Issue #214, opened 2022-12-04
Migrate to Terraform Plugin Framework


k-yomo:
Terraform Plugin Framework is the next major version of Terraform Plugin SDKv2(this provider is currently using).
https://developer.hashicorp.com/terraform/plugin/framework

There are newly added features in Terraform Plugin Framework.
https://developer.hashicorp.com/terraform/plugin/which-sdk#do-you-need-more-features-than-sdkv2-provides

> If you maintain a large existing provider, we recommend that you begin migrating from SDKv2 to the framework by developing new resources and data sources with the framework. You can use [terraform-plugin-mux](https://developer.hashicorp.com/terraform/plugin/mux) to refactor the provider over time. Refer to the [Migrate to Terraform Plugin Framework](https://developer.hashicorp.com/terraform/plugin/sdkv2#migrate-to-terraform-plugin-framework) documentation for more details.

We may want to gradually migrate resources and data-sources to use Terraform Plugin Framework as recommended in the doc.

k-yomo:
Terraform Plugin Framework 1.0 is now GA
https://www.hashicorp.com/blog/terraform-plugin-framework-is-now-generally-available

PR #215, opened 2022-12-04
Use proto v5 mux server

k-yomo:
#214 

This PR adds [mux server](https://github.com/hashicorp/terraform-plugin-mux) in reference to aws provider.
https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/provider/factory.go

With this change, we'll be able to use  terraform-plugin-framework's provider while keeping existing provider like below.
https://github.com/hashicorp/terraform-provider-aws/blob/d82c79b1e39200aeee5f7928f015af00cfcc633f/internal/provider/factory.go#L23-L26C3

This change won't affect to the users of the provider.


provider/factory.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
+	"github.com/hashicorp/terraform-plugin-mux/tf5muxserver"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// ProtoV5ProviderServerFactory returns a muxed terraform-plugin-go protocol v5 provider factory function.
+// The primary (Plugin SDK) provider server is also returned (useful for testing).
+func ProtoV5ProviderServerFactory(ctx context.Context, version string) (func() tfprotov5.ProviderServer, *schema.Provider, error) {
+	primary := New(version)
```
tobio: `primary` doesn't seem like the right name moving forward. Once we have a second provider based on the plugin framework then this would at best be `secondary`. 

Perhaps something like `sdk2Provider` or something similar instead?

internal/acctest/acctest.go
```diff
@@ -1,21 +1,28 @@
...
 	"testing"
 
 	"github.com/elastic/terraform-provider-elasticstack/provider"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-var Providers map[string]func() (*schema.Provider, error)
+var Providers map[string]func() (tfprotov5.ProviderServer, error)
 var Provider *schema.Provider
```
tobio, line 15: This variable is almost always just used to grab an Elasticsearch client and should be removed IMO. Calling it out here since it removes the need to return the `primaryProvider` from the server factory.

provider/factory.go
```diff
@@ -0,0 +1,26 @@
...
+	servers := []func() tfprotov5.ProviderServer{
+		primary.GRPCProvider,
+	}
+
+	muxServer, err := tf5muxserver.NewMuxServer(ctx, servers...)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return muxServer.ProviderServer, primary, nil
```
tobio: Similarly, returning the `primary` provider makes sense right now, but would start to become more confusing once the migration was actually underway. 

Happy to ignore this one for now, and remove the `acctest.Provider` variable in another PR.

tobio (commented):
LGTM, minor nit on the variable naming.

provider/factory.go
k-yomo: That's true!  Let me make another PR to remove the `acctest.Provider`!

tobio (approved):

PR #218, opened 2022-12-06
Deprecate elasticsearch connection

tobio:
Configuring provider connection configuration on a per-resource basis isn't supported during import operations, and duplicates the core Terraform aliases provider functionality. 

Marking these attributes as deprecated as the first step to removing them in a future version of the provider.

### DIFF
--- a/docs/data-sources/elasticsearch_security_role.md
+++ b/docs/data-sources/elasticsearch_security_role.md
@@ -35,7 +35,7 @@ output "role" {
 
 ### Optional
 
-- `elasticsearch_connection` (Block List, Max: 1) Elasticsearch connection configuration block. (see [below for nested schema](#nestedblock--elasticsearch_connection))
+- `elasticsearch_connection` (Block List, Max: 1, Deprecated) Elasticsearch connection configuration block. This property will be removed in a future provider version. Configure the Elasticsearch connection via the provider configuration instead. (see [below for nested schema](#nestedblock--elasticsearch_connection))
 - `run_as` (Set of String) A list of users that the owners of this role can impersonate.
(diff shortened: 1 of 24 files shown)


webfella (approved):
🎉 

Big 👍👍  for adding the guide for multiple deployments too mate!

orlandothoeny:
This change makes it impossible to create a new `ec_deployment` instance + resources like an Elasticsearch User inside a single module without ugly workarounds

orlandothoeny:
Got it to work using
```
provider "elasticstack" {
  elasticsearch {
    endpoints = [ec_deployment.deployment.elasticsearch.https_endpoint]
    username  = google_secret_manager_secret_version.root_user.secret_data
    password  = google_secret_manager_secret_version.root_password.secret_data
  }
}
```
However, had to manually remove the existing user resource from the state, always got an authentication error before.

mveroone:
> Got it to work using
> 
> ```
> provider "elasticstack" {
>   elasticsearch {
>     endpoints = [ec_deployment.deployment.elasticsearch.https_endpoint]
>     username  = google_secret_manager_secret_version.root_user.secret_data
>     password  = google_secret_manager_secret_version.root_password.secret_data
>   }
> }
> ```
> 
> However, had to manually remove the existing user resource from the state, always got an authentication error before.

That won't work if you're creating an undetermined number of deployments in the code and create elasticstack resources over all of them, while `elasticsearch_connection` does.

I disagree on the initial premise of this PR that the behavior "Duplicates core terraform provider alias feature". It allows for more. Platform-engineering using both providers will be a nightmare without that per-resource block.

PR #220, opened 2022-12-07
Dry up NewApiClient

### LINKED ISSUE #203: DRY up the elasticsearch connection schema and client construction.

webfella:
`NewApiClient` now returns `diag.Diagnostics` instead of `error`, and the shared code with `NewApiClientFunc` has been abstracted out to `newEsApiClient`.

Closes #203


internal/clients/api_client.go
```diff
@@ -205,80 +91,19 @@ func NewAcceptanceTestingClient() (*ApiClient, error) {
...
 	defaultClient := meta.(*ApiClient)
-	// if the config provided let's use it
-	if esConn, ok := d.GetOk("elasticsearch_connection"); ok {
-		config := elasticsearch.Config{}
-		config.Header = http.Header{"User-Agent": []string{fmt.Sprintf("elasticstack-terraform-provider/%s", defaultClient.version)}}
 
-		// there is always only 1 connection per resource
-		conn := esConn.([]interface{})[0].(map[string]interface{})
+	if _, ok := d.GetOk(esConnectionKey); ok {
+		apiClient, diags := newEsApiClient(d, esConnectionKey, defaultClient.version)
```
tobio: Can just `return newEsApiClient(...)` here as well.

internal/clients/api_client.go
```diff
@@ -205,80 +91,19 @@ func NewAcceptanceTestingClient() (*ApiClient, error) {
...
-		es, err := elasticsearch.NewClient(config)
-		if err != nil {
-			return nil, fmt.Errorf("Unable to create Elasticsearch client")
-		}
-		if logging.IsDebugOrHigher() {
-			es.Transport = newDebugTransport("elasticsearch", es.Transport)
-		}
-		return &ApiClient{es, defaultClient.version}, nil
+		return apiClient, nil
 	} else { // or return the default client
```
tobio: It never made sense for this to be in an else block. This fn should just be

```
defaultClient := ...
if esConn exists {
  return newApiClient(...)
}

return defaultClient, nil
```

internal/clients/api_client.go
```diff
@@ -360,3 +185,123 @@ func (a *ApiClient) ClusterID(ctx context.Context) (*string, diag.Diagnostics) {
...
+			}
+			if password, ok := esConfig["password"]; ok {
+				config.Password = password.(string)
+			}
+			if apikey, ok := esConfig["api_key"]; ok {
+				config.APIKey = apikey.(string)
+			}
+
+			// default endpoints taken from Env if set
+			if endpoints := os.Getenv("ELASTICSEARCH_ENDPOINTS"); endpoints != "" {
```
tobio: We should only be looking at this environment variable in the root provider case. Probably just need an argument here to say `useEnvAsDefault` or similar.

internal/elasticsearch/security/user_data_source.go
```diff
@@ -63,10 +63,9 @@ func DataSourceUser() *schema.Resource {
 }
 
 func dataSourceSecurityUserRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	var diags diag.Diagnostics
-	client, err := clients.NewApiClient(d, meta)
-	if err != nil {
-		return diag.FromErr(err)
+	client, diags := clients.NewApiClient(d, meta)
```
tobio, line 66: These look like they were fun :)

tobio (commented):
Looks much better, some minor nits.

internal/clients/api_client.go
webfella: Cheers, I wasn't sure if it was fine as a default for the overrides so I took them out.

I saw env vars for username and password used for `NewAcceptanceTestingClient`, should I bring those in too or those are only supported for testing?

internal/elasticsearch/security/user_data_source.go
webfella: Haha well there was something satisfying about checking them off and have the files turn green

internal/clients/api_client.go
webfella: I had it cleaner like this initially, but the acceptance tests were failing without the condition for `elasticsearch_connection` (it worked without it for the case from `elasticsearch`).

I'll try again to verify that was the case and not just the state I had it in at the time.

internal/clients/api_client.go
tobio: They're magic'd in as part of the provider [schema](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/schema/connection.go#L43). `endpoints` doesn't get the same treatment since it's a list in the schema and a CSV string in the env variable. Probably could set it with `DefaultFunc` tbh.

Issue #109, opened 2022-05-14
[Bug] include_global_state parameter ignored for elasticstack_elasticsearch_snapshot_lifecycle


SamuZad:
**Describe the bug**
The include_global_state parameter is ignored when creating a elasticstack_elasticsearch_snapshot_lifecycle

**To Reproduce**
Both of these will produce a snapshot policy where the global state is included:

terraform {
  required_providers {
    elasticstack = {
      source = "elastic/elasticstack"
      version = "0.3.3"
    }
  }
}

provider "elasticstack" {
  elasticsearch {
    endpoints = ["http://localhost:9200"]
  }
}

resource "elasticstack_elasticsearch_snapshot_lifecycle" "proba" {
  name = "proba"

  schedule      = "0 0 6 * * ?"
  snapshot_name = "<daily-snap-{now/d}>"
  repository    = "s3_backup"

  indices              = ["logstash*", "mesos*", "cluster-events*"]
  ignore_unavailable   = false
  include_global_state = false

  expire_after = "3d"
  min_count    = 1
  max_count    = 3
}

-----------------------------

terraform {
  required_providers {
    elasticstack = {
      source = "elastic/elasticstack"
      version = "0.3.3"
    }
  }
}

provider "elasticstack" {
  elasticsearch {
    endpoints = ["http://localhost:9200"]
  }
}

resource "elasticstack_elasticsearch_snapshot_lifecycle" "proba" {
  name = "proba"

  schedule      = "0 0 6 * * ?"
  snapshot_name = "<daily-snap-{now/d}>"
  repository    = "s3_backup"

  indices              = ["logstash*", "mesos*", "cluster-events*"]
  ignore_unavailable   = false
  include_global_state = true

  expire_after = "3d"
  min_count    = 1
  max_count    = 3
}

**Expected behavior**
I expect that the include_global_state parameter lets me disable the snapshotting of global state indices

**Versions (please complete the following information):**
 - OS: Mac Darwin
 - Terraform Version: v1.1.9
 - Provider version: v0.3.3
 - Elasticsearch Version: 7.16.3

PR #224, opened 2022-12-14
Apply zeroish values in SLM policies

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/109

`GetOk` returns `ok=false` for zero valued attributes, so this option was never being set in the policy and the default (true) was applied.


webfella (approved):
LGTM

PR #226, opened 2022-12-14
Support Kibana connections

webfella:
Adds support for Kibana within the provider connection block.


provider/provider.go
```diff
@@ -21,9 +21,51 @@ func init() {
...
+				Optional:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"username": {
+							Description:  "Username to use for API authentication to Kibana.",
+							Type:         schema.TypeString,
+							Optional:     true,
+							DefaultFunc:  schema.EnvDefaultFunc("ELASTICSEARCH_USERNAME", nil),
+							RequiredWith: []string{"kibana.0.password"},
+							ExactlyOneOf: []string{"elasticsearch.0.username", "kibana.0.username"},
```
tobio: This doesn't seem right. We should default to using the ES credentials if no Kibana credentials are supplied, but people should be able to use one account with ES and another account for Kibana.

webfella: This was my initial attempt at finding a way to support [example 1 and 2 here described in the description](https://github.com/elastic/terraform-provider-elasticstack/issues/100), but I wasn't sure so just left it to move forward on integrating the go API. for Kibana.

I'm happy to jump on a call to take stock of where things are before moving forward though to make sure we're on the same page?

internal/schema/connection.go
```diff
@@ -123,6 +123,48 @@ func GetConnectionSchema(keyName string, isProviderConfiguration bool) *schema.S
...
+				},
+				"password": {
+					Description:  "Password to use for API authentication to Kibana.",
+					Type:         schema.TypeString,
+					Optional:     true,
+					Sensitive:    true,
+					RequiredWith: []string{"kibana.0.username"},
+				},
+				"endpoints": {
+					Description: "A comma-separated list of endpoints where the terraform provider will point to, this must include the http(s) schema and port number.",
```
tobio:
```suggestion
					Description: "A list of endpoints where the terraform provider will point to, this must include the http(s) schema and port number.",
```

internal/clients/api_client.go
```diff
@@ -307,5 +346,87 @@ func newEsApiClient(d *schema.ResourceData, key string, version string, useEnvAs
...
+			Severity: diag.Error,
+			Summary:  "Unable to create Kibana client",
+			Detail:   err.Error(),
+		})
+	}
+
+	return kib, diags
+}
+
+func newApiClient(d *schema.ResourceData, version string, useEnvAsDefault bool, esKey string) (*ApiClient, diag.Diagnostics) {
```
tobio: We use this function when creating a client from the [`elasticsearch_connection` block](https://github.com/elastic/terraform-provider-elasticstack/blob/f764db23c71f1d723f63cd85aea772c30dd9dc14/internal/clients/api_client.go#L104), but the Kibana section won't work there. 

I think we should change the logic in `NewApiClient` to:
* Grab the default client
* Return that if the elasticsearch_connection block doesn't exist
* Otherwise build a new *ES* client using the `elasticsearch_connection` block
* Return a new `&ApiClient{elasticsearch: clientFromEsConnectionBlock, kibana: defaultClient.kibana...}`

internal/clients/api_client.go
```diff
@@ -307,5 +346,87 @@ func newEsApiClient(d *schema.ResourceData, key string, version string, useEnvAs
...
+	var diags diag.Diagnostics
+	baseConfig := buildBaseConfig(d, version, useEnvAsDefault)
```
tobio: This will do the wrong thing when it's used against `elasticsearch_connection`. That'll have to change with the previous comment but we need to make sure we build the base config from the correct config block.

internal/clients/api_client.go
```diff
@@ -307,5 +346,87 @@ func newEsApiClient(d *schema.ResourceData, key string, version string, useEnvAs
...
+	if !ok {
+		return nil, diags
+	}
+
+	config := kibana.Config{}
+
+	// Use ES details by default
+	config.Username = baseConfig.Username
+	config.Password = baseConfig.Password
+	config.Address = baseConfig.Addresses[0]
```
tobio: Kibana and ES won't share an address. It only really makes sense to share the credentials. 

```suggestion
	// Use ES details by default
	config := kibana.Config{
		Username: baseConfig.Username,
		Password: baseConfig.Password,
	}
```

internal/clients/api_client.go
```diff
@@ -214,82 +223,112 @@ func newEsApiClient(d *schema.ResourceData, key string, version string, useEnvAs
...
+	esConn, ok := d.GetOk(key)
+	if !ok {
+		return nil, diags
+	}
+
+	config := elasticsearch.Config{}
+	config.Header = baseConfig.Header
+	config.Username = baseConfig.Username
+	config.Password = baseConfig.Password
+	config.Addresses = baseConfig.Addresses
```
tobio:
```suggestion
	config := elasticsearch.Config{
		Header: baseConfig.Header,
		Username = baseConfig.Username,
		Password = baseConfig.Password,
		Addresses = baseConfig.Addresses,
	}
```

internal/clients/api_client.go
```diff
@@ -214,82 +223,112 @@ func newEsApiClient(d *schema.ResourceData, key string, version string, useEnvAs
 					for _, e := range strings.Split(endpoints, ",") {
 						addrs = append(addrs, strings.TrimSpace(e))
 					}
-					config.Addresses = addrs
+					baseConfig.Addresses = addrs
 				}
 			}
 
-			if endpoints, ok := esConfig["endpoints"]; ok && len(endpoints.([]interface{})) > 0 {
+			if endpoints, ok := config["endpoints"]; ok && len(endpoints.([]interface{})) > 0 {
```
tobio: As before, the endpoints aren't shared. We should probably move this logic into the newEsClient func instead and just have the credentials in the shared config.

internal/clients/api_client.go
```diff
@@ -116,7 +118,11 @@ func ensureTLSClientConfig(config *elasticsearch.Config) *tls.Config {
 }
 
 func (a *ApiClient) GetESClient() *elasticsearch.Client {
-	return a.es
+	return a.elasticsearch
```
tobio: There's nothing really enforcing the presence of these clients anymore, it's valid to only configure the Kibana endpoint, but attempt to manage ES resources for instance. 

IMO we should update the `Get*Client` funcs to return an error if the client is nil.

internal/clients/api_client.go
webfella: Thanks for the suggestions mate, was passing around the key to support `elasticsearch_connection` but didn't think of the consequences of Kibana being in there 👍

internal/clients/api_client.go
webfella: Cheers, these were the examples I said I knew could be done a better way 🙏

tobio (approved):
LGTM, we'll get some acceptance tests around this with the initial Kibana resources.

tobio (commented):
Can we add a changelog entry for this one since there's a bunch of changes to how the ES connection is built.

tobio (approved):
Time may change me
But I can't trace time

PR #229, opened 2022-12-16
Add `mapping_coerce` field to index resource

### LINKED ISSUE #228: [Feature] Make it possible to set `index.mapping.coerce` in index resource
https://www.elastic.co/guide/en/elasticsearch/reference/current/coerce.html#coerce-setting

k-yomo:
Resolves #228

### DIFF
--- a/internal/elasticsearch/index/index.go
+++ b/internal/elasticsearch/index/index.go
@@ -60,6 +60,7 @@ var (
 		"default_pipeline":                       schema.TypeString,
 		"final_pipeline":                         schema.TypeString,
 		"unassigned.node_left.delayed_timeout":   schema.TypeString,
+		"mapping.coerce":                         schema.TypeBool,
 		"search.slowlog.threshold.query.warn":    schema.TypeString,
 		"search.slowlog.threshold.query.info":    schema.TypeString,
 		"search.slowlog.threshold.query.debug":   schema.TypeString,
@@ -306,6 +307,12 @@ func ResourceIndex() *schema.Resource {
 			Type:        schema.TypeString,
 			Description: "Time to delay the allocation of replica shards which become unassigned because a node has left, in time units, e.g. `10s`",
 			Optional:    true,
+		},
+		"mapping_coerce": {
+			Type:        schema.TypeBool,
+			Description: "Set index level coercion setting that is applied to all mapping types. Defaults to `true`.",
+			Optional:    true,
+			Default:     true,
 		},
 		"search_slowlog_threshold_query_warn": {
 			Type:        schema.TypeString,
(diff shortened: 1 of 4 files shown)


internal/elasticsearch/index/index.go
```diff
@@ -307,6 +308,12 @@ func ResourceIndex() *schema.Resource {
 			Description: "Time to delay the allocation of replica shards which become unassigned because a node has left, in time units, e.g. `10s`",
 			Optional:    true,
 		},
+		"mapping_coerce": {
+			Type:        schema.TypeBool,
+			Description: "Set index level coercion setting that is applied to all mapping types. Defaults to `true`.",
+			Optional:    true,
+			Default:     true,
```
```suggestion
			Default:     false,
```

ES defaults to false and this would attempt to update the setting on any existing managed indices on upgrade. This setting is also static, so the attempted change would cause TF to fail to apply the change.

internal/elasticsearch/index/index.go
```diff
@@ -60,6 +60,7 @@ var (
 		"default_pipeline":                       schema.TypeString,
 		"final_pipeline":                         schema.TypeString,
 		"unassigned.node_left.delayed_timeout":   schema.TypeString,
+		"mapping.coerce":                         schema.TypeBool,
```
tobio: This should be in `staticSettingsKeys`, this setting cannot be updated on an open index.

internal/elasticsearch/index/index.go
```diff
@@ -307,6 +308,12 @@ func ResourceIndex() *schema.Resource {
 			Description: "Time to delay the allocation of replica shards which become unassigned because a node has left, in time units, e.g. `10s`",
 			Optional:    true,
 		},
+		"mapping_coerce": {
+			Type:        schema.TypeBool,
+			Description: "Set index level coercion setting that is applied to all mapping types. Defaults to `true`.",
```
```suggestion
			Description: "Set index level coercion setting that is applied to all mapping types. Defaults to `false`.",
```

internal/elasticsearch/index/index.go
k-yomo: Oh I thought it's true by checking number field, but it looks like global default is `false` but for some types, default is `true`.  Then I think we shouldn't have default value here, so let me remove it🙏

https://github.com/elastic/elasticsearch/blob/006e2acee33fc535578fbbb271f78fd77331d909/server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java#L63

https://github.com/elastic/elasticsearch/blob/006e2acee33fc535578fbbb271f78fd77331d909/server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java#L79

tobio (approved):

PR #232, opened 2022-12-16
Don't recreate when field is removed

### LINKED ISSUE #227: [Feature] Don't recreate when there is a diff in dynamic object field
When we have dynamic field like `metadata` below, it can be `ForceNew` because there will be type diff from the latest mapping which have dynamically added inner fields.
It would be great if we could suppress those diffs and not cause recreate.
```terraform
resource "elasticstack_elasticsearch_index" "test" {
  name     = "test"

  mappings = jsonencode({
    "dynamic" : false,
    "properties" : {
      "name" : {
        "type" : "keyword"
      },
      "metadata" : {
        "type" : "object",
        "dynamic" : true
      }
    },
    "dynamic_templates" : [
      {
        "metadata" : {
          "path_match" : "metadata.*",
          "mapping" : {
            "type" : "keyword",
            "fields" : {
              "text" : {
                "type" : "text",
                "analyzer" : "standard",
                "search_analyzer" : "standard"
              }
            }
          }
        }
      }
  }
```

### Diff
```diff
                  ~ metadata                       = {
                      ~ dynamic    = "true" -> true
                      - properties = {
                          ...
                          - someMetadata                       = {
                              - fields  = {
                                  - text = {
                                      - analyzer        = "standard"
                                      - search_analyzer = "standard"
                                      - type            = "text"
                                    }
                                }
                              - type    = "keyword"
                            }
                          },
                          ...
```

`elasticstack_elasticsearch_index`

Supress diff in dynamic object field.

**Additional context**

k-yomo:
Resolves #227 

This PR changes not to recreate when field is removed in a new mapping (= mapping defined in terraform), not to cause recreate mainly for dynamically added fields.
Ignoring field removal is the default behaviour of Elasticsearch, so it'll be aligned wit Elasticsearch too.

📝 This PR is branched from https://github.com/elastic/terraform-provider-elasticstack/pull/231

### DIFF
--- a/internal/elasticsearch/index/index.go
+++ b/internal/elasticsearch/index/index.go
@@ -909,7 +909,7 @@ func IsMappingForceNewRequired(ctx context.Context, old map[string]interface{}, 
 		newFieldSettings, ok := new[k]
 		// When field is removed, it'll be ignored in elasticsearch
 		if !ok {
-			tflog.Warn(ctx, fmt.Sprintf("removing %s field in mappings is ignored, if you neeed to remove it completely, please recreate the index", k))
+			tflog.Warn(ctx, fmt.Sprintf("removing %s field in mappings is ignored. Re-index to remove the field completely.", k))
 			continue
 		}
 		newSettings := newFieldSettings.(map[string]interface{})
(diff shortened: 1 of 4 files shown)


internal/elasticsearch/index/index.go
```diff
@@ -900,37 +903,37 @@ func resourceIndexDelete(ctx context.Context, d *schema.ResourceData, meta inter
...
-				if ns, ok := newSettings["type"]; ok {
-					if !reflect.DeepEqual(s, ns) {
-						return true
-					}
-					continue
-				} else {
+		newFieldSettings, ok := new[k]
+		// When field is removed, it'll be ignored in elasticsearch
+		if !ok {
+			tflog.Warn(ctx, fmt.Sprintf("removing %s field in mappings is ignored, if you neeed to remove it completely, please recreate the index", k))
```
```suggestion
			tflog.Warn(ctx, fmt.Sprintf("removing %s field in mappings is ignored. Re-index to remove the field completely.", k))
```

tobio (commented):
Small text update. Can we add an acceptance test around this as well?

tobio (approved):

PR #234, opened 2022-12-18
WIP: Add framework provider

k-yomo:
https://github.com/elastic/terraform-provider-elasticstack/issues/214

This PR adds provider for terraform-plugin-framework.

### TODO
- [x] Revert https://github.com/elastic/terraform-provider-elasticstack/pull/234/commits/683e8e4251bbc457dfd2a03624f45f617b79c16a when the below issue is fixed.
https://github.com/hashicorp/terraform-plugin-mux/issues/117


internal/clients/api_client.go
```diff
@@ -180,6 +213,113 @@ func (a *ApiClient) ClusterID(ctx context.Context) (*string, diag.Diagnostics) {
 	return nil, diags
 }
 
+func NewFWEsApiClient(ctx context.Context, esConn *ElasticSearchConnection, version string, useEnvAsDefault bool) (*ApiClient, fwdiag.Diagnostics) {
```
tobio, line 305:
```suggestion
func NewFWApiClient(ctx context.Context, esConn *ElasticSearchConnection, version string, useEnvAsDefault bool) (*ApiClient, fwdiag.Diagnostics) {
```

We're working on adding Kibana support here too

internal/clients/api_client.go
```diff
@@ -180,6 +213,113 @@ func (a *ApiClient) ClusterID(ctx context.Context) (*string, diag.Diagnostics) {
...
+		if endpoints := os.Getenv("ELASTICSEARCH_ENDPOINTS"); endpoints != "" {
+			for _, e := range strings.Split(endpoints, ",") {
+				addrs = append(addrs, strings.TrimSpace(e))
+			}
+		}
+	}
+	config.Addresses = addrs
+
+	envInsecure, _ := strconv.ParseBool(os.Getenv("ELASTICSEARCH_INSECURE"))
+	if esConn.Insecure.ValueBool() || envInsecure {
```
tobio, line 327: So this is inconsistent with the other settings, where the environment value is checked even when the value is specified in the provider configuration. I think we should be consistent here.

tobio:
This will conflict a bunch with https://github.com/elastic/terraform-provider-elasticstack/pull/226. @k-yomo if you want to keep working on this one right now that's fine, but it might be easier if we hold off here for a couple of weeks until #226 is merged.

internal/clients/api_client.go
```diff
@@ -93,6 +97,35 @@ func NewAcceptanceTestingClient() (*ApiClient, error) {
 
 const esConnectionKey string = "elasticsearch_connection"
 
+type ElasticSearchConnection struct {
```
tobio, line 139: IMO it would be cleaner to have all functions which build a new client parse the relevant config into this struct and then call a single function which builds an `elasticsearch.Config{}` instance from this struct.

It would mean consolidating all the CA, TLS and debugging code. Given the different diagnostic types we'd have to return a plain error but I think that's ok. 

WDYT?

tobio, line 139: nit: 
```suggestion
type ElasticsearchConnection struct {
```

k-yomo:
@tobio 
Oh I see. Thank you  for the heads-up!  
I'll work on this after #226 is done!

internal/schema/connection.go
```diff
@@ -3,9 +3,128 @@ package schema
...
+					Optional:            true,
+					Sensitive:           true,
+					Validators:          passwordValidators,
+				},
+				"api_key": fwschema.StringAttribute{
+					MarkdownDescription: "API Key to use for authentication to Elasticsearch",
+					Optional:            true,
+					Sensitive:           true,
+					Validators: []validator.String{
+						stringvalidator.ConflictsWith(path.MatchRoot(usernamePath)),
```
tobio, line 55: Can we instead use a relative path matcher in the new schema? 

```suggestion
						stringvalidator.ConflictsWith(path.MatchRelative().AtParent().AtName("username")),
```

go.mod
```diff
@@ -48,11 +50,13 @@ require (
...
+	golang.org/x/sys v0.0.0-20220722155257-8c9f86f7a55f // indirect
+	golang.org/x/text v0.4.0 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
 	google.golang.org/genproto v0.0.0-20220608133413-ed9918b62aac // indirect
-	google.golang.org/grpc v1.50.1 // indirect
+	google.golang.org/grpc v1.51.0 // indirect
 	google.golang.org/protobuf v1.28.1 // indirect
 )
+
+replace github.com/hashicorp/terraform-plugin-mux v0.7.0 => github.com/k-yomo/terraform-plugin-mux v0.0.0-20221218141521-7a01a5a64e24
```
tobio: Just flagging it so it doesn't get merged like this. Moving the maxitems validation to code in the short term is ok if getting your PR merged into `terraform-plugin-mux` turns out to be too slow.

k-yomo: The below update contains the fix!
https://github.com/elastic/terraform-provider-elasticstack/pull/235

k-yomo:
Sorry for no updates here for a while, I'll restart working on this PR🙏

PR #244, opened 2023-01-13
Add support for index creation query params

### LINKED ISSUE #242: [Feature] Support index creation query parameters
Currently there is no way to set query parameter options when creating an index

Resource
`elasticstack_elasticsearch_index`:  https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/elasticsearch_index

ES API
`Create index API`:  https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-create-index.html

Make available in the Terraform resource the query parameter options for the API:  https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-create-index.html#indices-create-api-query-params

The parameters `timeout`, `master_timeout`, and `include_type_name` cannot be set through index settings, and so are currently inaccessible in Terraform.

Since the majority of the query parameters cannot be defined by index settings, to use them you must invoke the ES API directly instead of using the Terraform provider resource.

k-yomo:
Resolve #242 

This PR adds support for the following query params that we can set for index creation.
- `include_type_name`
- `wait_for_active_shards`
- `master_timeout`
- `timeout`
https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-create-index.html#indices-create-api-query-params

📝 `master_timeout` and `timeout` are used in the other apis, so it might be better to make it reusable.
https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-put-lifecycle.html#ilm-put-lifecycle-query-params

### DIFF
--- a/internal/clients/elasticsearch/index.go
+++ b/internal/clients/elasticsearch/index.go
@@ -12,7 +12,6 @@ import (
 	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
 	"github.com/elastic/terraform-provider-elasticstack/internal/models"
 	"github.com/elastic/terraform-provider-elasticstack/internal/utils"
-	"github.com/hashicorp/go-version"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 )
 
@@ -210,8 +209,6 @@ func DeleteIndexTemplate(ctx context.Context, apiClient *clients.ApiClient, temp
 	return diags
 }
 
-var includeTypeNameUnsupportedVersion = version.Must(version.NewVersion("8.0.0"))
-
 func PutIndex(ctx context.Context, apiClient *clients.ApiClient, index *models.Index, params *models.PutIndexParams) diag.Diagnostics {
 	var diags diag.Diagnostics
 	indexBytes, err := json.Marshal(index)
@@ -226,11 +223,7 @@ func PutIndex(ctx context.Context, apiClient *clients.ApiClient, index *models.I
 		apiClient.GetESClient().Indices.Create.WithMasterTimeout(params.MasterTimeout),
 		apiClient.GetESClient().Indices.Create.WithTimeout(params.Timeout),
 	}
-	serverVersion, diags := apiClient.ServerVersion(ctx)
-	if diags.HasError() {
-		return diags
-	}
-	if serverVersion.LessThan(includeTypeNameUnsupportedVersion) {
+	if params.IncludeTypeName {
 		opts = append(opts, apiClient.GetESClient().Indices.Create.WithIncludeTypeName(params.IncludeTypeName))
 	}
 	res, err := apiClient.GetESClient().Indices.Create(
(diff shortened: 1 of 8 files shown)


internal/clients/elasticsearch/index.go
```diff
@@ -208,15 +210,33 @@ func DeleteIndexTemplate(ctx context.Context, apiClient *clients.ApiClient, temp
 	return diags
 }
 
-func PutIndex(ctx context.Context, apiClient *clients.ApiClient, index *models.Index) diag.Diagnostics {
+var includeTypeNameUnsupportedVersion = version.Must(version.NewVersion("8.0.0"))
```
tobio: nit: IMO it would be nice to include if this is an upper or lower bound in the name. 

```suggestion
var includeTypeNameMinUnsupportedVersion = version.Must(version.NewVersion("8.0.0"))
```

internal/elasticsearch/index/index.go
```diff
@@ -716,7 +743,23 @@ func resourceIndexCreate(ctx context.Context, d *schema.ResourceData, meta inter
 		}
 	}
 
-	if diags := elasticsearch.PutIndex(ctx, client, &index); diags.HasError() {
+	params := models.PutIndexParams{
+		WaitForActiveShards: d.Get("wait_for_active_shards").(string),
+		IncludeTypeName:     d.Get("include_type_name").(bool),
```
tobio, line 751: I think we should throw an error if `include_type_name = true` when running against an 8.x server. IMO it avoids potential confusion for users. 

Personally, I would probably do all the version checking in here, switch the `IncludeTypeName` struct property to a pointer and then set it in the create request if it's not nil in the params but I don't feel strongly about it.

internal/elasticsearch/index/index.go
```diff
@@ -519,6 +520,32 @@ If specified, this mapping can include: field names, [field data types](https://
 			Type:        schema.TypeString,
 			Computed:    true,
 		},
+		"include_type_name": {
+			Type:        schema.TypeBool,
+			Description: "If true, a mapping type is expected in the body of mappings. Defaults to false.",
```
tobio: We should detail the version constraint here
```suggestion
			Description: "If true, a mapping type is expected in the body of mappings. Defaults to false. Supported for Elasticsearch 7.x",
```

tobio (commented):
Nice, couple of minor comments

internal/elasticsearch/index/index.go
k-yomo: Fixed to return error when the version is unsupported!

> Personally, I would probably do all the version checking in here, switch the IncludeTypeName struct property to a pointer and then set it in the create request if it's not nil in the params but I don't feel strongly about it.

Agreed, moved the logic here.  We don't have a way to know if the value is zero value  (`false`) or  not set , I simply make it evaluate and set when `include_type_name = true`.

tobio (approved):
Nice, thanks!

PR #246, opened 2023-01-13
Cache cluster info to speed up application

k-yomo:
Cluster info is retrieved for every resource to use cluster uuid now.  
This PR caches cluster info to be able to speed up it.


tobio (approved):
Nice 🙂

PR #260, opened 2023-01-26
Add outputs to example and don't pin version

wandergeek:
This PR adds outputs to the cloud example and removes the pinned provider versions. This also makes the API key sensitive so it doesn't inadvertently print in outputs.


docs/guides/elasticstack-and-cloud.md
```diff
@@ -22,12 +22,10 @@ terraform {
 
   required_providers {
     ec = {
-      source  = "elastic/ec"
-      version = "~>0.3.0"
+      source = "elastic/ec"
     }
     elasticstack = {
-      source  = "elastic/elasticstack"
-      version = "~>0.5.0"
```
tobio, line 30: Wonder if it would be better to have `version = "~>0.5"` here so we're requiring a provider version that will work with the example?

tobio (approved):

docs/guides/elasticstack-and-cloud.md
wandergeek: Yeah that makes sense

tobio (approved):
LGTM

Issue #261, opened 2023-01-31
[Bug] Can not Create Cluster Settings on 8.5.3


Nirn917:
**Describe the bug**
The plugin crashes when we try to apply the creation of cluster settings on elasticsearch 8.5.3, it works well on elasticsearch 7.17.

**To Reproduce**
1. TF configuration used 
```
resource "elasticstack_elasticsearch_cluster_settings" "my_cluster_settings-dev" {
  persistent {
    setting {
        name = "action.auto_create_index"
        value = "true"
    }      
    setting {
        name = "xpack.monitoring.elasticsearch.collection.enabled"
        value = "true"
    }
    setting {
        name = "xpack.monitoring.collection.enabled"
        value = "true"
    }
  }

  transient {
    setting {
      name  = "action.auto_create_index"
      value = "true"
    }
    setting {
        name = "indices.recovery.max_bytes_per_sec"
        value = "75mb"
    }
  }
}
```

2. TF operations to execute to get the error : Terraform apply this config on a elasticsearch 8.5.3 cluster
3. See the error in the output : 
```
tack trace from the terraform-provider-elasticstack_v0.5.0 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xecca4e]

goroutine 84 [running]:
github.com/elastic/terraform-provider-elasticstack/internal/clients/elasticsearch.PutSettings({0x132c428?, 0xc000710480}, 0xc00033fd10, 0xc00035c9c0?)
	github.com/elastic/terraform-provider-elasticstack/internal/clients/elasticsearch/cluster.go:157 +0x1ce
github.com/elastic/terraform-provider-elasticstack/internal/elasticsearch/cluster.resourceClusterSettingsPut({0x132c428, 0xc000710480}, 0x0?, {0x10965a0, 0xc00033fd10})
	github.com/elastic/terraform-provider-elasticstack/internal/elasticsearch/cluster/settings.go:112 +0x42a
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).create(0xc000414620, {0x132c460, 0xc0007c1a70}, 0xd?, {0x10965a0, 0xc00033fd10})
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.24.1/helper/schema/resource.go:707 +0x12e
[...]

PR #262, opened 2023-02-01
Actually return errors which occur

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/261

This won't actually fix the linked issue, but it will stop the provider crash and at least return a more helpful error. 

Pretty sad to see how often we were actually just accidentally swallowing errors 🤦


webfella (approved):
🤐

PR #271, opened 2023-02-15
Add Kibana container to the acceptance tests

tobio:


webfella (approved):
Code LGTM, seems like we need to add in a pipeline here for buildkite?

tobio:
> seems like we need to add in a pipeline here for buildkite

Yea, it should only be for the actual release job for now though since everything else is in GH actions.

PR #272, opened 2023-02-15
Kibana spaces

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/5

PR #274, opened 2023-02-18
Fix broken make target docker-testacc due to missing \

jonjitsu:
The docker-testacc target was broken by #271 due to a missing ending slash \

The `docker-testacc` target runs successfully.


tobio:
Fixed in https://github.com/elastic/terraform-provider-elasticstack/pull/272

PR #284, opened 2023-03-07
Support for transforms

### LINKED ISSUE #49: [Feature] Support for Transforms
Add the ability to add and maintain transforms

Cristigeo:
This PR is intended to add support for transforms (https://www.elastic.co/guide/en/elasticsearch/reference/current/transforms.html).

This is my first attempt to contribute to an open-source project; being new to this, I apologize in advance if I inadvertently break some rules/guidelines. Also, full disclaimer, Go is not my main programming language and I haven't done any previous coding for Terraform providers.

Feedback is welcome (obviously). While in its current state the code is functional (at least during the minimal tests I've performed), I'd like to point out a couple of design choices (in the current code) that may not be optimal:
- most configuration is done through dedicated resources block, but the `pivot` and `latest` props are handled as raw json content;
- the start/stop functionality is mapped through the `enabled` property. A different (better?) choice might have been a dedicated resource (such as `elasticstack_elasticsearch_transform_state`), because start/stop is a different API flow (not part of the usual CRUD).

Open points:

- [x]  drift detection
- [x]  support for `import` functionality
- [x]  improved documentation / acceptance test
- [ ]  tests on different ES versions
- [x]  cleanup

Resolves #49


tobio:
> the start/stop functionality is mapped through the enabled property. A different (better?) choice might have been a dedicated resource (such as elasticstack_elasticsearch_transform_state), because start/stop is a different API flow (not part of the usual CRUD).

Both approaches will work, and there doesn't seem to be an idiomatic approach here (some providers have a property on the resource, some have a separate resource). If you're happy with the property then we can start with that.

docs/resources/elasticsearch_transform.md
```diff
@@ -0,0 +1,189 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: 'elasticstack_elasticsearch_transform Resource - terraform-provider-elasticstack'
+subcategory: ''
```
tobio: So a painful part of these generated docs is the inability to set the subcategory [from code](https://github.com/hashicorp/terraform-plugin-docs/issues/156).

We've got templates for all these docs [here](https://github.com/elastic/terraform-provider-elasticstack/blob/main/templates/resources/elasticsearch_index_lifecycle.md.tmpl) basically just so we can have the subcategory.

internal/clients/elasticsearch/index.go
```diff
@@ -544,3 +544,158 @@ func DeleteIngestPipeline(ctx context.Context, apiClient *clients.ApiClient, nam
 	}
 	return diags
 }
+
+func PutTransform(ctx context.Context, apiClient *clients.ApiClient, transform *models.Transform, params *models.PutTransformParams) diag.Diagnostics {
+	fmt.Println("entering PutTransform for", transform.Name)
```
tobio: We'd either want to remove this print statements of move them to tflog as Debug logs.

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
+package transform
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	//"reflect"
```
tobio: We should remove these commented imports `make format` should remove unused imports IIRC.

internal/clients/elasticsearch/index.go
```diff
@@ -544,3 +544,158 @@ func DeleteIngestPipeline(ctx context.Context, apiClient *clients.ApiClient, nam
...
+	if diags := utils.CheckError(res, fmt.Sprintf("Unable to create transform: %s", transform.Name)); diags.HasError() {
+		return diags
+	}
+
+	if params.Enabled {
+		startOptions := []func(*esapi.TransformStartTransformRequest){
+			esClient.TransformStartTransform.WithContext(ctx),
+			esClient.TransformStartTransform.WithTimeout(params.Timeout),
+		}
+		_, err := esClient.TransformStartTransform(transform.Name, startOptions...)
```
tobio: We should be checking for an error response here as well. Similar to the `utils.CheckError` call on ln 575.

internal/clients/elasticsearch/index.go
```diff
@@ -544,3 +544,158 @@ func DeleteIngestPipeline(ctx context.Context, apiClient *clients.ApiClient, nam
...
+	if diags := utils.CheckError(res, fmt.Sprintf("Unable to update transform: %s", transform.Name)); diags.HasError() {
+		return diags
+	}
+
+	if params.Enabled {
+		startOptions := []func(*esapi.TransformStartTransformRequest){
+			esClient.TransformStartTransform.WithContext(ctx),
+			esClient.TransformStartTransform.WithTimeout(params.Timeout),
+		}
+		_, err := esClient.TransformStartTransform(transform.Name, startOptions...)
```
tobio: Similarly should check for an error response here

internal/clients/elasticsearch/index.go
```diff
@@ -544,3 +544,158 @@ func DeleteIngestPipeline(ctx context.Context, apiClient *clients.ApiClient, nam
...
+		if err != nil {
+			return diag.FromErr(err)
+		}
+	} else {
+		stopOptions := []func(*esapi.TransformStopTransformRequest){
+			esClient.TransformStopTransform.WithContext(ctx),
+			esClient.TransformStopTransform.WithTimeout(params.Timeout),
+		}
+		_, err := esClient.TransformStopTransform(transform.Name, stopOptions...)
```
tobio: Similarly should check for an error response here

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+								"field": {
+									Description: "The date field that is used to calculate the age of the document.",
+									Type:        schema.TypeString,
+									Required:    true,
+								},
+								"max_age": {
+									Description:  "Specifies the maximum age of a document in the destination index.",
+									Type:         schema.TypeString,
+									Required:     true,
+									ValidateFunc: utils.StringIsElasticDuration,
```
tobio, line 183: ❤️

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+		},
+		"enabled": {
+			Type:        schema.TypeBool,
+			Description: "Controls wether the transform is started or stopped. Default is `false` (stopped).",
+			Optional:    true,
+			Default:     false,
+		},
+	}
+
+	utils.AddConnectionSchema(transformSchema)
```
tobio:
```suggestion
```

The resource level connection schema has been deprecated, no need to add it to new resources.

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+		DeferValidation: d.Get("defer_validation").(bool),
+	}
+
+	timeout, err := time.ParseDuration(d.Get("timeout").(string))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	params.Timeout = timeout
+
+	params.Enabled = d.Get("enabled").(bool)
```
tobio: nit

```suggestion}
	timeout, err := time.ParseDuration(d.Get("timeout").(string))
	if err != nil {
		return diag.FromErr(err)
	}
	params := models.PutTransformParams{
		DeferValidation: d.Get("defer_validation").(bool),
		Enabled:             d.Get("enabled").(bool),
		Timeout:             timeout,
	}
```

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+	transform, diags := elasticsearch.GetTransform(ctx, client, &transformName)
+	if transform == nil && diags == nil {
+		tflog.Warn(ctx, fmt.Sprintf(`Transform "%s" not found, removing from state`, compId.ResourceId))
+		d.SetId("")
+		return diags
+	}
+	if diags.HasError() {
+		return diags
+	}
+
```
tobio, line 367: We're not actually updating any state fields in the ResourceData here?

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+	updatedTransform, err := getTransformFromResourceData(ctx, d, transformName)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	updatedTransform.Pivot = nil
+	updatedTransform.Latest = nil
```
tobio, line 410: Is this required by the API or is this just added safety? It looks like these are marked as `ForceNew` so Terraform would handle the delete/create if there were changes.

I'm not suggesting a change here, just curious :)

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+		DeferValidation: d.Get("defer_validation").(bool),
+	}
+
+	timeout, err := time.ParseDuration(d.Get("timeout").(string))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	params.Timeout = timeout
+
+	params.Enabled = d.Get("enabled").(bool)
```
tobio: nit

```suggestion}
	timeout, err := time.ParseDuration(d.Get("timeout").(string))
	if err != nil {
		return diag.FromErr(err)
	}
	params := models.UpdateTransformParams{
		DeferValidation: d.Get("defer_validation").(bool),
		Enabled:             d.Get("enabled").(bool),
		Timeout:             timeout,
	}
```

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+			transform.Source.RuntimeMappings = runtimeMappings
+		}
+	}
+
+	if v, ok := d.GetOk("destination"); ok {
+
+		definedDestination := v.([]interface{})[0].(map[string]interface{})
+		transform.Destination = new(models.TransformDestination)
+
+		transform.Destination.Index = definedDestination["index"].(string)
```
tobio:
```suggestion
		transform.Destination = &models.TransformDestination{Index: definedDestination["index"].(string)}
```

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+		}
+		transform.Latest = latest
+	}
+
+	if v, ok := d.GetOk("frequency"); ok {
+		transform.Frequency = v.(string)
+	}
+
+	if v, ok := d.GetOk("metadata"); ok {
+		metadata := make(map[string]interface{})
```
tobio: I think this should be fine as a var declaration?

```suggestion
		var metadata map[string]interface{}
```

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+		if err := json.NewDecoder(strings.NewReader(v.(string))).Decode(&metadata); err != nil {
+			return nil, err
+		}
+		transform.Meta = metadata
+	}
+
+	if v, ok := d.GetOk("retention_policy"); ok && v != nil {
+		definedRetentionPolicy := v.([]interface{})[0].(map[string]interface{})
+		retentionTime := models.TransformRetentionPolicyTime{}
```
tobio: Can move this var into the if block below

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+				retentionTime.MaxAge = ma.(string)
+			}
+			transform.RetentionPolicy = new(models.TransformRetentionPolicy)
+			transform.RetentionPolicy.Time = retentionTime
+		}
+	}
+
+	if v, ok := d.GetOk("sync"); ok {
+		definedRetentionPolicy := v.([]interface{})[0].(map[string]interface{})
+		syncTime := models.TransformSyncTime{}
```
tobio: Similarly this var belongs in the if

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+		if v, ok := definedRetentionPolicy["time"]; ok {
+			var definedRetentionTime = v.([]interface{})[0].(map[string]interface{})
+			if f, ok := definedRetentionTime["field"]; ok {
+				retentionTime.Field = f.(string)
+			}
+			if ma, ok := definedRetentionTime["max_age"]; ok {
+				retentionTime.MaxAge = ma.(string)
+			}
+			transform.RetentionPolicy = new(models.TransformRetentionPolicy)
+			transform.RetentionPolicy.Time = retentionTime
```
tobio:
```suggestion
			transform.RetentionPolicy = &models.TransformRetentionPolicy{
				Time: retentionTime,
			}
```

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+		if v, ok := definedRetentionPolicy["time"]; ok {
+			var definedRetentionTime = v.([]interface{})[0].(map[string]interface{})
+			if f, ok := definedRetentionTime["field"]; ok {
+				syncTime.Field = f.(string)
+			}
+			if d, ok := definedRetentionTime["delay"]; ok {
+				syncTime.Delay = d.(string)
+			}
+			transform.Sync = new(models.TransformSync)
+			transform.Sync.Time = syncTime
```
tobio:
```suggestion
			transform.Sync = &models.TransformSync{
				Time: syncTime,
			}
```

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+		}
+	}
+
+	if v, ok := d.GetOk("settings"); ok {
+		definedSettings := v.([]interface{})[0].(map[string]interface{})
+		settings := models.TransformSettings{}
+		if v, ok := definedSettings["align_checkpoints"]; ok {
+			settings.AlignCheckpoints = new(bool)
+			*settings.AlignCheckpoints = v.(bool)
```
tobio:
```suggestion
			ac := v.(bool)
			settings.AlignCheckpoints = &ac
```

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,548 @@
...
+		}
+		if v, ok := definedSettings["dates_as_epoch_millis"]; ok {
+			settings.DatesAsEpochMillis = new(bool)
+			*settings.DatesAsEpochMillis = v.(bool)
```
tobio:
```suggestion
			d := v.(bool)
			*settings.DatesAsEpochMillis = &d
```

and similarly for below

internal/utils/validation.go
```diff
@@ -18,3 +18,53 @@ func StringIsDuration(i interface{}, k string) (warnings []string, errors []erro
 
 	return nil, nil
 }
+
+// StringIsElasticDuration is a SchemaValidateFunc which tests to make sure the supplied string is valid duration using Elastic time units:
+// d, h, m, s, ms, micros, nanos. (see https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#time-units)
+func StringIsElasticDuration(i interface{}, k string) (warnings []string, errors []error) {
```
tobio, line 25: Can we add some unit tests for this one too.

internal/utils/validation.go
```diff
@@ -18,3 +18,53 @@ func StringIsDuration(i interface{}, k string) (warnings []string, errors []erro
...
+	if !ok {
+		return nil, []error{fmt.Errorf("expected type of %s to be string", k)}
+	}
+
+	if v == "" {
+		return nil, []error{fmt.Errorf("%q contains an invalid duration: [empty]", k)}
+	}
+
+	firstPartCount := 0
+	for v != "" {
```
tobio: I think this whole validation is clearer as a regex (untested, but something like`^[0-9]+(?:\.[0-9]+)?(?:d|h|m|s|ms|micros|nanos)$`) WDYT?

tobio (commented):
This is awesome! Some comments inline.

tobio:
Also can we add a changelog entry [here](https://github.com/elastic/terraform-provider-elasticstack/blob/main/CHANGELOG.md) in this PR

internal/elasticsearch/transform/transform.go
Cristigeo: It's API-related. I'm using the same struct (model), marshaled into a json payload, for both Create and Update API calls. However, as per Elastic API specs, there is a small difference: properties `pivot` and `latest` are not expected on Update operations, even if I were to send the exact same `pivot` or `latest` as on creation.
The alternative would have been to use 2 models, one for Create, one for Update, but that felt like overkill, given how similar they are.

internal/elasticsearch/transform/transform.go
Cristigeo: I haven't identified any need yet, but this might change when I start implementing the import functionality.

Cristigeo:
@tobio , thank you for the feedback, really appreciated. As you can tell, I'm still learning Go :) . I made a couple of pushes to fix most (if not all) of the issues. Not really sure about the "Resolve conversation" policy (who should be pushing those buttons), so I left that untouched.

tobio:
> Not really sure about the "Resolve conversation" policy

It's totally ok for you to hit those buttons. Personally I find that easier since it's clear that you've made changes to address that.

internal/elasticsearch/transform/transform.go
tobio: At the moment, I'd expect the following to not work as expected:

1. Create a transform via TF
2. Update the transform outside of TF via the ES API directly
3. Re-run the TF plan/apply

Expected:
TF corrects the drift

IIUC, currently TF won't see the configuration drift, since we're not updating the current state in the function. In that case, the plan will come back clean (incorrectly) and no changes will be applied.

internal/utils/validation.go
Cristigeo: Clearly better with regex, I don't know what I was thinking 😕

internal/elasticsearch/transform/transform.go
Cristigeo: Indeed, I initially misunderstood the role of the "read" callback; I also got tricked by the fact that the initial tests with a locally-built provider (using the provider_installation / filesystem_mirror config) seemed to work fine for create/update scenarios.
I've added "drift detection" in the todo list.

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,843 @@
...
+			Description: "Specifies whether the transform should deduce the destination index mappings from the transform config.",
+			Type:        schema.TypeBool,
+			Optional:    true,
+			Default:     true,
+		},
+		"docs_per_second": {
+			Description:  "Specifies a limit on the number of input documents per second. Default (unset) value disables throttling.",
+			Type:         schema.TypeFloat,
+			Optional:     true,
+			Default:      -1,
```
tobio: This isn't the real default. It sounds like we can just remove the default value here. 
```suggestion
```

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,843 @@
...
+			ValidateFunc: validation.FloatAtLeast(0),
+		},
+		"max_page_search_size": {
+			Description:  "Defines the initial page size to use for the composite aggregation for each checkpoint. Default is 500.",
+			Type:         schema.TypeInt,
+			Optional:     true,
+			Default:      -1,
```
tobio: This isn't the real default. It sounds like we can just remove the default value here. 
```suggestion
```

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,843 @@
...
+			ValidateFunc: validation.IntBetween(10, 65536),
+		},
+		"num_failure_retries": {
+			Description:  "Defines the number of retries on a recoverable failure before the transform task is marked as failed. The default value is the cluster-level setting num_transform_failure_retries.",
+			Type:         schema.TypeInt,
+			Optional:     true,
+			Default:      -2,
```
tobio: This isn't the real default. It sounds like we can just remove the default value here. 
```suggestion
```

internal/elasticsearch/transform/transform.go
```diff
@@ -0,0 +1,843 @@
...
+	if v, ok := d.GetOk("max_page_search_size"); ok && v.(int) > 10 && isSettingAllowed(ctx, "max_page_search_size", serverVersion) {
+		setSettings = true
+		mpss := v.(int)
+		settings.MaxPageSearchSize = &mpss
+	}
+	if v, ok := d.GetOk("num_failure_retries"); ok && v.(int) >= -1 && isSettingAllowed(ctx, "num_failure_retries", serverVersion) {
+		setSettings = true
+		nfr := v.(int)
+		settings.NumFailureRetries = &nfr
+	}
```
tobio: Are we able to remove the `v.(int) > 10` style conditions in these? Terraform should be validating these options before executing the provider

internal/elasticsearch/transform/transform_test.go
```diff
@@ -0,0 +1,387 @@
...
+  }
+
+	retention_policy {
+    time {
+      field   = "order_date"
+      max_age = "7d"
+    }
+  }
+
+	max_page_search_size = 1000
```
tobio:
```suggestion
	max_page_search_size = 2000
```

It looks like this acceptance test is failing in 7.13 because we're adjusting this setting. The simplest path forward for this PR might be to stop updating this setting.

internal/elasticsearch/transform/transform_test.go
```diff
@@ -0,0 +1,387 @@
...
+					resource.TestCheckResourceAttr("elasticstack_elasticsearch_transform.test_pivot", "defer_validation", "false"),
+				),
+			},
+		},
+	})
+}
+
+// create a transform referencing a non-existing source index;
+// because validations are deferred, this should pass
+func testAccResourceTransformWithPivotCreate(name string) string {
```
tobio, line 109: Can we tidy up the indentation on the TF code in these tests?

internal/elasticsearch/transform/transform_test.go
Cristigeo: Yes, after some investigations, it looks like ES versions below 7.16 have some issues when attempting to update some settings on a transform that was never started (probably because some internal metadata was not created, hence the error).  I've modified the acc test as you suggested.

internal/elasticsearch/transform/transform.go
Cristigeo: That `Default` was an attempt to fix an issue that wasn't really an issue, my bad. I've corrected the use of defaults for several attributes (kept those that would/could lead to "continuous drift" if removed).

tobio (approved):
Excellent work! There's a lot of code in this contribution, thanks for taking the time to improve the provider :)

Issue #282, opened 2023-03-02
[Feature] Allow managing enrich policies


jonjitsu:
**Describe the resource you would like to have implemented.**
This ticket is about increasing the API coverage of this provider to include the enrich policy API.

**Describe the solution you'd like**

```terraform
resource "elasticstack_elasticsearch_enrich_policy" "my-policy" {
  # required
  policy_type = "match"
  indices = []
  match_field = ""
  enrich_fields = []
  # optional
  query = ""
  # maybe 
  execute = false
}
```
By default it could execute any created policies, but through an optional `execute` flag can be configured to not execute.

**Describe alternatives you've considered**
I've used Mastercard/restapi to manage enrich policies but it's clunky since all updates need the `terraform -replace=resource...` mechanism.

I can send a PR for this with some guidance.

tobio:
> I can send a PR for this with some guidance.

👍 that would be great! I'm happy to help with any reviews or guidance on how the provider is put together. The existing Elasticsearch resources are all within [this](https://github.com/elastic/terraform-provider-elasticstack/tree/main/internal/elasticsearch) package. The [data stream](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/index/data_stream.go) resource might be a good example to look at. 

If you do look at this, feel free to ping me with any questions you have :)

jonjitsu:
A quick update, I started on this yesterday, so it's partially done (my golang is a bit rusty).

For PRs how big do we make them? For the Enrich API, in order to have complete coverage, there would be one resource and two data sources (one for looking up enrich policies and one for the stats endpoint). Should all this be in a single PR or split out into three, I lean towards smaller units so three PRs.

> If you do look at this, feel free to ping me with any questions you have :)

Where to ping you? I'm in the elastic slack workspace.

tobio:
> Where to ping you?

Here is fine.

> For PRs how big do we make them?

+1 to splitting this across 2/3 PRs. Definitely worth keeping resources and data sources independent, there's been cases where it's made sense to do two data sources in a single PR. It doesn't sound like that's the case here, but not a problem if you end up combining them.

Kushmaro:
Closing, this has been implemented in v0.6

PR #286, opened 2023-03-13
Enrich Policy Resource

jonjitsu:
This PR starts adding in support for the enrich API #282 . It adds a new resource `elasticstack_elasticsearch_enrich_policy`.


jonjitsu:
I've signed the contributor agreement 3 times. Once today and twice a couple weeks ago when I created the 2 character PR to fix the Makefile. I'm not sure what the next steps are to make the CLA bot happy.

tobio:
> I'm not sure what the next steps are to make the CLA bot happy.

I can see the signature against the @jonjitsu handle. It looks like the email address associated with the commits in this PR isn't publicised in you're GH handle and so it's not linking up with the CLA sig. 

You can either rewrite the commits with the existing email associated with @jonjitsu, or include your Gmail address in your GH profile so it gets linked correctly.

docs/resources/elasticsearch_enrich_policy.md
```diff
@@ -0,0 +1,89 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "elasticstack_elasticsearch_enrich_policy Resource - terraform-provider-elasticstack"
+subcategory: ""
```
tobio: There's a shortcoming with the tf docs generator that doesn't allow setting the subcategory from code. 

Can we add a template ([example](https://github.com/elastic/terraform-provider-elasticstack/blob/main/templates/resources/elasticsearch_index.md.tmpl#L2))for this resource setting this field.

internal/elasticsearch/enrich/policy.go
```diff
@@ -0,0 +1,184 @@
...
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+		},
+		"query": {
+			Description:  "Query used to filter documents in the enrich index. The policy only uses documents matching this query to enrich incoming documents. Defaults to a match_all query.",
+			Type:         schema.TypeString,
+			Optional:     true,
+			ForceNew:     true,
+			ValidateFunc: validation.StringIsJSON,
```
tobio:
```suggestion
			ValidateFunc: validation.StringIsJSON,
			DiffSuppressFunc: utils.DiffJsonSuppress,
```

Should we be suppressing a diff of equivalent JSON objects here?

internal/elasticsearch/enrich/policy.go
```diff
@@ -0,0 +1,184 @@
...
+	policy, diags := elasticsearch.GetEnrichPolicy(ctx, client, compName.ResourceId)
+	if policy == nil {
+		tflog.Warn(ctx, fmt.Sprintf(`Enrich policy "%s" not found, removing from state`, compName.ResourceId))
+		d.SetId("")
+		return diags
+	}
+	if diags == nil {
+		tflog.Warn(ctx, fmt.Sprintf(`Diags nil for resource ID "%s"`, compName.ResourceId))
+		return diags
+	}
```
tobio:
```suggestion
```

Why do we care about the distinction between nil/empty diags here?

tobio (commented):
Looking great, thanks for taking the time to put this change together. Couple of small comments inline.

internal/elasticsearch/enrich/policy.go
jonjitsu: `elasticsearch.GetEnrichPolicy` can return nil when http.StatusNotFound which would make the next line `if diags.HasError()` throw some kind of null dereference error. I could make `elasticsearch.GetEnrichPolicy` return a `diag.FromErr(fmt.Error)` instead for that situation.

From looking at many other resources ex: [index.GetIndex()](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/clients/elasticsearch/index.go#L317-L319) and it's use [resource](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/index/index.go#L916-L917) it combines the `diags != nil` test with the `model != nil` test. I could follow that pattern instead.

tobio: > I could follow that pattern instead.

👍

PR #291, opened 2023-03-21
COB-3434: Updated queue.max_bytes with unsupported number and units

yaseen-nbs:
COB-3434: Updated queue.max_bytes with unsupported number and units

### DIFF
--- a/internal/elasticsearch/logstash/pipeline.go
+++ b/internal/elasticsearch/logstash/pipeline.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"regexp"
 	"time"
 
 	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
@@ -136,7 +137,7 @@ func ResourceLogstashPipeline() *schema.Resource {
 		"queue_max_bytes": {
 			Description:  "Units for the total capacity of the queue when persistent queues are enabled.",
 			Type:         schema.TypeString,
-			ValidateFunc: validation.StringInSlice([]string{"b", "kb", "mb", "gb", "tb", "pb"}, false),
+			ValidateFunc: validation.StringMatch(regexp.MustCompile("^[0-9]+[kmgtp]?b$"), "must be valid size unit"),
 			Optional:     true,
 		},
 		"queue_max_events": {
--- a/internal/elasticsearch/logstash/pipeline.go
+++ b/internal/elasticsearch/logstash/pipeline.go
@@ -29,8 +29,7 @@ var (
 		"queue.checkpoint.retry":       schema.TypeBool,
 		"queue.checkpoint.writes":      schema.TypeInt,
 		"queue.drain":                  schema.TypeBool,
-		"queue.max_bytes.number":       schema.TypeInt,
-		"queue.max_bytes.units":        schema.TypeString,
+		"queue.max_bytes":        schema.TypeString,
 		"queue.max_events":             schema.TypeInt,
 		"queue.page_capacity":          schema.TypeString,
 		"queue.type":                   schema.TypeString,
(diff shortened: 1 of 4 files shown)


tobio (commented):
Thanks for taking the time to improve the provider! I'm happy to merge this one once the CLA is signed. 

Can you add an entry to the [CHANGELOG](https://github.com/elastic/terraform-provider-elasticstack/blob/main/CHANGELOG.md) for this one please?

It also looks like lint is failing since the docs haven't been updated with this schema change. You should be able to just run `make docs-generate` and commit the result to clear that up.

internal/elasticsearch/logstash/pipeline.go
```diff
@@ -134,15 +133,10 @@ func ResourceLogstashPipeline() *schema.Resource {
...
-			Description: "The total capacity of the queue when persistent queues are enabled.",
-			Type:        schema.TypeInt,
-			Optional:    true,
-		},
-		"queue_max_bytes_units": {
+		"queue_max_bytes": {
 			Description:  "Units for the total capacity of the queue when persistent queues are enabled.",
 			Type:         schema.TypeString,
-			ValidateFunc: validation.StringInSlice([]string{"b", "kb", "mb", "gb", "tb", "pb"}, false),
+			ValidateFunc: validation.StringInSlice([]string{"2b", "2kb", "2mb", "2gb", "2tb", "2pb"}, false),
```
RobsonSutton: @yaseen-nbs - Forgive me if I'm wrong but will this not just limit potential values to having a value of 2? e.g. will this not error if you set the test to 5gb instead or another value other than 2

yaseen-nbs: that's how it is defined the parameters for string option? I can change it back if that confusing. thanks.

RobsonSutton: So previously the number and units were split, so it's easier to just validate the number is an int and that the units match one of the strings defined in the validation slice (b, kb etc.). As the number field and unit values are now both being combined into a string, you'd need to find a way to validate that the string matches some pattern ([0-9]b, [0-9]kb or something similar), but I don't think validation.StringInSlice will allow for you to regex match against the string and literally matches so you may need to find a different method / approach for applying this input validation 👍

internal/elasticsearch/logstash/pipeline.go
```diff
@@ -134,12 +133,7 @@ func ResourceLogstashPipeline() *schema.Resource {
...
-		"queue_max_bytes_number": {
-			Description: "The total capacity of the queue when persistent queues are enabled.",
-			Type:        schema.TypeInt,
-			Optional:    true,
-		},
-		"queue_max_bytes_units": {
+		"queue_max_bytes": {
 			Description:  "Units for the total capacity of the queue when persistent queues are enabled.",
 			Type:         schema.TypeString,
 			ValidateFunc: validation.StringInSlice([]string{"b", "kb", "mb", "gb", "tb", "pb"}, false),
```
tobio:
```suggestion
			ValidateFunc: validation.Any(
				validation.StringIsEmpty,
				validation.StringMatch(regexp.MustCompile("^[0-9]+[kmgtp]?b$"), "must be valid size unit"),
			),
```

Might need to add the `regexp` import but I think this should do the right validation.

RobsonSutton: @tobio - Would you have permissions to add this by any chance? I don't think I have permissions to update this PR and Yaseen was a part of our team but left last week unfortunately so I can imagine this would just get left otherwise! Happy to get another PR raised to fix this if you'd prefer though

tobio: 👍 I'll try and take a look at it this week

tobio (approved):

Issue #48, opened 2022-01-28
[Feature] Support for Alerts


RobDTech:
Having the ability to add and maintain alerts

RobDTech:
https://www.elastic.co/guide/en/kibana/master/alerting-apis.html
Create/Update/Delete

acrewdson:
+1, would be really helpful

nicholastucker:
+2, would be great!

nbs-rsutton:
+3, would be very helpful!

alexlafreniere:
+4, this would be amazing for us

nicholas-r-king:
+5 we really need this.

cjr125:
+6 please support

cerdman:
+1 please add support

chanlawrencet:
+1 support for this would be a great help!

kryspetrie:
+1 PLEASE. Our support team really needs this!

rodya-mirov:
+1, This would be a big help for us, would appreciate any updates

hdeweirdt:
+1 great help for us

timmatsam:
+1 would greatly appreciate support for this

aleedom:
+1 could really use this feature for our workflow.

ModeEngage:
Would be great to use tools we already utilize to create our alerts!

jordan-donovan:
+1 this would be great for our team

mandragorn:
+1 this would really help reduce the load on our engineering support team.

desaip01:
+1 would greatly appreciate support for this

eczernikowski:
+1 please support really need this!

RobsonSutton:
@Kushmaro / @olksdr - I started taking a look at building this out the other day due to the level of demand from different people, but I couldn't seem to find an elastic supported go client for the Kibana API... I can see the elasticsearch one that is currently being used (https://github.com/elastic/go-elasticsearch). Please can you advise on the plan for delivering any of the Kibana based functionality via terraform? Is the plan to create a new repo for this client as is in place for elasticsearch, or are you looking to bake this into this repo somehow? If a new repo is being built for this, is this something we are able to contribute to in order to help speed up the delivery of the kibana functionality in this one?

Also, I'm assuming any Kibana based functionality will be blocked until this client can be built?

P.S. Still new to the Go / TF provider building world so please correct me if my understanding on any of the above is incorrect!

aydasraf:
[...]

Issue #111, opened 2022-05-26
[Bug] provider cannot auth w/ PWs containing irregular chars


neckenth:
**Describe the bug**
When creating a new Elasticsearch user via ElasticCloud UI, passwords with irregular characters are accepted. When using a password with irregular characters as the password for the `elastic/elasticstack` provider, cluster authentication fails. Perhaps the set of irregular characters are those that require URL encoding. ([MDN docs here](https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding)). `&` was the culprit for me.

The solution is to change the password to only include regular characters. When you do so in the ElasticCloud web app, however, this creates a need to `terraform destroy` the elasticstack user and associated resources or `terraform rm` those resources from the Terraform state. Otherwise, the bad password _continues_ to get used to try to connect to your ElasticSearch cluster by the `elastic/elasticstack` provider before the Terraform process arrives at using the the good password to authenticate the user.

For the sake of reproducing the issue, the password used was `eM&oM9Mh$FHN$#bo` (rotated and no longer in use, obviously)

**To Reproduce**
1. [config] In the ElasticCloud UI, create a new User (or update an existing User) and set the password to one that includes an `&`.
2. [config] In your `provider.tf` file, set the `elasticsearch` argument in the `elasticstack` provider object to use these new credentials:
```
provider "elasticstack" {
  elasticsearch {
    username  = <username>
    password  = <pw_w_irregular_chars>
    endpoints = [...]
  }
}
```
3. [operations] `terraform apply`
4. [error]
```
Failed with: {"error":{"root_cause":[{"type":"security_exception","reason":"unable to authenticate user [<username>] for REST request
│ [/<index_name>?flat_settings=true]","header":{"WWW-Authenticate":["Basic realm=\"security\" charset=\"UTF-8\"","Bearer
│ realm=\"security\"","ApiKey"]}}],"type":"security_exception","reason":"unable to authenticate user [<username>] for REST request
│ [/<index_name>?flat_settings=true]","header":{"WWW-Authenticate":["Basic realm=\"security\" charset=\"UTF-8\"","Bearer
│ realm=\"security\"","ApiKey"]}},"status":401}
```
**Expected behavior**
It is expected that the same password requirements would be used by both the ElasticCloud web app and the `elastic/elasticstack` provider. Any characters accepted in passwords by ElasticCloud should be supported by the `elastic/elasticstack` provider.

**Debug output**
[...]

Issue #71, opened 2022-02-15
[Feature]Need API key to authenticate from terraform to Elasticstack


sidharthvijayakumar:
Hey,

Need API key as an authentication mechanism instead of username and password. 

https://registry.terraform.io/providers/elastic/elasticstack/latest/docs

Currently, we are using username and password to authenticate with Elasticstack but as state files will have these details is there any secured way to authenticate with elasticstack/Kibana ?

Kushmaro:
@sidharthvijayakumar , you can actually authenticate using environment variables, if you don't want them to be stored in the state file.

ivan-penchev:
https://github.com/elastic/terraform-provider-elasticstack/pull/130 solves this.

PR #292, opened 2023-03-22
Add alert rule management

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/48

Adds support for a Kibana alert rule resource. This PR includes a generated Kibana alerting API client which we utilise to manage alert rules.


Makefile
```diff
@@ -227,3 +230,18 @@ release-notes: ## greps UNRELEASED notes from the CHANGELOG
 .PHONY: help
 help: ## this help
 	@ awk 'BEGIN {FS = ":.*##"; printf "Usage: make \033[36m<target>\033[0m\n\nTargets:\n"} /^[a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-10s\033[0m\t%s\n", $$1, $$2 }' $(MAKEFILE_LIST) | column -s$$'\t' -t
+
+.PHONY: generate-alerting-client
+generate-alerting-client:
```
dimuon: This should add the target with the description to the output of `help` target
```suggestion
generate-alerting-client: ## generate Kibana alerting client
```

internal/models/kibana.go
```diff
@@ -0,0 +1,34 @@
+package models
```
dimuon: nit: Shall we rename the file to something like `alerting_rule.go`

internal/clients/api_client.go
```diff
@@ -70,10 +73,12 @@ func NewApiClientFunc(version string) func(context.Context, *schema.ResourceData
 }
 
 func NewAcceptanceTestingClient() (*ApiClient, error) {
+	ua := buildUserAgent("tf-acceptance-testing")
```
dimuon, line 76: nit: ideally `NewAcceptanceTestingClient` should be put into a test package.

internal/kibana/alerting.go
```diff
@@ -0,0 +1,319 @@
+package kibana
```
dimuon, line 1: nit: while that's totally fine now, I suppose we will add more resources into `kibana` package. If so, does it make sense to introduce `alertingrule` package specifically for this resource?
Anyway I can refactor this once I start working on alert connector api. So please feel free to ignore the nit.

internal/kibana/alerting.go
```diff
@@ -0,0 +1,319 @@
...
+		e := v.(bool)
+		rule.Enabled = &e
+	}
+
+	if v, ok := d.GetOk("throttle"); ok {
+		t := v.(string)
+		rule.Throttle = &t
+	}
+
+	if v, ok := d.GetOk("params"); ok {
```
dimuon: Why do we use `GetOk`? `params` is required so I guess we have to use `Get`.

internal/clients/kibana/alerting.go
```diff
@@ -0,0 +1,181 @@
...
+func GetAlertingRule(ctx context.Context, apiClient *clients.ApiClient, id, spaceID string) (*models.AlertingRule, diag.Diagnostics) {
+	client, err := apiClient.GetAlertingClient()
+	if err != nil {
+		return nil, diag.FromErr(err)
+	}
+
+	ctxWithAuth := apiClient.SetAlertingAuthContext(ctx)
+	req := client.GetRule(ctxWithAuth, id, spaceID)
+	ruleRes, res, err := req.Execute()
+	if err != nil && res == nil {
```
dimuon, line 175: Why do we check that `res == nil` here? Does it make sense to check that `err != nil` only?

internal/clients/kibana/alerting.go
```diff
@@ -0,0 +1,181 @@
...
+func DeleteAlertingRule(ctx context.Context, apiClient *clients.ApiClient, ruleId string, spaceId string) diag.Diagnostics {
+	client, err := apiClient.GetAlertingClient()
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	ctxWithAuth := apiClient.SetAlertingAuthContext(ctx)
+	req := client.DeleteRule(ctxWithAuth, ruleId, spaceId).KbnXsrf("true")
+	res, err := req.Execute()
+	if err != nil && res == nil {
```
dimuon, line 195: The same question as for line #155

internal/clients/kibana/alerting.go
```diff
@@ -0,0 +1,181 @@
...
+		Params:     rule.Params,
+		Schedule: alerting.Schedule{
+			Interval: &rule.Schedule.Interval,
+		},
+		Tags:     rule.Tags,
+		Throttle: *alerting.NewNullableString(rule.Throttle),
+	}
+	req := client.UpdateRule(ctxWithAuth, rule.ID, rule.SpaceID).KbnXsrf("true").UpdateRuleRequest(reqModel)
+	ruleRes, res, err := req.Execute()
+	if err != nil && res == nil {
```
dimuon, line 131: Why do we check that res == nil here? Does it make sense to check that err != nil only?

internal/clients/kibana/alerting.go
```diff
@@ -0,0 +1,181 @@
...
+			RuleTypeId: rule.RuleTypeID,
+			Schedule: alerting.Schedule{
+				Interval: &rule.Schedule.Interval,
+			},
+			Tags:     rule.Tags,
+			Throttle: *alerting.NewNullableString(rule.Throttle),
+		}
+		req := client.CreateRule(ctxWithAuth, rule.SpaceID, "").KbnXsrf("true").CreateRuleRequest(reqModel)
+		ruleRes, res, err := req.Execute()
+		if err != nil && res == nil {
```
dimuon: The same question as for line #111

internal/clients/kibana/alerting.go
```diff
@@ -0,0 +1,181 @@
...
+			return nil, diag.FromErr(err)
+		}
+
+		defer res.Body.Close()
+		return ruleResponseToModel(rule.SpaceID, ruleRes), utils.CheckHttpError(res, "Unabled to create alerting rule")
```
dimuon: It looks like we don't use the result model besides its id. The same goes for Update. Does it makes sense to return only id and error?

internal/clients/kibana/alerting.go
```diff
@@ -0,0 +1,181 @@
...
+	actions := []alerting.ActionsInner{}
+	for _, action := range rule.Actions {
+		actions = append(actions, alerting.ActionsInner{
+			Group:  &action.Group,
+			Id:     &action.ID,
+			Params: action.Params,
+		})
+	}
+
+	if isCreate {
```
dimuon: Does it make sense to split the function into Create and Update ones? It looks like the only common part for these operations is the lines #57 - #71 that can be moved into a helper func. IMO upsert functions tend to be more convoluted and error prone than a pair of create and update.

internal/clients/kibana/alerting.go
```diff
@@ -0,0 +1,181 @@
...
+	"net/http"
+
+	"github.com/elastic/terraform-provider-elasticstack/generated/alerting"
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
+	"github.com/elastic/terraform-provider-elasticstack/internal/models"
+	"github.com/elastic/terraform-provider-elasticstack/internal/utils"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+)
+
+func ruleResponseToModel(spaceID string, res *alerting.RuleResponseProperties) *models.AlertingRule {
```
dimuon, line 23: Does it make sense to add unit tests to cover converting from TF data to the internal model, from the model to the request data and the opposite flow? The only tests we have are the acceptance tests.

dimuon (approved):

internal/clients/kibana/alerting.go
tobio: Yea, this is a bit of a gotcha IMO. Previous HTTP clients wouldn't return an error as a result of a HTTP level error (e.g 4xx response codes). This generated client however reverses that, returning both the failed HTTP response, along with a go error. 

The go error generated by the client here only includes the status line, so we end up with pretty unhelpful logs (e.g `400 Bad Request`). Instead, when there's a http response (`res != nil`) we're parsing the response body in order to provide a more helpful error message to the end user.

internal/clients/kibana/alerting.go
tobio: https://github.com/elastic/terraform-provider-elasticstack/pull/292/files#r1145424005

internal/clients/kibana/alerting.go
tobio: https://github.com/elastic/terraform-provider-elasticstack/pull/292/files#r1145424005

internal/clients/kibana/alerting.go
tobio: https://github.com/elastic/terraform-provider-elasticstack/pull/292/files#r1145424005

internal/kibana/alerting.go
tobio: I'm happy either way here, looking through the Kibana API I didn't see many cases where we would group resources and so it seemed like we'd end up with a bunch of packages containing only a single resource which felt a bit pointless IMO.

internal/clients/kibana/alerting.go
tobio: Yea, I thought about doing this. I don't have a strong justification either way, I decided to return the model since it was consistent with the API behaviour, rather than just returning what the current consumer wanted 🤷 happy to change it if you feel strongly.

internal/clients/api_client.go
tobio: Yea, you're right. Probably worth doing in another PR though.

dimuon (approved):
👍

Issue #278, opened 2023-02-20
[Bug] Logstash pipeline pipeline_metadata.version read is failing 


jassi-singh-wig:
**Describe the bug**
resource "elasticstack_elasticsearch_logstash_pipeline" fails with the below error, when the pipeline is updated from Kibana UI or an existing pipeline is imported
```
Error: pipeline_metadata.version: '' expected type 'string', got unconvertible type 'float64', value: '1'
```

**To Reproduce**
1. Create a logstash pipeline using terraform resource elasticstack_elasticsearch_logstash_pipeline
2. Open the pipeline in Kibana, click on Create & deploy
3. Re-run Step1 terraform
4. or, try importing an existing pipeline in terraform state

**Expected behavior**
No errors should be produced, existing logstash pipeline should be overwritten/imported by the terraform 

**Debug output**
Error: pipeline_metadata.version: '' expected type 'string', got unconvertible type 'float64', value: '1'
```
terraform import elasticstack_elasticsearch_logstash_pipeline.pipeline <cluster_uuid>/test-id
elasticstack_elasticsearch_logstash_pipeline.pipeline: Importing from ID "<cluster_uuid>/test-id"...
elasticstack_elasticsearch_logstash_pipeline.pipeline: Import prepared!
  Prepared elasticstack_elasticsearch_logstash_pipeline for import
elasticstack_elasticsearch_logstash_pipeline.pipeline: Refreshing state... [id=<cluster_uuid>/test-id]
╷
│ Error: pipeline_metadata.version: '' expected type 'string', got unconvertible type 'float64', value: '1'
│ 
│ 
╵
```

**Screenshots**

**Versions (please complete the following information):**
 - OS: macOS
 - Terraform Version v1.3.7
 - Provider version v0.5.0
 - Elasticsearch Version v8.6.2, v.7.17.8

**Additional context**

aydasraf:
+1 on the issue. Any timeline for resolution? Will 0.6 with altering features come anytime soon?

RobsonSutton:
@aydasraf - This should be fixed in the next version bump but will introduce a breaking change (just add jsonencode() around any existing config and it will work though). It seems 0.6.0 should be released at some point this month, once Alert Connector support has been added 👍

PR #294, opened 2023-03-24
[ISSUE-278] Resolve `pipeline_metadata.version` error

RobsonSutton:
- Raised to resolve - #278 
- **[Breaking Change]** Change type of `pipeline_metadata` to schema.TypeString - This setting throws an error when the pipeline is updated manually, as the subsequent terraform plan cannot handle the version value being of type: float64.
- Update docs to reflect #291 
- Reorder Logstash pipeline tests to match the order they are being checked in


RobsonSutton:
@tobio - Hey, another small one for when you have a spare min or two 👍 

Unfortunately I couldn't find a way to handle multiple data types in TypeMap... (please do enlighten me if this is possible though!)

Have also just deprecated the field as opposed to introducing a breaking change by replacing, so technically the error still exists in the old `pipeline_settings`, but this should provide a workaround for those that may run into it until the older field is fully removed

tobio:
> I'm inclined to just make this a breaking change.

@RobsonSutton, lets do this. It's nicer to break this now and completely close of this bug.

RobsonSutton:
@tobio - Should be updated to be a breaking change now 👍

tobio (approved):
LGTM  🎉 thanks for tidying this code up :)

PR #309, opened 2023-04-19
Generate SLO API client

wandergeek:
This PR adds another target similar to `generate-alerting-client` which generates the client for the SLO API. A subsequent PR will add this client to the provider itself. 

Rel: O11Y-190


generated/slo/test/api_slos_test.go
```diff
@@ -0,0 +1,139 @@
+/*
+SLOs
+
+Testing SlosApiService
+
+*/
+
+// Code generated by OpenAPI Generator (https://openapi-generator.tech);
+
+package slo
```
tobio: Shouldn't this file be deleted from the make target?

tobio (commented):
Pulling this branch and running `make generate-slo-client` results in a dirty git state. That doesn't feel like it's expected?

tobio (approved):
LGTM

PR #311, opened 2023-04-27
Add Fleet API Client

taylor-swanson:
- Add API client for Fleet
- Add code generator for Fleet API types, uses OpenAPI spec
- Add related docs and tests
- Data sources and resources will come in later PRs


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/elastic/terraform-provider-elasticstack
 go 1.19
 
 require (
+	github.com/deepmap/oapi-codegen v1.12.4
```
tobio: I think this should be in `tools/go.mod`

internal/clients/api_client.go
```diff
@@ -503,6 +542,92 @@ func buildAlertingClient(baseConfig BaseConfig, config kibana.Config) *alerting.
...
+				config.CACerts = append(config.CACerts, vStr)
+			}
+		}
+	}
+	if v, ok := fleetData["insecure"].(bool); ok {
+		config.Insecure = v
+	}
+
+	// Set variables from environment variables.
+	if v := os.Getenv("FLEET_ENDPOINT"); v != "" {
```
tobio: Any reason to not also use the `KIBANA_*` env variables here too?

internal/clients/fleet/fleetapi/generate.go
tobio, line 1: This file terrifies me, why do we need to be editing the openapi spec?

internal/clients/fleet/fleetapi/generate.go
```diff
@@ -0,0 +1,321 @@
...
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"os"
+	"strings"
+)
+
+const (
+	fleetOpenapiURL = "https://raw.githubusercontent.com/elastic/kibana/main/x-pack/plugins/fleet/common/openapi/bundled.json"
```
tobio: I think we should pin this to a specific version rather than just track main.

tobio:
This is awesome! Thanks for the work here. Are we able to split this one up a little to simplify the review process. It feels like it should be relatively easy to have a PR per resource/data source, and maybe an initial one adding the fleet client?

taylor-swanson:
Okay, removed all of the data source/resources from this PR, so it will just be about the fleet client. I'll add those back in later PRs.

Also, noticed buildkite is running but it doesn't seem to have a pipeline file, so it failed.

internal/clients/fleet/debug.go
tobio, line 1: It seems like there's a lot of duplication between the ES debug logger and this file. Perhaps we should move them both to a debug package and share some of the code here instead?

tobio (commented):
Couple of comments inline, generally lgtm. Modifying the openapi spec during generation feels pretty dangerous, is there a reason we need that here, shouldn't we be able to publish a working spec?

internal/clients/fleet/fleetapi/generate.go
taylor-swanson: The fleet API is still very much in the works (hasn't even been GA-ed yet, but that should be coming in 8.8 as far as I know). For now, I think we should keep it at main, then pin it once it is GA.

internal/clients/fleet/fleetapi/generate.go
taylor-swanson: This was originally written by @andrewkroh and I continued to use it as it seemed to work fine.

Andrew, did we edit it to work around some issues with the API, and is it even necessary anymore?

internal/clients/api_client.go
taylor-swanson: I was told by the Fleet API team that Fleet may not always be under Kibana, so I separated it out here. I could change some of these back to KIBANA_*, but there are some values that won't carry over, since there isn't an option for it in the config (API_KEY for example). Schema would probably have to change too, so it would get a bit messy.

internal/clients/fleet/fleetapi/generate.go
taylor-swanson: Also, I'm looking at using `openapitools/openapi-generator-cli` to see if that works well with the Fleet API. If it works, I'll use that instead, as I'd rather be consistent with everything else.

internal/clients/fleet/debug.go
taylor-swanson: Good point, didn't even notice the ES debug logger. I work on cleaning that up.

internal/clients/fleet/fleetapi/generate.go
taylor-swanson: Okay, so OpenAPI Generator cannot generate valid code with the Fleet spec:

```shell
$ make build-ci             
go build -o terraform-provider-elasticstack
# github.com/elastic/terraform-provider-elasticstack/generated/fleet
generated/fleet/model_package_policy.go:39:6: NewPackagePolicy redeclared in this block
        generated/fleet/model_new_package_policy.go:21:6: other declaration of NewPackagePolicy
make: *** [build-ci] Error 1
```

Also, it added some serious bloat to this change:
```
411 files changed, 86208 insertions(+), 541 deletions(-)
```

If we could limit what models are generated, that may help. We are waiting on some changes to be made in the API for package policies, so we don't want them being generated now anyway.

internal/clients/fleet/debug.go
taylor-swanson: After another look, the only real overlap between these two loggers is the print json lines functions, which I moved to the utils package. The ES debug logger implements an interface from the ES package and the fleet logger is just an extension of http.RoundTripper.

internal/clients/api_client.go
tobio: > so it would get a bit messy.

So re-reading the code here a little. All we'd need to do to default to re-using the Kibana config values would be to accept a `kibana.Config` instead of the `BaseConfig` parameter to this function. The rest of this can remain as is, optionally overriding the Kibana config with Fleet specifics. 

> I was told by the Fleet API team that Fleet may not always be under Kibana, so I separated it out here.

Are there actual plans to separate those APIs? IMO we should be simplifying the UX based on the current architecture. If such a split occurs then we can adjust the provider. 

Having to configure essentially the same thing twice feels pretty nasty, and it feels pretty straightforward to avoid that.

internal/clients/fleet/fleetapi/generate.go
tobio: @andrewkroh are you able to help me understand the need for this file?

> Andrew, did we edit it to work around some issues with the API, and is it even necessary anymore?

Maybe we can try generating the client without this file and see what the result is?

internal/clients/api_client.go
taylor-swanson: That's fair, I'll go ahead and do that then.

internal/clients/fleet/fleetapi/generate.go
andrewkroh: I wrote this when I was hacking a PoC. In it's current state I never intended it for production. It does a few things. Some of which should be incorporated into the upstream schema.

- It filters the generated code to only include the parts that are used. This removes a bunch of unused and deprecated models.
- It fixes some issues with the generated code to avoid anonymous nested structs. This makes the generated code easier to use.
- It adds some parameters that are missing from the openapi spec, but are used by the real API as observed by inspecting Kibana's requests to the API.

taylor-swanson: I cleaned this code up, so hopefully less "hacks". It's driven mostly by structs, so a bit safer to deal with now instead of blind maps. That said, I left in the filtering of paths and methods, since that is actually very useful to us and cuts down on the code bloat. I also kept in the filtering of the kbn-xsrf parameters. If I understand it correctly, we don't need this in generated code as we set the header in our Fleet client (correct me if I'm wrong, @andrewkroh).

taylor-swanson: @tobio, any other comments on this?

I think given the current state of the Fleet OpenAPI spec, the filtering that we are currently doing (especially with my latest round of changes), is fair. To recap, all I am doing now is:
- Filtering out unneeded/deprecated API endpoints and methods. As Andrew mentioned, this dramatically cuts down on the amount of code generated.
- Removed the kbn-xsrf parameter from generated code. As mentioned, we already set the header in the API client. I have not noticed any issues when removing this code.

Since this is all still considered experimental code, I'd rather get this in and start iterating on it sooner than later. We plan to use this extensively on the Security External Integrations team for our demo environment to continuously test our integrations.

tobio: Yep, they're all good points. 👍 on getting this in and making progress.

tobio (approved):
🎉 thanks for all the effort here, looking forward to getting the resources added too.

internal/clients/fleet/fleetapi/doc.go
```diff
@@ -0,0 +1,5 @@
+package fleetapi
+
+//go:generate go run generate.go -o fleet-filtered.json
```
andrewkroh, line 3: Do you think we should use `-v v8.7.1` or some release tag instead of generating from `main` (from default)?

andrewkroh (commented):
Got a chance to look it over today. Nice work.

internal/clients/fleet/fleetapi/doc.go
taylor-swanson: Yeah I think we should once the API goes GA (or whenever we get the package policy changes?). I could quickly test it right now and if there's not any significant differences, we could lock to `v8.7.1` or something equivalent

PR #322, opened 2023-05-09
Add Fleet agent policy and enrollment tokens

taylor-swanson:
- Add Fleet agent policy resource type
- Add Fleet enrollment token data source type
- Add new types to Fleet generator
- Add missing requestBody field in Fleet OpenAPI struct


internal/fleet/agent_policy_resource_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients/fleet"
+	"github.com/elastic/terraform-provider-elasticstack/internal/versionutils"
+	"github.com/hashicorp/go-version"
+	sdkacctest "github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+var minVersionAgentPolicy = version.Must(version.NewVersion("8.6.0"))
```
tobio, line 18: Can we bump the [STACK_VERSION](https://github.com/elastic/terraform-provider-elasticstack/blob/main/Makefile#L18) to 8.6.0 so `make docker-testacc` Just Works™️

internal/fleet/agent_policy_resource.go
```diff
@@ -0,0 +1,284 @@
...
+	monitorMetrics = "metrics"
+)
+
+func ResourceAgentPolicy() *schema.Resource {
+	agentPolicySchema := map[string]*schema.Schema{
+		"id": {
+			Description: "Internal identifier of the resource.",
+			Type:        schema.TypeString,
+			Computed:    true,
+		},
```
tobio:
```suggestion
```

We don't need these ID attributes, and we're essentially duplicating it with the `policy_id` one.

internal/fleet/agent_policy_resource.go
```diff
@@ -0,0 +1,284 @@
...
+		"policy_id": {
+			Description: "Unique identifier of the agent policy.",
+			Type:        schema.TypeString,
+			Computed:    true,
```
tobio, line 22:
```suggestion
			Computed:    true,
			Optional:       true, 
			ForceNew:     true,
```

Should users be able to set this as well? Reading the create code it feels like they should. Also, if they can then changing the ID should force a new resource.

internal/fleet/enrollment_tokens_data_source.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSourceEnrollmentTokens() *schema.Resource {
+	enrollmentTokenSchema := map[string]*schema.Schema{
+		"id": {
+			Description: "Internal identifier of the data source.",
+			Type:        schema.TypeString,
+			Computed:    true,
+		},
```
tobio: Similarly, we probably don't need to expose this in the schema.

internal/fleet/enrollment_tokens_data_source_test.go
```diff
@@ -0,0 +1,39 @@
...
+func TestAccDataSourceEnrollmentTokens(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acctest.PreCheck(t) },
+		ProtoV5ProviderFactories: acctest.Providers,
+		Steps: []resource.TestStep{
+			{
+				SkipFunc: versionutils.CheckIfVersionIsUnsupported(minVersionEnrollmentTokens),
+				Config:   testAccDataSourceEnrollmentToken,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.elasticstack_fleet_enrollment_tokens.test", "policy_id", "223b1bf8-240f-463f-8466-5062670d0754"),
```
tobio, line 24: If there going to be a resource to create enrolment tokens? If not then maybe we should check that the returned list is empty? This test just feels a bit like "yay it doesn't blow up" right now.

internal/fleet/enrollment_tokens_data_source.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"context"
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients/fleet"
+	"github.com/elastic/terraform-provider-elasticstack/internal/utils"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSourceEnrollmentTokens() *schema.Resource {
```
tobio, line 12: TIL Enrollment is US English, aka incorrect.

tobio (commented):
This is looking great, couple of small comments.

internal/fleet/enrollment_tokens_data_source.go
tobio: Just to be clear, I'm not suggesting we change it. The ES resource is an enrollment token, I've just never read the word that closely before to notice it's poorly spelled.

tobio: Also, I'm joking. It's just that, well English is in the name so maybe the version used in England is arguably more correct. 

Maybe we should just discuss it over a nice biscuit whilst relaxing in our thongs.

taylor-swanson: 🇺🇸 🦅 🦅 🦅 🇺🇸 

Jokes aside, I had no idea that there was a difference. The documentation uses the US English variant, so I ended up going with that spelling.

As an aside, while I usually prefer the US variants for most things, I really wish we could adopt the metric system.

internal/fleet/enrollment_tokens_data_source_test.go
taylor-swanson: As far as I know, not at this time. An enrollment token is automatically generated for an agent policy, so the most typical case right now is retrieving the token associated with that policy.

> If not then maybe we should check that the returned list is empty?

That would only work in situations where there are no existing tokens. I usually test on a stack brought up by `elastic-package`, which has a default agent policy and token, so this check would always fail. I'm not sure what the right thing to do here would be. We could either ignore it like I'm currently doing (yuck, I know), or develop this into a resource and check it that way (which is beyond my current scope of work).

internal/fleet/enrollment_tokens_data_source.go
tobio: > The documentation uses the US English variant, so I ended up going with that spelling.

:100: Internal consistency >>>> My personal opinion on the matter.

internal/fleet/enrollment_tokens_data_source_test.go
tobio: > An enrollment token is automatically generated for an agent policy, so the most typical case right now is retrieving the token associated with that policy.

I see you're doing that in the `agent_policy_resource_test` file. I think we should do the same here, and assert that the data source retrieves a matching enrolment token. WDYT? Alternatively we could create a policy, and then get all the enrolment tokens and verify that at least one for the managed policy exists. 

The magic GUID feels weird here, perhaps there's some context I'm missing there and that GUID is meaningful in some way?

taylor-swanson: > Alternatively we could create a policy, and then get all the enrolment tokens and verify that at least one for the managed policy exists.

I think this is the better option, at least we are testing something real and will work in every circumstance. It's no longer testing in isolation (we are involving agent policy now), but I'm fine with that given the circumstances. I'll go ahead and add that.

> The magic GUID feels weird here, perhaps there's some context I'm missing there and that GUID is meaningful in some way?

Nope, it was just a random UUID and has no special meaning.

tobio (approved):
🎉

PR #326, opened 2023-05-15
Guarantee indices order in SLM

### LINKED ISSUE #325: [Bug] indices order is not guaranteed in SLM
**Describe the bug**
Order of `indices` field in `elasticstack_elasticsearch_snapshot_lifecycle` resource is not guaranteed as it uses [set](https://developer.hashicorp.com/terraform/language/expressions/type-constraints#set).
It causes unexpected behavior when we want to exclude some indices.
For example, when we want to include all indices except hidden indices, the configuration would be like the below, but since the order is not guaranteed, it'll include all indices by applying it in unexpected order.
```
indices = ["*", "-.*"]
=>  applied as `-.*, *`
```

**To Reproduce**
1. TF configuration used
```
resource "elasticstack_elasticsearch_snapshot_lifecycle" "example" {
  name     = "example"

  schedule      = "0 */30 * * * ?"
  snapshot_name = "<example-snapshot-{now/d}>"
  repository    = "example-repository"

  indices              = ["*", "-.*"] # Exclude system indices which have '.' prefix.
  ignore_unavailable   = false
  include_global_state = true

  expire_after = "1d"
  min_count    = 10
  max_count    = 100
}
```
3. TF operations to execute to get the error '...' [e.g `terraform plan`,`terraform apply`, `terraform destroy`]
4. See the error in the output '...'

**Expected behavior**
Values of the indices field are ordered same as the input.

**Debug output**
Run `terraform` command with `TF_LOG=trace` and provide extended information on TF operations.

**Screenshots**

 - OS: Linux
 - Terraform Version 1.2.4
 - Provider version 0.5.0
 - Elasticsearch Version 8.6.0

**Additional context**

k-yomo:
Resolves #325 by changing type of `indices` field to `List`.

### DIFF
--- a/internal/elasticsearch/cluster/slm_test.go
+++ b/internal/elasticsearch/cluster/slm_test.go
@@ -31,7 +31,8 @@ func TestAccResourceSLM(t *testing.T) {
 					resource.TestCheckResourceAttr("elasticstack_elasticsearch_snapshot_lifecycle.test_slm", "max_count", "50"),
 					resource.TestCheckResourceAttr("elasticstack_elasticsearch_snapshot_lifecycle.test_slm", "ignore_unavailable", "false"),
 					resource.TestCheckResourceAttr("elasticstack_elasticsearch_snapshot_lifecycle.test_slm", "include_global_state", "false"),
-					resource.TestCheckTypeSetElemAttr("elasticstack_elasticsearch_snapshot_lifecycle.test_slm", "indices.*", "data-*"),
+					resource.TestCheckResourceAttr("elasticstack_elasticsearch_snapshot_lifecycle.test_slm", "indices.0", "data-*"),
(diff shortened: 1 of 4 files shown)


internal/elasticsearch/cluster/slm_test.go
```diff
@@ -68,7 +69,7 @@ resource "elasticstack_elasticsearch_snapshot_lifecycle" "test_slm" {
   snapshot_name = "<daily-snap-{now/d}>"
   repository    = elasticstack_elasticsearch_snapshot_repository.repo.name
 
-  indices              = ["data-*", "important"]
+  indices              = ["data-*", "abc"]
```
k-yomo, line 72: String `set` will be converted to list in alphabetical order, so replace it to make sure the result is not alphabetically re-ordered.
https://developer.hashicorp.com/terraform/language/expressions/type-constraints#conversion-of-complex-types

tobio (approved):
Awesome!

Issue #33, opened 2022-01-10
[feature] support Elasticsearch indexes


paulojmdias:
Hi all 👋 

First of all, congratulations for building this project 🙌 

Do we have any plans to support managing Elasticsearch Indexes?

Thanks

olksdr:
hey @paulojmdias, 
I'm pretty sure were are going to support indices as well. And @Kushmaro can provide more info on the roadmap here.

Kushmaro:
Thanks for opening this @paulojmdias !
This is definitely something that we intend to support, as of now we're waiting for some feedback from our early users (such as yourself) to determine priority and importance, and this definitely resonates with things that we intend to support.

Do you mind if I ask for your general use case around our terraform provider in conjunction with an Index resource?
will you be looking to enforce specific index settings over indices using the provider? or is it for something more along the lines of CI/CD purposes?

paulojmdias:
Hello @Kushmaro @olksdr,

Many thanks for at least hearing my request! :)

I want to manage Elasticsearch indices across multiple environments and clusters using Terraform and GitOps with [Atlantis](https://www.runatlantis.io/)

Of course, I want to abstract some settings in the index resource using a Terraform module, but this is for removing the dependency so everyone doesn't need to know what value could fill in a specific setting and avoiding copy-paste errors.

Regarding the details of what type of operations I want to cover with a Terraform resource:
* Creating Elasticsearch indices and defining all the possible options settings like number of shards, replicas, refresh interval, routing partition size, alias, and also mappings
* Ability to add a field to mapping without destroying the index
   * If someone trying to make a change of type in a current field mapping the behavior should re-create the index
* Ability to move alias between indices
   * Example: moving the alias dummy from index dummy_v1 to dummy_v2 without destroying the index
* Have a flag `force_destroy` that could manage  the ability to protect the index destroy if the value is `false`

We already use a Terraform provider, but most of these operations aren't allowed without destroying the index and Elasticsearch supports them without destroying and this has a huge impact on the automation process.

Feel free to ask more questions, I am able to help to grow this project. 🙌

gtheodorio:
I also would love the ability to create ES indices. Currently getting around this by creating an index template, life cycle config, and adding an alias, but would be nice to just create indices and aliases.

JoaoSerras:
+1 for this feature 🙌 
It would be lovely to have this available

paulojmdias:
Many thanks, @olksdr 🙏 

[...]

PR #343, opened 2023-05-31
Add plugin framework 

wandergeek:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/214

Adds full support for the plugin framework. Adds a resource to import saved objects in Kibana as a POC for using the framework in this provider.


internal/schema/connection.go
```diff
@@ -145,7 +357,7 @@ func GetKibanaConnectionSchema() *schema.Schema {
 					RequiredWith: []string{"kibana.0.username"},
 				},
 				"endpoints": {
-					Description: "A list of endpoints where the terraform provider will point to, this must include the http(s) schema and port number.",
+					Description: "A comma-separated list of endpoints where the terraform provider will point to, this must include the http(s) schema and port number.",
```
tobio, line 344:
```suggestion
					Description: "A list of endpoints where the terraform provider will point to, this must include the http(s) schema and port number.",
```

It's not comma separated though, it's a TF list.

go.mod
```diff
@@ -8,6 +8,8 @@ require (
 	github.com/elastic/go-elasticsearch/v7 v7.17.10
 	github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320
 	github.com/hashicorp/go-version v1.6.0
+	github.com/hashicorp/terraform-plugin-framework v1.2.0
```
dimuon: Does it make sense to update it to the latest? (1.4.0)

go.mod
```diff
@@ -8,6 +8,8 @@ require (
...
+	github.com/hashicorp/terraform-plugin-framework-validators v0.10.0
```
dimuon: The same for validators - a newer version is available.

docs/resources/kibana_import_saved_objects.md
```diff
@@ -0,0 +1,99 @@
...
+## Example Usage
+
+```terraform
+provider "elasticstack" {
+  kibana {}
+}
+
+resource "elasticstack_kibana_import_saved_objects" "settings" {
+  overwrite     = true
+  file_contents = <<-EOT
```
dimuon, line 22: Does it make sense to use `jsonencode` instead of heredoc - https://developer.hashicorp.com/terraform/language/expressions/strings#generating-json-or-yaml?

dimuon: Actually, maybe it makes sense to use heredoc which may be a more convenient tool for inserting the file generated by the API.

provider/provider.go
```diff
@@ -17,6 +17,8 @@ import (
 )
 
 const esKeyName = "elasticsearch"
+const kbKeyName = "kibana"
```
dimuon, line 20: I suppose these constants should also be used in lines #33-34.

provider/factory_test.go
```diff
@@ -0,0 +1,51 @@
...
+		elasticsearch {
+		  username  = "sup"
+		  password  = "dawg"
+		  endpoints = ["http://localhost:9200"]
+		}
+	  }
+	`
+	resource.Test(t, resource.TestCase{
+		ProtoV5ProviderFactories: map[string]func() (tfprotov5.ProviderServer, error){
+			"elasticstack": func() (tfprotov5.ProviderServer, error) {
```
dimuon, line 24: The func is almost identical to `ProtoV5ProviderServerFactory`. Shall we call (and test) it instead?

internal/kibana/import_saved_objects/schema.go
```diff
@@ -0,0 +1,157 @@
...
+// func (r *Resource) ConfigValidators(context.Context) []resource.ConfigValidator {
+// 	return []resource.ConfigValidator{
+// 		resourcevalidator.Conflicting(
+// 			path.MatchRoot("create_new_copies"),
+// 			path.MatchRoot("overwrite"),
+// 			path.MatchRoot("compatibility_mode"),
+// 		),
+// 	}
+// }
+
```
dimuon, line 35: Can we remove it?

internal/kibana/import_saved_objects/acc_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccResourceImportSavedObjects(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acctest.PreCheck(t) },
+		ProtoV5ProviderFactories: acctest.Providers,
+		Steps: []resource.TestStep{
```
dimuon, line 14: It makes sense to add a step to check Update.

internal/schema/connection.go
```diff
@@ -3,9 +3,207 @@ package schema
...
+	usernamePath := path.MatchRelative().AtParent().AtName("username")
+	passwordPath := path.MatchRelative().AtParent().AtName("password")
+	caFilePath := path.MatchRelative().AtParent().AtName("ca_file")
+	caDataPath := path.MatchRelative().AtParent().AtName("ca_data")
+	certFilePath := path.MatchRelative().AtParent().AtName("cert_file")
+	certDataPath := path.MatchRelative().AtParent().AtName("cert_data")
+	keyFilePath := path.MatchRelative().AtParent().AtName("key_file")
+	keyDataPath := path.MatchRelative().AtParent().AtName("key_data")
+
+	return fwschema.ListNestedBlock{
```
dimuon, line 25: Why not use `SingleNestedBlock`?

internal/clients/config/base.go
```diff
@@ -0,0 +1,97 @@
...
+		baseConfig.ApiKey = esConfig.APIKey.ValueString()
+	}
+
+	return baseConfig.withEnvironmentOverrides()
+}
+
+func (b baseConfig) withEnvironmentOverrides() baseConfig {
+	b.Username = withEnvironmentOverride(b.Username, "ELASTICSEARCH_USERNAME")
+	b.Password = withEnvironmentOverride(b.Password, "ELASTICSEARCH_PASSWORD")
+	b.ApiKey = withEnvironmentOverride(b.ApiKey, "ELASTICSEARCH_API_KEY")
```
dimuon, line 66: Shall we add `ELASTICSEARCH_INSECURE` as well?

dimuon: Please ignore this - I found it in the ES config.

internal/schema/connection.go
dimuon: The same question for Kibana and Fleet.

internal/clients/config/base_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"net/http"
+	"os"
+	"testing"
+
+	providerSchema "github.com/elastic/terraform-provider-elasticstack/internal/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewBaseConfigFromSDK(t *testing.T) {
```
dimuon, line 14: Does it make sense to test also `newBaseConfigFromFramework`?

internal/clients/config/client.go
```diff
@@ -0,0 +1,14 @@
+package config
+
+import (
+	"github.com/disaster37/go-kibana-rest/v8"
+	"github.com/elastic/go-elasticsearch/v7"
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients/fleet"
+)
+
+type Client struct {
```
dimuon, line 9: It contains configurations for clients. Shall we rename it to `ClientConfig`?

internal/clients/config/kibana.go
```diff
@@ -0,0 +1,101 @@
...
+func newKibanaConfigFromSDK(d *schema.ResourceData, base baseConfig) (kibanaConfig, sdkdiags.Diagnostics) {
+	var diags sdkdiags.Diagnostics
+
+	kibConn, ok := d.GetOk("kibana")
+	if !ok {
+		return kibanaConfig{}, diags
+	}
+
+	// Use ES details by default
+	config := base.toKibanaConfig()
```
dimuon: It looks combersome. The logic allows creating Kibana resources using creds from ES config and endpoints definition from an env variable but do we really need this?
It sounds like a legacy but maybe we can refactor that - IMO it complicates the logic without real benefits. Kibana config should be taken from `kibana` atrribute and the corresponding env vars. 
The similar approach is used for reading Fleet config.

internal/clients/config/base.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type baseConfig struct {
```
dimuon, line 11: Does it make to rename it to `EsBaseConfig`?

internal/clients/config/base.go
```diff
@@ -0,0 +1,97 @@
...
+	Username  string
+	Password  string
+	ApiKey    string
+	UserAgent string
+	Header    http.Header
+}
+
+func newBaseConfigFromSDK(d *schema.ResourceData, version string, esKey string) baseConfig {
```
dimuon, line 19: Shall we rename it to something like `newESbaseConfigFromSDK`? The func is called only for `elasticsearch` attribute.

internal/clients/config/elasticsearch.go
```diff
@@ -0,0 +1,218 @@
+package config
```
dimuon, line 1: The package lacks unit tests.

dimuon:
@wandergeek , @tobio I reviewed the PR. Thanks for the initiative 👍

internal/kibana/import_saved_objects/schema.go
tobio: Added a comment here. The kibana client doesn't support `create_new_copies` and `compatibility_mode` which we could add when https://github.com/elastic/terraform-provider-elasticstack/pull/372 is merged.

internal/schema/connection.go
tobio: There's strict compatibility between the SDK and FW schemas which forces the use of ListNestedBlock here.

internal/clients/config/client.go
tobio: It's within a `config` package, so usage would be `config.Client` `config.ClientConfig` is pretty stuttery IMO.

internal/clients/config/kibana.go
tobio: > IMO it complicates the logic without real benefits.

Whether there's benefit would depend heavily on the usage, if people are using the provider to configure both Kibana and ES resources then sharing credentials simplifies their configuration. FWIW I have only ever configured the ES credentials and relied on Kibana using those credentials.

internal/clients/config/base.go
tobio: There's nothing ES specific within this struct though.

internal/clients/config/base.go
tobio: Nope, as above it's not ES specific. It'd be inconsistent with everything else in this package IMO.

tobio:
@dimuon I think I've covered all the points in the PR, are you able to take another look here?

dimuon (dismissed):
LGTM 👍 

it needs changelog :)

tobio (approved):

PR #344, opened 2023-05-31
HTTP logging for Kibana action connectors

dimuon:
Enable HTTP logging for Kibana action connectors client.


tobio (approved):
Would be good to include a changelog entry here too :)

tobio (approved):

PR #346, opened 2023-06-02
Update README.md

kunisen:
Add support statement per an earlier internal discussion.


tobio (changes requested):
From the EC PR:

So we don't publicly announce such a policy on other client repositories. I'm curious about why we need to do so here? 

 > however, since terraform is a 3rd party tool (by Hashicorp) which Elastic has no full control over
 
I don't think it's helpful to base the policy on this. The same argument could be made for Elasticsearch, the operating system is a 3rd party tool we have no control over. There's a clearly defined API between Terraform and the provider, we test against that API, in theory we should stand behind that testing. This argument just sounds like we're terrible software engineers. 

> Also, since urgent production-related terraform issues can be resolved via regular API or UI interaction, we ask customers to resort to such methods in case of urgent deployment downtime or issue.

If we're to have this caveat on the repo, we should base it on this. The initial aim for sev1's is to restore normality by any means available, since we can restore normality by bypassing Terraform, a provider bug can't be classed as a sev1. 

Related to that, let's assume there was a provider bug which resulted in deployments being terminated. A customer should quite rightly be able to open a sev1 about such a bug. There is no public facing API to restore a terminated deployment, and we would absolutely work on fixing the provider bug immediately. 

It's not clear why we need this caveat in the provider repos, but not in other clients? I think it's potentially misleading i.e there's situations where a sev1 is likely the right thing to do. 

Finally, the logic presented in this PR is IMO damaging to the reputation of our engineering team. If we need to include this warning, we should do so based on the availability of alternate management methods, not because we don't trust our product to adhere to an API. 

Specifically for this one: 
> in case of urgent deployment downtime or issue.

This provider manages stack resources, not deployments. We should probably remove the 'deployment' wording here, or come up with something less Cloud specific.

kunisen:
@Kushmaro 
Sorry to trouble you directly. Toby and I had an internal discussion and he suggested asking for your follow up on how we want to organize the public statement. 
The comment for reference - https://github.com/elastic/terraform-provider-elasticstack/pull/346#pullrequestreview-1461812596

Please let us know if anything is not clear or if you prefer to discuss this internally and thanks!

Also thanks @tobio for the detailed comment and internal discussion! 🙏

Kushmaro (dismissed):
lgtm

Kushmaro:
looks like a charged conversation :) 
Let's take this internally, please.

tobio (approved):
LGTM

PR #357, opened 2023-06-09
Add skip_destory flag for Fleet Agent Policies

taylor-swanson:
Add a flag which can skip the attempt to remove the Agent Policy from Fleet using the API. This is useful in cases where an Agent is not managed directly by Terraform and it is not safe to remove the policy, as there may still be agents assigned to the policy.

### DIFF
--- a/internal/fleet/agent_policy_resource.go
+++ b/internal/fleet/agent_policy_resource.go
@@ -2,6 +2,7 @@ package fleet
 
 import (
 	"context"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
 
 	"github.com/elastic/terraform-provider-elasticstack/internal/clients/fleet"
 	"github.com/elastic/terraform-provider-elasticstack/internal/clients/fleet/fleetapi"
@@ -70,6 +71,11 @@ func ResourceAgentPolicy() *schema.Resource {
 		},
 		"monitor_metrics": {
 			Description: "Enable collection of agent metrics.",
+			Type:        schema.TypeBool,
+			Optional:    true,
+		},
+		"skip_destroy": {
+			Description: "Set to true if you do not wish the agent policy to be deleted at destroy time, and instead just remove the agent policy from the Terraform state.",
 			Type:        schema.TypeBool,
 			Optional:    true,
 		},
@@ -264,6 +270,11 @@ func resourceAgentPolicyRead(ctx context.Context, d *schema.ResourceData, meta i
 }
 
 func resourceAgentPolicyDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	if d.Get("skip_destroy").(bool) {
+		tflog.Debug(ctx, "Skipping destroy of Agent Policy", map[string]interface{}{"policy_id": d.Id()})
+		return nil
+	}
+
 	fleetClient, diags := getFleetClient(d, meta)
 	if diags.HasError() {
 		return diags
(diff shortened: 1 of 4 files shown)


andrewkroh (commented):
Could you please add an acceptance test for this. I think it would be good to have an assertion that the state associated with the resource is gone. (Not knowing the TF SDK well, I wasn't sure if a `d.SetID("")` is required, and a test will confirm this.)

taylor-swanson:
> Could you please add an acceptance test for this. I think it would be good to have an assertion that the state associated with the resource is gone. (Not knowing the TF SDK well, I wasn't sure if a `d.SetID("")` is required, and a test will confirm this.)

I did test it manually and it worked as expected. Only reason why I was opposed to adding an acceptance test for it was I didn't care for the idea of leaving around resources after a test. As long as we are okay with that, it's trivial to add the acceptance test and assert the resource is still there.

Any thoughts on this, @tobio?

> I wasn't sure if a d.SetID("") is required

I'm a little fuzzy on this as well. I *think* it's redundant here since we are already actioning on deleting the resource from state. What we're bypassing is the call out to fleet to delete.

andrewkroh:
> As long as we are okay with that, it's trivial to add the acceptance test and assert the resource is still there.

I would rather have it cleaned up. I think the test could do this though. Use one of the hooks, verify through the API that the resource does exist, and then delete it.

andrewkroh (approved):

internal/fleet/agent_policy_resource_test.go
```diff
@@ -27,31 +27,57 @@ func TestAccResourceAgentPolicy(t *testing.T) {
...
 					resource.TestCheckResourceAttr("elasticstack_fleet_agent_policy.test_policy", "monitor_metrics", "true"),
+					resource.TestCheckResourceAttr("elasticstack_fleet_agent_policy.test_policy", "skip_destroy", "false"),
 				),
 			},
 		},
 	})
 }
 
-func testAccResourceAgentPolicyCreate(id string) string {
+func TestAccResourceAgentPolicySkipDestroy(t *testing.T) {
```
tobio, line 56: ❤️

internal/fleet/agent_policy_resource.go
```diff
@@ -264,6 +270,11 @@ func resourceAgentPolicyRead(ctx context.Context, d *schema.ResourceData, meta i
 }
 
 func resourceAgentPolicyDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	if d.Get("skip_destroy").(bool) {
+		tflog.Debug(ctx, "Skipping destroy of Agent Policy", map[string]interface{}{"policy_id": d.Id()})
```
tobio, line 274: IMO this should be at a higher log level so it's clear to users that we're not destroying this policy during apply.

tobio (approved):
LGTM. Not totally on board with the debug log level, I'd prefer it to be info, wdyt?

taylor-swanson:
> LGTM. Not totally on board with the debug log level, I'd prefer it to be info, wdyt?

I made it debug level based on what I saw in the AWS provider (here's an example: https://github.com/hashicorp/terraform-provider-aws/blob/70260ef76efac73597e49841908e23e574d478e0/internal/service/logs/group.go#L211)

Since it's an opt-in option, it'll never be enabled by default, so I'm okay with it being at debug level for this reason. Also, logs won't appear unless a user opts in with TF_LOG=level.

Once this PR is merged, could we get a release cut for us (a 0.6.2, for instance)? This is currently blocking progress on us adopting the provider.

tobio:
> could we get a release cut for us (a 0.6.2, for instance)?

I'll look at cutting a release this week.

tobio:
@taylor-swanson @ebeahan 0.6.2 got pushed out today.

Issue #360, opened 2023-06-16
[Bug] Terraform does not import SLM policy name when importing the resource


glisignoli:
Importing a `elasticstack_elasticsearch_snapshot_lifecycle` resource does not import the name field of the existing policy, which then forces a replacement of the existing policy as the name field has `ForceNew: true`. 

Original issue
-----

**Describe the bug**
When importing an existing elasticstack_elasticsearch_snapshot_repository resource, Terraform does not import the name field of the existing repository, which then forces a replacement of the current resource since the name field is required for that resource.

This was previously reported here: https://github.com/elastic/terraform-provider-elasticstack/issues/116 and closed, but the issue seems to be re-occurring.

**To Reproduce**
1. `terraform import elasticstack_elasticsearch_snapshot_lifecycle.terraform-daily-snapshots xLG5mdQuXTijLXiypWBXew/terraform-daily-snapshots`
2. `terraform state show elasticstack_elasticsearch_snapshot_lifecycle.terraform-daily-snapshots`

**Expected behavior**
The name field should be imported

**Versions (please complete the following information):**
 - OS: Linux
 - Terraform Version 1.5.0
 - Provider version 0.6.1
 - Elasticsearch Version 8.8.0

**Additional context**
```
$ terraform import elasticstack_elasticsearch_snapshot_lifecycle.terraform-daily-snapshots xLG5mdQuXTijLXiypWBXew/terraform-daily-snapshots
elasticstack_elasticsearch_snapshot_lifecycle.terraform-daily-snapshots: Importing from ID "xLG5mdQuXTijLXiypWBXew/terraform-daily-snapshots"...
elasticstack_elasticsearch_snapshot_lifecycle.terraform-daily-snapshots: Import prepared!
  Prepared elasticstack_elasticsearch_snapshot_lifecycle for import
elasticstack_elasticsearch_snapshot_lifecycle.terraform-daily-snapshots: Refreshing state... [id=xLG5mdQuXTijLXiypWBXew/terraform-daily-snapshots]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.

$ terraform state show elasticstack_elasticsearch_snapshot_lifecycle.terraform-daily-snapshots
# elasticstack_elasticsearch_snapshot_lifecycle.terraform-daily-snapshots:
resource "elasticstack_elasticsearch_snapshot_lifecycle" "terraform-daily-snapshots" {
[...]

PR #363, opened 2023-06-20
Set the policy name when importing an SLM policy

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/360

Does what it says on the tin. Importing an SLM policy currently doesn't set the name property, resulting in the policy being recreated on the next apply.


dimuon (approved):
LGTM

PR #367, opened 2023-06-27
[DOCS] Rules & Connectors/Actions Management with Terraform

lcawl:
Relates to https://github.com/elastic/kibana/issues/158647

This PR drafts a new guide to explain how to create a Kibana rule and connector. The example also includes the creation of component templates, an index template, ilm policy, and data stream as the source for the rule and an index for the target of the connector.


tobio (approved):
Thanks for working on improving the docs here. Great to see contributions from the Stack side 💪.

@jcger can you take a look at the Kibana ccontent here.

examples/resources/elasticstack_kibana_alerting_rule/resource-index-rule.tf
```diff
@@ -0,0 +1,35 @@
+provider "elasticstack" {
```
jcger: Not sure if we should add the provider block to each file as it would throw an error if doing so. What about a common/global reference to it (maybe in prerequisites)?

Maybe the intention is just to let the user know that they'll have to add it somewhere, but then we should add the kibana property here too as without it the user won't be able to use a Kibana API like the rules API

examples/resources/elasticstack_kibana_alerting_rule/resource-index-rule.tf
```diff
@@ -0,0 +1,35 @@
...
+  elasticsearch {}
+}
+
+resource "elasticstack_kibana_alerting_rule" "my_rule" {
```
jcger: What about using more realistic names for easier understanding/following? If I'm understanding the props right, this one could be something like `DailyDocumentCountThresholdExceeded`

jcger (approved):
nice!

examples/resources/elasticstack_kibana_alerting_rule/resource-index-rule.tf
lcawl: I've removed the repeated provider blurb via b10a90b5be59f3f263201fb398431282b83c348c and also added a comment in the prerequisites section saying that you need to already have your provider set up for your cluster or deployment.

examples/resources/elasticstack_kibana_alerting_rule/resource-index-rule.tf
lcawl: Good tip, I've made that change too in b10a90b5be59f3f263201fb398431282b83c348c

tobio (approved):

PR #385, opened 2023-07-14
Add SLO support to provider

wandergeek:
## Overview

This PR adds kibana [SLO](https://github.com/elastic/kibana/blob/main/x-pack/plugins/observability/dev_docs/slo.md) management capabilities to the provider. 

## Testing

The custom metrics and histogram indicators are only supported in >8.10, so you will need to launch latest and greatest stack. I've been using kibana to do this:

### Launch ES
```
	cd ~/work/kibana
	yarn kbn bootstrap
	yarn es snapshot --license trial --use-cached
```

### Launch KB
```
	cd ~/work/kibana
	yarn kbn bootstrap
	yarn start --no-base-path
```

### Run tests

```
export ELASTICSEARCH_ENDPOINTS=http://localhost:9200 \
KIBANA_ENDPOINT=http://localhost:5601 \
ELASTICSEARCH_USERNAME=elastic \
ELASTICSEARCH_PASSWORD=changeme

make lint test testacc docs-generate
```


tobio:
Other version specific resources [define](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/security/api_key_test.go#L30) a `SkipFunc` to avoid the failing tests.

internal/clients/api_client.go
```diff
@@ -586,6 +597,20 @@ func buildConnectorsClient(baseConfig BaseConfig, config kibana.Config) (*connec
 	)
 }
 
+func buildSloClient(baseConfig BaseConfig, config kibana.Config) *slo.APIClient {
+	//again this is the same -- worth pulling this out into a common place?
```
tobio: What's the same here?

internal/clients/api_client.go
```diff
@@ -246,8 +257,8 @@ func (a *ApiClient) GetFleetClient() (*fleet.Client, error) {
 	return a.fleet, nil
 }
 
-func (a *ApiClient) SetAlertingAuthContext(ctx context.Context) context.Context {
-	return context.WithValue(ctx, alerting.ContextBasicAuth, alerting.BasicAuth{
+func (a *ApiClient) SetGeneratedClientAuthContext(ctx context.Context) context.Context {
+	return context.WithValue(ctx, slo.ContextBasicAuth, slo.BasicAuth{
```
tobio, line 261: Does this still work for alerting?

internal/clients/kibana/slo.go
```diff
@@ -0,0 +1,171 @@
...
+	s.ID = sloRes.Id
+
+	return &s, diag.Diagnostics{}
+}
+
+func responseIndicatorToCreateSloRequestIndicator(s slo.SloResponseIndicator) (slo.CreateSloRequestIndicator, error) {
+	var ret slo.CreateSloRequestIndicator
+
+	ind := s.GetActualInstance()
+	switch ind.(type) { //nolint:all
```
tobio: Why do we need to disable linting here?

internal/clients/kibana/slo_test.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"testing"
+
+	"github.com/elastic/terraform-provider-elasticstack/generated/slo"
+	"github.com/elastic/terraform-provider-elasticstack/internal/models"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_sloResponseToModel(t *testing.T) {
+	// now := time.Now()
```
tobio:
```suggestion
```

internal/clients/kibana/slo.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return ret, nil
+}
+
+func sloResponseToModel(spaceID string, res *slo.SloResponse) *models.Slo {
+	if res == nil {
+		return nil
+	}
+
+	return &models.Slo{
```
tobio, line 161: Won't this blow up if any of these fields are nil? What's ensuring that these exist?

internal/kibana/slo.go
```diff
@@ -0,0 +1,477 @@
...
+			Required:    true,
+		},
+		"description": {
+			Description: "A description for the SLO.",
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"indicator": {
+			Type:     schema.TypeList,
+			Required: true,
```
tobio: Probably want to have a min/max length on this one.

internal/kibana/slo.go
```diff
@@ -0,0 +1,477 @@
...
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"type": {
+						Type:         schema.TypeString,
+						Required:     true,
+						ValidateFunc: validation.StringInSlice([]string{"sli.kql.custom", "sli.apm.transactionErrorRate", "sli.apm.transactionDuration", "sli.histogram.custom"}, false),
+					},
+					"params": {
+						Type:     schema.TypeList,
```
tobio, line 62: Similarly, min/max length

internal/kibana/slo.go
```diff
@@ -0,0 +1,477 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+		"time_window": {
+			Description: "Currently support calendar aligned and rolling time windows. Any duration greater than 1 day can be used: days, weeks, months, quarters, years. Rolling time window requires a duration, e.g. 1w for one week, and isRolling: true. SLOs defined with such time window, will only consider the SLI data from the last duration period as a moving window. Calendar aligned time window requires a duration, limited to 1M for monthly or 1w for weekly, and isCalendar: true.",
+			Type:        schema.TypeList,
+			Required:    true,
```
tobio, line 336: Similarly

internal/kibana/slo.go
```diff
@@ -0,0 +1,477 @@
...
+		},
+		"budgeting_method": {
+			Description:  "An occurrences budgeting method uses the number of good and total events during the time window. A timeslices budgeting method uses the number of good slices and total slices during the time window. A slice is an arbitrary time window (smaller than the overall SLO time window) that is either considered good or bad, calculated from the timeslice threshold and the ratio of good over total events that happened during the slice window. A budgeting method is required and must be either occurrences or timeslices.",
+			Type:         schema.TypeString,
+			Required:     true,
+			ValidateFunc: validation.StringInSlice([]string{"occurrences", "timeslices"}, false),
+		},
+		"objective": {
+			Description: "The target objective is the value the SLO needs to meet during the time window. If a timeslices budgeting method is used, we also need to define the timesliceTarget which can be different than the overall SLO target.",
+			Type:        schema.TypeList,
```
tobio, line 360: Similarly

internal/kibana/slo.go
```diff
@@ -0,0 +1,477 @@
...
+					"timeslice_window": {
+						Type:     schema.TypeString,
+						Optional: true,
+					},
+				},
+			},
+		},
+		"settings": {
+			Description: "The default settings should be sufficient for most users, but if needed, these properties can be overwritten.",
+			Type:        schema.TypeList,
```
tobio, line 383: Similarly

internal/kibana/slo.go
```diff
@@ -0,0 +1,477 @@
...
+		return diags
+	}
+
+	id, diags := client.ID(ctx, res.ID)
+	if diags.HasError() {
+		return diags
+	}
+
+	id.ResourceId = res.ID
+	d.SetId(id.String())
```
tobio, line 623: This should be a composite ID with the space to support imports.

internal/kibana/slo.go
```diff
@@ -0,0 +1,477 @@
...
+			"params": []map[string]interface{}{{
+				"index":           params.Index,
+				"filter":          params.Filter,
+				"good":            params.Filter,
+				"total":           params.Total,
+				"timestamp_field": params.TimestampField,
+			}},
+		})
+	} else {
+		return diag.Errorf("unknown indicator type")
```
tobio: Can we include what the indicator type was in this error message

internal/kibana/slo.go
```diff
@@ -0,0 +1,477 @@
...
+	}
+	if err := d.Set("indicator", indicators); err != nil {
+		return diag.FromErr(err)
+	}
+
+	time_window := []interface{}{}
+	time_window = append(time_window, map[string]interface{}{
+		"duration": s.TimeWindow.Duration,
+		"type":     s.TimeWindow.Type,
+	})
```
tobio: This can just be a slice literal, no need for the append.

internal/kibana/slo.go
```diff
@@ -0,0 +1,477 @@
...
+	if err := d.Set("time_window", time_window); err != nil {
+		return diag.FromErr(err)
+	}
+
+	objective := []interface{}{}
+	objective = append(objective, map[string]interface{}{
```
tobio: Similarly.

internal/kibana/slo.go
```diff
@@ -0,0 +1,477 @@
...
+		"target":           s.Objective.Target,
+		"timeslice_target": s.Objective.TimesliceTarget,
+		"timeslice_window": s.Objective.TimesliceWindow,
+	})
+	if err := d.Set("objective", objective); err != nil {
+		return diag.FromErr(err)
+	}
+
+	settings := []interface{}{}
+	settings = append(settings, map[string]interface{}{
```
tobio: In the create/update settings are optional, but we're always setting them here. Should this be conditional on something?

internal/kibana/slo_test.go
```diff
@@ -0,0 +1,180 @@
...
+		ProtoV5ProviderFactories: acctest.Providers,
+		Steps: []resource.TestStep{
+			{
+				SkipFunc: versionutils.CheckIfVersionIsUnsupported(minSupportedVersion),
+				Config:   testAccResourceSloCreate(sloName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("elasticstack_kibana_slo.test_slo", "name", sloName),
+					resource.TestCheckResourceAttr("elasticstack_kibana_slo.test_slo", "description", "fully sick SLO"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_slo.test_slo", "indicator.0.type", "sli.apm.transactionDuration"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_slo.test_slo", "space_id", "default"),
```
tobio, line 49: Can we validate the other fields are set here.

internal/kibana/slo_test.go
```diff
@@ -0,0 +1,180 @@
...
+				),
+			},
+			{
+				SkipFunc: versionutils.CheckIfVersionIsUnsupported(minSupportedVersion),
+				Config:   testAccResourceSloUpdate(sloName),
```
tobio: It would be nice to check that we're correctly handling optional fields in these tests. Also would be good to validate the other slo types.

internal/clients/api_client.go
wandergeek: The underlying config struct is exactly the same:
```
type Configuration struct {
	Host             string            `json:"host,omitempty"`
	Scheme           string            `json:"scheme,omitempty"`
	DefaultHeader    map[string]string `json:"defaultHeader,omitempty"`
	UserAgent        string            `json:"userAgent,omitempty"`
	Debug            bool              `json:"debug,omitempty"`
	Servers          ServerConfigurations
	OperationServers map[string]ServerConfigurations
	HTTPClient       *http.Client
}
```

internal/clients/api_client.go
wandergeek: Hmmm, good question. I see why this function used to have cursing in it. The acceptance tests passed, but perhaps this isn't used there?

internal/clients/kibana/slo.go
wandergeek:
```
internal/clients/kibana/slo.go:127:9: S1034: assigning the result of this type assertion to a variable (switch ind := ind.(type)) could eliminate type assertions in switch cases (gosimple)
	switch ind.(type) {
	       ^
internal/clients/kibana/slo.go:129:11: S1034(related information): could eliminate this type assertion (gosimple)
		i, _ := ind.(*slo.IndicatorPropertiesApmAvailability)
		        ^
internal/clients/kibana/slo.go:133:11: S1034(related information): could eliminate this type assertion (gosimple)
		i, _ := ind.(*slo.IndicatorPropertiesApmLatency)
		        ^
internal/clients/kibana/slo.go:137:11: S1034(related information): could eliminate this type assertion (gosimple)
		i, _ := ind.(*slo.IndicatorPropertiesCustomKql)
		        ^
internal/clients/kibana/slo.go:141:11: S1034(related information): could eliminate this type assertion (gosimple)
		i, _ := ind.(*slo.IndicatorPropertiesCustomMetric)
		        ^
internal/clients/kibana/slo.go:145:11: S1034(related information): could eliminate this type assertion (gosimple)
		i, _ := ind.(*slo.IndicatorPropertiesHistogram)
```

internal/clients/api_client.go
wandergeek: Do we really need to recreate this for each generated client?

internal/clients/kibana/slo.go
wandergeek: image

From the model I suppose. Most of these fields except for settings are compulsory according to the spec. Settings is a pointer so if its nil, it should be ok right? That said, is it still worth ensuring these fields are there?

internal/kibana/slo.go
wandergeek: cool, done

internal/kibana/slo.go
wandergeek: Can you elaborate here? This is the same pattern I'm seeing everywhere in the codebase.

Should it instead be:
```go
	compId, diags := clients.CompositeIdFromStr(d.Id())
	if diags.HasError() {
		return diags
	}
	d.SetId(compId.String())
```

internal/kibana/slo.go
wandergeek: Na, since I'm just looking to see which indicator type is set. If it hits here, the case is actually that the indicator is somehow unset. Will update log message.

internal/kibana/slo.go
wandergeek: like this?
```go

	time_window := []interface{}{
		map[string]interface{}{
			"duration": s.TimeWindow.Duration,
			"type":     s.TimeWindow.Type,
		},
	}
```

internal/kibana/slo.go
wandergeek: Something like this maybe?
```go
	var settings *slo.Settings
	if _, ok := d.GetOk("settings"); ok {
		settings = &slo.Settings{
			SyncDelay: getOrNilString("settings.0.sync_delay", d),
			Frequency: getOrNilString("settings.0.frequency", d),
		}
	} else {
		settings = nil
.
.
.
Settings:        settings,
```

internal/kibana/slo_test.go
wandergeek: good idea

internal/kibana/slo.go
wandergeek: When I did this, I got
```
        Error: Wrong resource ID.

          with elasticstack_kibana_slo.test_slo,
          on terraform_plugin_test.tf line 12, in resource "elasticstack_kibana_slo" "test_slo":
          12: resource "elasticstack_kibana_slo" "test_slo" {

        Resource ID must have following format: <cluster_uuid>/<resource identifier>
```

wandergeek:
A note from testing to deal with later:

I restarted the stack, but still had some terraform state laying around. When I tried to do an apply, I got the following:

```
elasticstack_kibana_slo.test_slo: Refreshing state... [id=SozILMG9RtCl6hFyeubVow/7374c7b0-21e3-11ee-bd67-4fa609eac85f]
╷
│ Error: 404 Not Found
│
│   with elasticstack_kibana_slo.test_slo,
│   on provider.tf line 25, in resource "elasticstack_kibana_slo" "test_slo":
│   25: resource "elasticstack_kibana_slo" "test_slo" {
│
╵
```

A 404 is expected here since obviously that SLO is gone with the restart, but wondering if the provider should be smart enough to recreate it or just fail like this? I feel like at least the error message should be a bit more specific?

internal/clients/api_client.go
wandergeek: Nope, I broke it. 😂 
Annoyingly, this is exactly the same on both generated clients. Wondering how we can just have one version of this.

internal/clients/kibana/slo.go
tobio: But why not make the suggested change here? It's a legit linting issue which makes the code nicer to consume. 

```go
	ind := s.GetActualInstance()
	switch ind := ind.(type) {
	case *slo.IndicatorPropertiesApmAvailability:
		ret.IndicatorPropertiesApmAvailability = ind

	case *slo.IndicatorPropertiesApmLatency:
		ret.IndicatorPropertiesApmLatency = ind

	case *slo.IndicatorPropertiesCustomKql:
		ret.IndicatorPropertiesCustomKql = ind

	case *slo.IndicatorPropertiesCustomMetric:
		ret.IndicatorPropertiesCustomMetric = ind

	case *slo.IndicatorPropertiesHistogram:
		ret.IndicatorPropertiesHistogram = ind
```

internal/clients/api_client.go
tobio: Basically, we can't. We could do something like:

```
func SetAlertingAuthContext(ctx) ctx {
  return SetGeneratedClisnterAuthContext(ctx, alerting.ContextBasicAuth)
}

func SetGeneratedClisnterAuthContext(ctx, key any) ctx {
  return context.WithValue(ctx, key, ...)
}
```

But we can't reduce this to a single function. Golang context keys require strict type equivalence, `slo.contextKey("hello") != alerting.contextKey("hello")` since the underlying types are different even though they both derive from string. This is designed to avoid key collisions on common string values.

internal/clients/kibana/slo.go
tobio: Non of them are marked as required in the openapi spec, otherwise they wouldn't be pointers in the generated code. I'd probably update our internal model to accept the optionality or explicitly deal with it here. Leaving it like this will create provider crashes.

internal/kibana/slo.go
tobio: More like

```
if s.Settings != nil {
  if err := d.Set("settings", []interface{}{
    map[string]interface{}{
      "sync_delay": s.Settings.SyncDelay,
      "frequency": s.Settings.Frequency,
    }
  }); err != nil {
    return diag.FromErr(err)
  }
}
```

tobio:
> A 404 is expected here since obviously that SLO is gone with the restart, but wondering if the provider should be smart enough to recreate it or just fail like this? I feel like at least the error message should be a bit more specific?

The resource should be treating this error as if the resource no longer exists and needs to be recreated.

internal/kibana/slo.go
tobio: Probably because you had an existing resource, You'll have to delete any resources before updating this to something like:

```
	id := &clients.CompositeId{ClusterId: rule.SpaceID, ResourceId: res.ID}
	d.SetId(id.String())
```

internal/clients/kibana/slo.go
wandergeek: Ah that is much cleaner, thanks!

internal/clients/api_client.go
wandergeek: TIL, good to know, will add a version for each generated client then.

internal/clients/kibana/slo.go
wandergeek: Can you elaborate here? OpenAPI spec shows everything but settings and id are required. 
image

wandergeek: So this is interesting, should our internal model reflect [CreateSloRequest](https://github.com/wandergeek/terraform-provider-elasticstack/blob/O11Y-236-add-slo-pr-2/generated/slo/model_create_slo_request.go#L21-L33) or [SloResponse](https://github.com/wandergeek/terraform-provider-elasticstack/blob/O11Y-236-add-slo-pr-2/generated/slo/model_slo_response.go#L21-L42)? Both have very different optionality.

internal/kibana/slo.go
wandergeek: Ok, cool. Will give this a whirl.

internal/kibana/slo.go
wandergeek: Ok, sweet. Updated.

internal/clients/kibana/slo.go
tobio: The request, it's what a user needs to provide to create the resource.

wandergeek:
> > A 404 is expected here since obviously that SLO is gone with the restart, but wondering if the provider should be smart enough to recreate it or just fail like this? I feel like at least the error message should be a bit more specific?
> 
> The resource should be treating this error as if the resource no longer exists and needs to be recreated.

Sweet, this is fixed in https://github.com/elastic/terraform-provider-elasticstack/pull/385/commits/811c59f464256478beab1ffe3fa24d861e6c1c61. Was erroring out too quickly in the chain.

internal/clients/kibana/slo.go
wandergeek: This is fixed by making the response have required fields

wandergeek: https://github.com/elastic/kibana/pull/163430

dimuon:
> Using a static spec file until [elastic/kibana#161922](https://github.com/elastic/kibana/pull/161922) is merged.
> ##

@wandergeek , the PR is merged. Shall we switch to getting OpenAPI specs from Kibana repo instead of using the static file?

internal/kibana/slo.go
```diff
@@ -0,0 +1,506 @@
...
+			Type:     schema.TypeList,
+			Required: true,
+			MinItems: 1,
+			MaxItems: 1,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"type": {
+						Type:         schema.TypeString,
+						Required:     true,
+						ValidateFunc: validation.StringInSlice([]string{"sli.kql.custom", "sli.apm.transactionErrorRate", "sli.apm.transactionDuration", "sli.histogram.custom", "sli.metric.custom"}, false),
```
dimuon: Does it make sense to omit unsupported types `sli.histogram.custom` and `sli.metric.custom` instead of reporting an error later on lines #268-271?

internal/kibana/slo.go
```diff
@@ -0,0 +1,506 @@
...
+					},
+					"params": {
```
dimuon: `params` can be represented by, e.g. JSON string (similar to `params` in `elasticstack_kibana_alerting_rule` or `config` in `elasticstack_kibana_action_connector` - internal/kibana/connector.go#42). This should allow having different objects (JSON strings) for `params` attribure based on `type`'s value that makes it possible to support `sli.histogram.custom` and `sli.metric.custom` types.

dimuon: Another approach to support these 2 currently unsupported fields is to replace general `indicator` attribute with more specific indicator attributes (e.g. `histogram_custom_indicator` for `sli.histogram.custom`, `metric_custom_indicator` for `sli.metric.custom`, etc.) and add a validation that only one of them is present in configuration (but at least one should be specified).

dimuon: @Kushmaro, what do you think?

internal/clients/kibana/slo_test.go
```diff
@@ -0,0 +1,33 @@
...
+		sloResponse   *slo.SloResponse
+		expectedModel *models.Slo
+	}{
+		{
+			name:          "nil response should return a nil model",
+			spaceId:       "space-id",
+			sloResponse:   nil,
+			expectedModel: nil,
+		},
+	}
```
dimuon, line 87: Shall we add a test case with data in `sloResponse` and `expectedModel`?

dimuon:
@wandergeek , please update `CHANGELOG.md`.

internal/kibana/slo.go
```diff
@@ -0,0 +1,506 @@
...
+	if s == nil && diags == nil {
+		d.SetId("")
+		return nil
+	}
+	if diags.HasError() {
+		return diags
+	}
+
+	indicator := []interface{}{}
+	if s.Indicator.IndicatorPropertiesApmAvailability != nil {
```
dimuon: Does it make sense to replace `if` with `switch`?

internal/kibana/slo_test.go
```diff
@@ -0,0 +1,266 @@
...
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients/kibana"
+	"github.com/elastic/terraform-provider-elasticstack/internal/versionutils"
+	"github.com/hashicorp/go-version"
+	sdkacctest "github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+var minSupportedVersion = version.Must(version.NewSemver("8.9.0"))
```
dimuon: Does it make sense to move the definition to the test func to reduce its visibility?

internal/kibana/slo_test.go
```diff
@@ -0,0 +1,266 @@
...
+			{
+				SkipFunc:    versionutils.CheckIfVersionIsUnsupported(minSupportedVersion),
+				Config:      getTFConfig(sloName, "sli.metric.custom", true),
+				ExpectError: regexp.MustCompile("the sli.metric.custom indicator type is currently unsupported"),
+			},
+		},
+	})
+}
+
+func getTFConfig(name string, indicatorType string, settingsEnabled bool) string {
```
dimuon: The name is too short for such a specific function that is used only in one test. IMO it should be renamed to be more verbose or, ideally, moded to the test function scope.

dimuon:
@wandergeek , thanks for the contribution 👍 
There are some nitpicks and few thoughts about adding support for 2 currently unsupported indicator types.

wandergeek:
> > Using a static spec file until [elastic/kibana#161922](https://github.com/elastic/kibana/pull/161922) is merged.
> 
> @wandergeek , the PR is merged. Shall we switch to getting OpenAPI specs from Kibana repo instead of using the static file?

Yep, good idea. One thing I'm grappling with right now is whether to point the URL to `main` or a specific commit. Ideally I'd like to point it to a particular kibana version tag, but it looks like there's a number of commits that haven't been backported for some reason.  

Curious what you think here. I've pointed it to main for now.

internal/kibana/slo.go
wandergeek: The intention is to support them eventually. My thinking here is that I'd be able to provide the user with a more informative error message until we can support them. It also sets us up for when we're able to add the new types.

internal/kibana/slo.go
wandergeek: > params can be represented by, e.g. JSON string 

I thought about this, but don't we lose the validation at that point? This is a pretty complex piece of the API, and it'd be nice to validate that the params per indicator are correct. 

> Another approach to support these 2 currently unsupported fields is to replace general indicator attribute with more specific indicator attributes (e.g. histogram_custom_indicator for sli.histogram.custom, metric_custom_indicator for sli.metric.custom, etc.) and add a validation that only one of them is present in configuration (but at least one should be specified).

Can you elaborate here? I'm not totally following. Perhaps you can provide some pseudocode?

internal/clients/kibana/slo_test.go
wandergeek: Good idea, I had something there before, but didn't add it back for some reason.

internal/kibana/slo.go
wandergeek: I'm not sure how that's possible? I'm not really switching on a value here.

wandergeek: Sure, if you like. Seems to fit the style of the rest of the project better. Will just have a little `if` at the end to catch the case where an indicator isn't set for some reason.

internal/kibana/slo_test.go
wandergeek: My thinking was that it could be used for subsequent tests. Happy to move it into the test func in the meantime.

internal/kibana/slo_test.go
wandergeek: Yeah good point. Happy to move it in the test scope.

dimuon:
> > > Using a static spec file until [elastic/kibana#161922](https://github.com/elastic/kibana/pull/161922) is merged.
> > 
> > 
> > @wandergeek , the PR is merged. Shall we switch to getting OpenAPI specs from Kibana repo instead of using the static file?
> 
> Yep, good idea. One thing I'm grappling with right now is whether to point the URL to `main` or a specific commit. Ideally I'd like to point it to a particular kibana version tag, but it looks like there's a number of commits that haven't been backported for some reason.
> 
> Curious what you think here. I've pointed it to main for now.

I guess we can follow [the same approach as for alerting](https://github.com/elastic/terraform-provider-elasticstack/blob/main/Makefile#L242C55-L242C72).

internal/kibana/slo.go
dimuon: Yes, we will lose some convenience of TF with JSON string.

Regarding the second approach. The idea is to specify separate `indicator` field for each unique use case. E.g. we can define the following indicator attribute for `sli.metric.custom` in addition to the `indicator` attribute that you implemented:

```
               "indicator": {
               // definition omitted
               }
		"metric_custom_indicator": {
			Type:         schema.TypeList,
			Required:     true,
			MinItems:     1,
			MaxItems:     1,
			ExactlyOneOf: []string{"indicator", "metric_custom_indicator", "histogram_custom_indicator"},
			Elem: &schema.Resource{
				Schema: map[string]*schema.Schema{
					"params": {
						Type:     schema.TypeList,
						Required: true,
						MinItems: 1,
						MaxItems: 1,
						Elem: &schema.Resource{
							Schema: map[string]*schema.Schema{
								"index": {
									Type:     schema.TypeString,
									Required: true,
								},
								"filter": {
									Type:     schema.TypeString,
									Optional: true,
								},
								"good": {
									Type:     schema.TypeList,
									Required: true,
									MinItems: 1,
									MaxItems: 1,
									Elem: &schema.Resource{
										Schema: map[string]*schema.Schema{
											"metrics": {
												Type:     schema.TypeList,
												Required: true,
												Elem: &schema.Resource{
													Schema: map[string]*schema.Schema{
														"name": {
															Type:     schema.TypeString,
															Required: true,
														},
														"aggregation": {
															Type:     schema.TypeString,
															Required: true,
														},
														"field": {
															Type:     schema.TypeString,
															Required: true,
														},
														"filter": {
															Type:     schema.TypeString,
															Optional: true,
														},
													},
												},
											},
											"equation": {
												Type:     schema.TypeString,
												Required: true,
											},
										},
									},
								},
								"total": {
									Type:     schema.TypeList,
									Required: true,
									// Elem should be defined similar to "good"
								},
								"timestamp_field": {
									Type:     schema.TypeString,
									Required: true,
								},
							},
						},
					},
				},
			},
		},

               "histogram_custom_indicator": {
              // definition omitted
              }
```

The code implies that the schema defines 3 indicator attributes - `indicator` attribute defined as before covers `sli.kql.custom`, `sli.apm.transactionErrorRate` and `sli.apm.transactionDuration` types, `metric_custom_indicator` covers `sli.metric.custom` and `histogram_custom_indicator` attribute (I didn't provide a definition for it) covers `sli.histogram.custom`.

`ExactlyOneOf` ([help](https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/validators-predefined)) field in the schema definition (it should be added to all 3 indicator attributes) should make sure that TF configuration defines exactly only one of the 3 mentioned indicator attributes. However, I didn't use `ExactlyOneOf` before and generally didn't implement such approach previously so additional check is needed.

Also, if we follow the approach, the schema definition would be quite verbose, so we can extract some attribute definitions to helper functions, e.g. like [here](https://github.com/elastic/terraform-provider-ec/blob/master/ec/ecresource/deploymentresource/elasticsearch/v2/schema.go#L134).

Does it make sense?

wandergeek: Yep, makes perfect sense. Thanks for the extra context and example! I'll see if I can make this work.

wandergeek:
> I guess we can follow [the same approach as for alerting](https://github.com/elastic/terraform-provider-elasticstack/blob/main/Makefile#L242C55-L242C72).

This is what I'd like to do, but not all changes I want are in 8.9. I guess I could just point it at 8.10.

wandergeek:
Ok, just pushed a mammoth update to this. I've broken up the indicators into their own subtypes as you suggested to work around the SDKv2 limitations. There's still a little clean up work to do, but thought I'd share so you could have a look overnight (if you have time).

For tomorrow, I need to work out whether the custom metrics indicator will actually ever have multiple metric objects. Still question marks here.

dimuon:
@wandergeek , it looks good 👍

dimuon (approved):
LGTM 👍

internal/clients/kibana/slo.go
```diff
@@ -21,6 +21,9 @@ func GetSlo(ctx context.Context, apiClient *clients.ApiClient, id, spaceID strin
 	ctxWithAuth := apiClient.SetSloAuthContext(ctx)
 	req := client.GetSloOp(ctxWithAuth, "default", id).KbnXsrf("true")
 	sloRes, res, err := req.Execute()
+	if err != nil && res == nil {
```
tobio:
```suggestion
	if res == nil {
```

This will still panic if `res == nil && err == nil`. Granted that probably shouldn't be possible, but asserting that `err != nil` here seems a little pointless.

tobio (approved):

tobio (approved):

Issue #397, opened 2023-08-09
[DOCS] elasticstack_kibana_action_connector secrets is Required


przemekhys:
The "Secret" field is mandatory for every request, it must always be provided even if it is empty

https://www.elastic.co/guide/en/kibana/8.5/create-connector-api.html

If not used, an error will be returned

`Error: Bad Request: [request body.secrets]: expected value of type [object] but got [null]`

currently in the docs is optional

https://github.com/elastic/terraform-provider-elasticstack/blob/main/docs/resources/kibana_action_connector.md?plain=1#L61

przemekhys:
Add pull request https://github.com/elastic/terraform-provider-elasticstack/pull/398

tobio:
From the Kibana docs:

> Secrets configuration properties vary depending on the connector type. For information about the secrets configuration properties, refer to [Connectors](https://www.elastic.co/guide/en/kibana/8.5/action-types.html).

Some connector types require secrets, others do not. It sounds like the Kibana docs could do with some improvement here, but the connector schema is correct here.

PR #398, opened 2023-08-09
secret field as mandatory, adding the secret field to the example

przemekhys:
A small improvement in the documentation should make it easier for new users to use the provider.

https://github.com/elastic/terraform-provider-elasticstack/issues/397

### DIFF
--- a/docs/resources/kibana_action_connector.md
+++ b/docs/resources/kibana_action_connector.md
@@ -24,7 +24,7 @@ resource "elasticstack_kibana_action_connector" "example" {
     refresh = true
   })
   connector_type_id = ".index"
-  secrets = jsonencode({})
+  secrets           = jsonencode({})
 }
 
 resource "elasticstack_kibana_action_connector" "pagerduty-connector" {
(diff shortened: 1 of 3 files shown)


docs/resources/kibana_action_connector.md
tobio, line 1: These docs are generated, instead you'll want to:

* Update the provider [schema](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/kibana/connector.go#L52 ) to be `Required: true`
* Add the secrets to the index [example](https://github.com/elastic/terraform-provider-elasticstack/blob/main/examples/resources/elasticstack_kibana_action_connector/resource-index-connector.tf)

przemekhys: thanks for the guidance, I adjusted as indicated

tobio:
The failing acceptance tests suggest this parameter is in fact not required by all connector types (they were passing when secrets were not being supplied previously). I'm going to close this PR since the schema for this field is currently correct.

PR #419, opened 2023-09-08
Add support for the .slack_api connector

nkvoll:


generated/connectors/bundled.yaml
```diff
@@ -2590,6 +2664,7 @@ components:
         - .servicenow-sir
         - .server-log
         - .slack
+        - .slack-api
```
tobio: It doesn't look like this in the version of this spec in the [Kibana repo](https://raw.githubusercontent.com/elastic/kibana/main/x-pack/plugins/actions/docs/openapi/bundled.yaml) is it in the process of getting added there?

tobio (dismissed):
LGTM

The [readme](https://github.com/elastic/terraform-provider-elasticstack/blob/main/generated/connectors/README.md) for this generated client links through to the 8.7 version of the API spec. Can we update the reference there as well?

generated/connectors/bundled.yaml
nkvoll: It seems to be missed in that location in the OpenAPI spec, I'm reaching out to them to eventually find out what's up. I have confirmed manually that it exists in the response payload I get from Kibana.

dimuon (approved):
LGTM

Issue #345, opened 2023-05-31
Non Elasticsearch resources should not require Elasticsearch configuration. 


tobio:
In https://github.com/elastic/terraform-provider-elasticstack/issues/342 the Kibana space resource is fetching the ES Cluster ID in order to build the resource ID. 

It should be possible to use the provider to manage Kibana resources without also configuring Elasticsearch endpoints.

PR #425, opened 2023-09-12
Don't require elasticsearch for Kibana resources

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/345

Kibana spaces shouldn't need Elasticsearch to be configured to work.

### DIFF
--- a/internal/kibana/space.go
+++ b/internal/kibana/space.go
@@ -2,7 +2,6 @@ package kibana
 
 import (
 	"context"
-	"strings"
 
 	"github.com/disaster37/go-kibana-rest/v8/kbapi"
 	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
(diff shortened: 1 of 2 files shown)


dimuon (approved):
LGTM 👍, it just needs changelog :)

dimuon:
Still, it looks like we also have to change [the way](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/clients/api_client.go#L704) kibana client is built.

internal/kibana/space.go
```diff
@@ -136,11 +132,14 @@ func resourceSpaceRead(ctx context.Context, d *schema.ResourceData, meta interfa
 	if diags.HasError() {
 		return diags
 	}
-	compId, diags := clients.CompositeIdFromStr(d.Id())
-	if diags.HasError() {
-		return diags
+	id := d.Id()
+	if strings.Contains(id, "/") {
```
dimuon: It's a trivial check, but maybe it makes sense to create a helper function (e.g. `IsCompositeId`) next to `func CompositeIdFromStr` to keep the details of composite ID creation in one place.

tobio:
@dimuon I might defer this one until https://github.com/elastic/terraform-provider-elasticstack/pull/343 is done since it will simplify the creation changes.

dimuon (approved):
LGTM

PR #434, opened 2023-09-27
Update fleet output resource to new API schema format

### LINKED ISSUE #428: Update elasticstack_fleet_output resource to use new API schema types
Starting in fleet/kibana API version 8.10.0, the fleet output types have been updated to use more strongly typed data structures to hold the output configuration. The `elasticstack_fleet_output` resource type will need to be updated to support these different structures. This issue will handle supporting the `elasticsearch` and `logstash` types (in general, already supported by the resource) and will lay the foundation to allow `kafka` to be added in the future (see #395).

taylor-swanson:
- Updated fleet output to use new schema format, where each output kind now has a strongly typed data structure to hold the relevant configuration items.
- Relocated the fleet api generators to match repo style, and cleaned up the schema downloader and filtering functions.
- Expose 'ca_trusted_fingerprint' config value
- Add test case for logstash output
- Update Fleet API version to `v8.10.0`

Closes #428


tobio:
@taylor-swanson what happens when this is run against an older Elasticsearch version?

tobio (approved):
LGTM pending the BWC question.

taylor-swanson:
@tobio

> It looks like we're still running the relevant acceptance tests on 8.6+ so I'm assuming there's no compatibility issues?

That's my understanding as well. The changes to the Fleet OpenAPI schema were to better define the output types and how the output API should be used. As far as I can tell, this doesn't actually change what data we send to the Fleet API. What it did change was the code that was generated, which necessitated this change. Thankfully, we have the acceptance tests running against the different versions to validate things still work.

PR #442, opened 2023-10-10
CP-2463/BK pipeline for releasing

dimuon:


.buildkite/release.yml
```diff
@@ -0,0 +1,6 @@
+steps:
+  - label: Release
+    agents:
+      image: golang:${GO_VERSION}
```
tobio, line 4: Should we define a default for this within the pipeline, otherwise I think we can set something like this at the top of the pipeline so the build fails if the version isn't set:

```yaml
env:
  GO_VERSION: ${GO_VERSION?}
```

catalog-info.yaml
```diff
@@ -0,0 +1,29 @@
...
+      name: terraform-provider-elasticstack / release
+      description: Terraform provider Elasticstack - release
+    spec:
+      repository: elastic/terraform-provider-elasticstack
+      pipeline_file: ".buildkite/release.yml"
+      teams:
+        control-plane-stateful-applications:
+          access_level: MANAGE_BUILD_AND_READ
+        everyone:
+          access_level: READ_ONLY
```
tobio, line 31: So I wouldn't be able to trigger a release with this permission set (which in theory should be fine). I wonder if we might want to grant manage build to serverless-core-arch or something as well. 

It's not clear what the plans for triggering this are, if we're expecting to have this pipeline triggered when a vX.Y branch is pushed then I guess it shouldn't matter too much.

tobio (dismissed):
What's here lgtm, couple of small questions.

catalog-info.yaml
dimuon: My idea was to start with manual triggering so I guess we can grant manage perms to your team as well.
Maybe later we can add a GH workflow trigger too.

dimuon: @tobio, I fixed perms for your group.

.buildkite/release.yml
dimuon: Added a hook with a default for the var.

PR #448, opened 2023-10-13
Fix Fleet Agent Policy monitoring settings

### LINKED ISSUE #432: [Bug] Updating Existing Fleet Policy to Remove `Agent Logging` Fails
# Describe the bug

Agent logging does not update from `elastic_fleet_agent_policy` if both `monitor_logs: false` and `monitor_metrics: false` are set.

# To Reproduce

1. Define the following agent policy:
  ```
  resource "elasticstack_fleet_agent_policy" "my_fleet_agent_policy" {
    name            = "my_fleet_agent_policy"
    namespace       = "default"
    sys_monitoring  = true
    monitor_logs    = true
    monitor_metrics = false
  }
  ```
2. Apply the configuration.
3. Review new fleet policy to validate that `Collect agent logs` is checked under settings.
4. Set `monitor_logs = false` in the agent policy.
5. Apply the updated configuration.
6. Review updated fleet policy to validate that `Collect agent logs` is still checked under settings; `terraform.tfstate` also shows `monitor_logs: true`.

# Expected behavior
`Collect agent logs` should not be set.

> **Note:** If `monitor_logs: false` and `monitor_metrics: true` are both set, then the results are as expected.

# Debug output

```json
PUT /api/fleet/agent_policies/<policy-id> HTTP/1.1
Host: <redacted>
User-Agent: Go-http-client/1.1
Content-Length: 158
Authorization: ***************************************************
Content-Type: application/json
Kbn-Xsrf: true
Accept-Encoding: gzip

{
 "data_output_id": null,
 "download_source_id": null,
 "fleet_server_host_id": null,
 "monitoring_output_id": null,
 "name": "my_fleet_agent_policy",
 "namespace": "default"
}
```

# Versions (please complete the following information)
 - OS: macos
 - Terraform Version: `1.5.7`
 - Provider version: `0.7.0, 0.8.0`
 - Elastic Cloud Version: `8.10.2`

# Additional context

This can be reproduced with the API from Dev Tools in Kibana. To do so, follow steps 1-3 under "steps to reproduce," then apply the following update to the policy from Dev Tools to get the same results:

```json
PUT kbn:/api/fleet/agent_policies/<policy-id>
{
 "data_output_id": null,
 "description": "Test Agent Policy",
 "download_source_id": null,
 "fleet_server_host_id": null,
 "monitoring_output_id": null,
 "name": "test-policy-1",
 "namespace": "default"
}
```

To get the expected results, it is necessary to provide an empty array for `monitoring_enabled`, such as:

```json
PUT kbn:/api/fleet/agent_policies/<policy-id>
{
 "data_output_id": null,
 "description": "Test Agent Policy",
 "download_source_id": null,
 "fleet_server_host_id": null,
 "monitoring_output_id": null,
 "monitoring_enabled": [],
 "name": "test-policy-1",
 "namespace": "default"
}
```

Explicitly setting `monitoring_enabled: null` generates the following error:

```json
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": "[request body.monitoring_enabled]: expected value of type [array] but got [null]"
}
```

taylor-swanson:
- Fixed Fleet Agent Policy monitoring settings not being applied correctly, particularily when the values would be set to false.
- Update Fleet Agent Policy acceptance tests to verify disabling monitoring works as intended.
- Fix ELASTICSEARCH_INSECURE environment variable not being set.

Related:
- Fixes #432


internal/clients/config/elasticsearch.go
```diff
@@ -207,7 +207,7 @@ func (c elasticsearchConfig) withEnvironmentOverrides() elasticsearchConfig {
 	}
 
 	if insecure, ok := os.LookupEnv("ELASTICSEARCH_INSECURE"); ok {
-		if insecureValue, err := strconv.ParseBool(insecure); err != nil {
+		if insecureValue, err := strconv.ParseBool(insecure); err == nil {
```
tobio, line 210: 👍

internal/fleet/agent_policy_resource_test.go
```diff
@@ -111,8 +111,8 @@ resource "elasticstack_fleet_agent_policy" "test_policy" {
   name        = "%s"
   namespace   = "default"
   description = "This policy was updated"
-  monitor_logs = true
-  monitor_metrics = true
+  monitor_logs = false
+  monitor_metrics = false
```
tobio: Is it worth having these with different values (i.e `logs = true, metrics = false`) to better test the parsing of each type?

tobio (dismissed):
LGTM

internal/fleet/agent_policy_resource_test.go
taylor-swanson: I think that's a great idea, it would act as a good regression test for parsing each type.

tobio (approved):
LGTM

taylor-swanson:
@tobio, looks like the merge cleared your review, so if you could re-approve and merge, that'd be great!

tobio (approved):

PR #447, opened 2023-10-13
Fix fleet resources not having ID set on import

taylor-swanson:
- Fixed fleet objects not having an ID set when importing.
- When creating a new resource, if the ID was not explicitly specified, then the `*_id` value is used from the schema. For all other modes (update, read, delete), the existing/provided ID is used. This resolves the issue where the `*_id` value (even if it was empty) would override the provided ID value (i.e., during import).
- Update fleet documentation to remove kibana space id. Fleet objects aren't scoped by spaces, but rather use their own IDs.

Related issues:
- Fixes #396
- Fixes #369

PR #449, opened 2023-10-16
Fix schema validation for kibana and fleet configuration

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/pull/447#issuecomment-1762141753

The path references for username/password were wrong for both these provider config schemas.


dimuon (approved):
LGTM

PR #451, opened 2023-10-17
Switch to protocol 6

dimuon:


dimuon:
Kibana SLO tests fail for 8.10.3 - have to investigate.

taylor-swanson:
> Kibana SLO tests fail for 8.10.3 - have to investigate.

As an FYI, I noticed these tests failing in my other PRs (#447 and #448), so don't think they're related to this change, but rather something to do with stack version 8.10.x.

tobio (commented):
Code change LGTM. Would be worth a changelog, and an update to the provider docs mentioning TF 1.0+ as a requirement.

dimuon:
@tobio , I addressed your comment, would you mind to approve the PR?

tobio (dismissed):
Lgtm

dimuon:
@tobio , looks like I need your approve one more time - I had to resolve a conflict in CHANGELOG after the 0.10.0 release so your last review was dismissed.

tobio (approved):
Still lgtm

PR #454, opened 2023-10-25
Add Fleet package policies and epm support

taylor-swanson:
- Add Fleet EPM (Elastic Package Manager) support, currently only provides for installing and uninstalling packages.
- Add Fleet Package Policy support.
- Add acceptance tests
- Add documentation


internal/clients/config/elasticsearch.go
```diff
@@ -207,7 +207,7 @@ func (c elasticsearchConfig) withEnvironmentOverrides() elasticsearchConfig {
 	}
 
 	if insecure, ok := os.LookupEnv("ELASTICSEARCH_INSECURE"); ok {
-		if insecureValue, err := strconv.ParseBool(insecure); err != nil {
+		if insecureValue, err := strconv.ParseBool(insecure); err == nil { // TODO: Remove
```
taylor-swanson: I'll remove this once #448 is merged. Had to add it for testing.

internal/fleet/package_resource.go
```diff
@@ -0,0 +1,120 @@
...
+
+	return *hash
+}
+
+func ResourcePackage() *schema.Resource {
+	packageSchema := map[string]*schema.Schema{
+		"name": {
+			Description: "The package name.",
+			Type:        schema.TypeString,
+			Required:    true,
```
tobio, line 26: Should the name/version have `ForceNew: true`?

internal/fleet/package_resource.go
```diff
@@ -0,0 +1,120 @@
...
+			Description: "The package version.",
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"force": {
+			Description: "Set to true to force the requested action.",
+			Type:        schema.TypeBool,
+			Optional:    true,
+		},
+		"skip_destroy": {
```
tobio, line 39: ubernit?

```suggestion
		"skip_uninstall": {
```

I think there's a better name here. You could argue that we're not skipping destroy (the resource _will_ be destroyed by TF, but the package won't be uninstalled). 

We have a `deletion_protection` attribute on the Elasticsearch index resource. That behaves differently to this (destroy will fail unless it's set to false), but I think it sets a precedent for this to be more API action focussed (e.g `skip_uninstall`) rather than TF action focussed.  

That's a lot of words for a minor naming discussion, I'm grasping for a clear argument :)

templates/resources/fleet_package.md.tmpl
```diff
@@ -0,0 +1,20 @@
...
+subcategory: "Fleet"
+layout: ""
+page_title: "Elasticstack: elasticstack_fleet_package Resource"
+description: |-
+  Installs or uninstalls a Fleet integration package.
+---
+
+# Resource: elasticstack_fleet_package
+
+Installs or uninstalls a Fleet integration package.
```
tobio:
```suggestion
Installs or uninstalls a Fleet integration package. The Kibana Fleet UI can be used to view available packages. 
```

It's probably worth linking through to the [docs](https://www.elastic.co/guide/en/fleet/8.10/install-uninstall-integration-assets.html) from here as well, even though they just point back to Kibana.

internal/fleet/package_resource.go
```diff
@@ -0,0 +1,120 @@
...
+
+func ResourcePackage() *schema.Resource {
+	packageSchema := map[string]*schema.Schema{
+		"name": {
+			Description: "The package name.",
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"version": {
+			Description: "The package version.",
```
tobio, line 29: IIRC it's possible to just install the 'latest'. Am I making that up? If not it might be worth adding how to do that here.

internal/fleet/package_resource.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/elastic/terraform-provider-elasticstack/internal/utils"
+)
+
+func getPackageID(name, version string) string {
+	hash, _ := utils.StringToHash(name + version)
+
+	return *hash
+}
+
+func ResourcePackage() *schema.Resource {
```
tobio, line 20: It feels like this resource may be difficult to use without a data source to list the available packages. I'm thinking about how I'd use this right now, where I'd be flicking between the Kibana UI and TF code, but what happens when the package in the repo is upgraded to a new version? 

Do you have plans to add that data source in a follow up? If not could we?

internal/fleet/package_resource.go
```diff
@@ -0,0 +1,120 @@
...
+			Description: "The package name.",
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"version": {
+			Description: "The package version.",
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"force": {
```
tobio, line 34: The [docs](https://www.elastic.co/guide/en/fleet/8.10/install-uninstall-integration-assets.html#reinstall-integration-assets) have a section on re-installing. IIUC this isn't relevant here, and and if someone did _really_ want to reinstall via TF they could `terraform taint <package>` but just checking if I'm missing anything.

templates/resources/fleet_package_policy.md.tmpl
```diff
@@ -0,0 +1,31 @@
...
+
+# Resource: elasticstack_fleet_package_policy
+
+Creates or updates a Fleet Package Policy.
+
+It is highly recommended that all inputs and streams are provided in the
+Terraform plan, even if some are disabled. Otherwise, differences may appear
+between what is in the plan versus what is returned by the Fleet API.
+
+The Kibana Fleet UI can be used as a reference for what data needs to be provided.
```
tobio: Similarly it might be worth pointing back to the [docs](https://www.elastic.co/guide/en/fleet/8.10/add-integration-to-policy.html) here too.

internal/fleet/package_resource.go
tobio: We have `skip_destroy` on the agent policy, let's be consistent with that.

tobio (commented):
Couple of usability questions on the package resource.

internal/fleet/package_resource.go
taylor-swanson: I definitely agree that `skip_uninstall` would make more sense in this context, but yeah, I called it `skip_destroy` to be consistent with agent policies.

internal/fleet/package_resource.go
taylor-swanson: Doesn't look like it, unfortunately. I tried putting in `latest` and Kibana kicked back a 500 (also did this just hitting the API directly through Postman). I checked the logs and it just said `500 server error` 🤷 

```
[2023-10-31T13:13:37.007+00:00][ERROR][http] 500 Server Error
```

I believe it wants a semver, since if you give it something that looks more like semver, but isn't, then it produces a semver validation error. If I had to guess, Kibana is throwing an exception trying to parse `latest` as a semver, but not surfacing the error.

internal/fleet/package_resource.go
taylor-swanson: These are great points. I've been trying to think of usability concerns (and there are a few when dealing with packages and package policies).

We didn't have explicit plans for a package data source, but we can definitely look into doing that. At a quick glance, the "list packages" endpoint for EPM seems to list the latest package versions only (it doesn't give you a full manifest of what's available in the package registry), as far as I can tell. This should be fine, though, since we really only care about the latest package version.

internal/fleet/package_resource.go
taylor-swanson: Internally, I'm not sure what Kibana is doing to reinstall (if there's something special happening, or if it is just doing an uninstall/install). The API doesn't have anything for reinstall. As long as taint triggers an uninstall/install, I think we're good here.

internal/fleet/package_resource.go
taylor-swanson: Yeah that makes sense, I'll add it.

internal/fleet/package_resource.go
tobio: Wdyt about timing here? I'm inclined to merge this and have the data sources in a follow up?

taylor-swanson: I would agree as well. I can write a follow up issue for the package data source.  (issue here: #468)

tobio (approved):
LGTM 🚀

Kushmaro:
@taylor-swanson @tobio , we should definitely have a package data resource as well

taylor-swanson:
@tobio @Kushmaro, see #469 for the package data source PR.

Kushmaro:
hey @taylor-swanson , i'm looking into the UX of the new resources, I might want us to change their name / terminology a bit.
the Fleet UI doesn't mention or discusses at all "Package policies" rather, "integrations". That is what our users know... our UX / wording should be consistent where we can have them so

Kushmaro:
I agree @taylor-swanson that it's not best to wait here. 
I know the API refers to package policies, but that kind of excludes the actual user experience. when users go to the fleet management UI , they have no concept of "package policies" 

the fact that the UI & API use different terminologies is a separate issue (mainly for the fleet team) 
but i'd want us to align the Terraform client experience towards what the user will natively understand from the UI / Product.

taylor-swanson:
@Kushmaro, that makes sense. I'll write up an issue to rename the package policy resource and update documentation. Do you have any thoughts about the package resource and data source (`elasticstack_fleet_package`)? There are a few things that identify them as "packages" (Elastic Package Registry, Elastic Package Manager, etc), but the Kibana UI and the main docs site seems to refer to them as "integrations" at this point. Although, I thought I've heard that not every package is an integration? (I could be wrong about that, though).

taylor-swanson:
Link to refactoring issue: https://github.com/elastic/terraform-provider-elasticstack/issues/475

nimarezainia:
agree with both @taylor-swanson and @Kushmaro - average user of the platform would be more familiar with the concept of an "integration" rather than packages which are perhaps more developer focuses. Thanks for making all these changes.

PR #469, opened 2023-11-02
Add data source for Fleet integration packages

### LINKED ISSUE #468: Create data source for Fleet integration packages
With the addition of Fleet package policies in #454, it would be a good idea to include a data source for Fleet packages as a quick way to get the latest version of a particular package.

The Fleet API has an endpoint to get all the latest packages (`GET /epm/packages`), we would just need to filter this list down and extract the version. The endpoint also has parameters for including prerelease and experimental packages. I'm not sure yet how this affects the results, so we may need to handle multiple returned results and have logic to select the latest version (i.e., sorting based on semver).

taylor-swanson:
- Used for getting the latest version of an integration package
- Add acceptance test
- Add docs

## Related

- Closes #468


tobio (approved):
LGTM

PR #477, opened 2023-11-10
Increase Error and Logging Verbosity for Connections

RoseSecurity:
## Why:

- I encountered a connection error with Elasticsearch where I accidentally attempted to connect over HTTP rather than HTTPS. Received a blanket "EOF" error each time

### Curl Ouput:

```
➜  ~ curl -u elastic:test http://10.0.0.100:9200 -v -k 
*   Trying 10.0.0.60:9200...
* Connected to 10.0.0.100 (10.0.0.60) port 9200 (#0)
* Server auth using Basic with user 'elastic'
> GET / HTTP/1.1
> Host: 10.0.0.60:9200
> Authorization: Basic test
> User-Agent: curl/8.1.2
> Accept: */*
> 
* Empty reply from server
* Closing connection 0
curl: (52) Empty reply from server
```

### Terraform Output:

Terraform Output

## Goals:

- Increase verbosity for debug logging
- Update from `v7` to `v8` for Elasticsearch Go client packages 

## What:

- Updated from the deprecated `estransport` to `elastic-transport-go`
- Cleaned up an `else` statement
- Utilized the JSON logger

## Future Development Plans:

- Can we add more visible logging for empty server responses?


internal/clients/config/debug.go
```diff
@@ -22,7 +23,7 @@ const logRespMsg = `%s API Response for [%s] Details:
 %s
 -----------------------------------------------------`
 
-var _ estransport.Logger = &debugLogger{}
+var _ elastictransport.Logger = &elastictransport.JSONLogger{}
```
tobio, line 26: We don't need this change, this code is just checking that `debugLogger` implements the logger Logger interface. 

https://golangbyexample.com/go-check-if-type-implements-interface/

To change the logger you'll want to change [this](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/clients/config/elasticsearch.go#L108-L111) block. 

That said, I'm not sure why we need to change the logger, rather than just having the code change you've made here?

PR #372, opened 2023-06-30
Kibana api key

andrianjardan:
Fixes #364

PR #481, opened 2023-11-15
Add bearer token header

SquareShark:
- Added the Authorization field, this will allow the user to just put in a token, the base config will create the `"Authorization"` : "Bearer <value>"` so the value from the provider will replace <value>
- Added the ES-Client-Authentication field. The base config will create `"ES-Client-Authentication" : "value"`


internal/clients/config/base.go
```diff
@@ -23,6 +23,20 @@ func newBaseConfigFromSDK(d *schema.ResourceData, version string, esKey string)
 		Header:    http.Header{"User-Agent": []string{userAgent}},
 	}
 
+	if elasticDataRaw, ok := d.GetOk("elasticsearch"); ok {
```
tobio: This should be added in `config/elasticsearch` instead. It's not supported by the Kibana API.

internal/schema/connection.go
```diff
@@ -250,7 +267,22 @@ func GetEsConnectionSchema(keyName string, isProviderConfiguration bool) *schema
...
 					DefaultFunc:   withEnvDefault("ELASTICSEARCH_API_KEY", nil),
-					ConflictsWith: []string{usernamePath, passwordPath},
+					ConflictsWith: []string{usernamePath, passwordPath, bearerTokenPath},
+				},
+				"bearer_token": {
+					Description:   "Bearer Token to use for authentication to Elasticsearch",
+					Type:          schema.TypeString,
+					Optional:      true,
+					Sensitive:     true,
+					DefaultFunc:   withEnvDefault("ELASTICSEARCH_BEARER_TOKEN", nil),
```
tobio, line 289: These environment variables should be read in `config/elasticsearch` instead. These schema defaults won't be read as part of TF Plugin Framework resources so for now we're centralising them in the config code.

tobio (commented):
Thanks for looking into this, will be great to support token auth.

We'll want to move some of this code around before merging.

tobio (commented):
We'll also want to add an entry to [CHANGELOG](https://github.com/elastic/terraform-provider-elasticstack/blob/main/CHANGELOG.md) and manually [document](https://github.com/elastic/terraform-provider-elasticstack/blob/main/templates/index.md.tmpl#L52-L56) the new environment variables  too :)

SquareShark:
@tobio All changes have been updated as suggested, Thank you for the help!, I also cleaned up the couple of code smells too to make it a bit more tidy now I am getting more used to GO.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Added
 - Switch to Terraform [protocol version 6](https://developer.hashicorp.com/terraform/plugin/terraform-plugin-protocol#protocol-version-6) that is compatible with Terraform CLI version 1.0 and later.
 - Add 'elasticstack_fleet_package' data source ([#469](https://github.com/elastic/terraform-provider-elasticstack/pull/469))
+- Add support for Authoization header - Bearer Token and ES-Client-Authentication fields added.
```
tobio:
```suggestion
- Add support for Authorization header - Bearer Token and ES-Client-Authentication fields added.
```

internal/clients/config/elasticsearch.go
```diff
@@ -38,6 +38,14 @@ func newElasticsearchConfigFromSDK(d *schema.ResourceData, base baseConfig, key
 			config.Addresses = addrs
 		}
 
+		if bearer_token, ok := esConfig["bearer_token"].(string); ok && bearer_token != "" {
+			base.Header.Set("Authorization", bearer_token)
+		}
+
+		if es_client_authentication, ok := esConfig["es_client_authentication"].(string); ok && es_client_authentication != "" {
+			base.Header.Set("ES-Client-Authentication", es_client_authentication)
+		}
```
tobio: This will work for ES (unfortunately, it kind of exposes a bug in how we deal with this header collection TBH), but is likely to break Kibana/Fleet clients. 

We should:
* Give the ES client it's own copy of the base headers `b.Header.Clone()` [here](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/clients/config/base.go#L80)
* Set these headers on the ES config rather than the base config. 

```suggestion
		if bearer_token, ok := esConfig["bearer_token"].(string); ok && bearer_token != "" {
			config.Header.Set("Authorization", bearer_token)
		}

		if es_client_authentication, ok := esConfig["es_client_authentication"].(string); ok && es_client_authentication != "" {
			config.Header.Set("ES-Client-Authentication", es_client_authentication)
		}
```

internal/clients/config/elasticsearch.go
```diff
@@ -132,6 +140,13 @@ func newElasticsearchConfigFromFramework(ctx context.Context, cfg ProviderConfig
 		config.Addresses = endpoints
 	}
 
+	if esConfig.BearerToken.ValueString() != "" || esConfig.BearerToken.ValueString() != "null" {
```
tobio:
```suggestion
	if esConfig.BearerToken.ValueString() != "" {
```

If the value is `null` in the TF module, ValueString will return `""`

internal/clients/config/elasticsearch.go
```diff
@@ -132,6 +140,13 @@ func newElasticsearchConfigFromFramework(ctx context.Context, cfg ProviderConfig
...
+		config.Header.Set("Authorization", esConfig.BearerToken.ValueString())
+		if esConfig.ESClientAuthentication.ValueString() != "" || esConfig.ESClientAuthentication.ValueString() != "null" {
```
tobio:
```suggestion
		if esConfig.ESClientAuthentication.ValueString() != "" {
```

Similarly

internal/schema/connection.go
```diff
@@ -43,9 +45,22 @@ func GetEsFWConnectionBlock(keyName string) fwschema.Block {
...
+					Optional:            true,
+					Sensitive:           true,
+					Validators: []validator.String{
+						stringvalidator.ConflictsWith(usernamePath, passwordPath, apiKeyPath),
+					},
+				},
+				"es_client_authentication": fwschema.StringAttribute{
+					MarkdownDescription: "ES Client Authentication field to be used with the bearer token",
+					Optional:            true,
+					Sensitive:           true,
```
tobio, line 62:
```suggestion
					Sensitive:           true,
					Validators: []validator.String{
						stringvalidator.ConflictsWith(usernamePath, passwordPath, apiKeyPath),
						stringvalidator.AlsoRequires(bearerTokenPath),
					},
```

It only really makes sense for this to be used in conjunction with a bearer token IIUC.

tobio (commented):
Couple of more changes sorry. I wonder if it makes sense to add a helper to generate a service account token in the Makefile like has been added [here](https://github.com/elastic/terraform-provider-elasticstack/pull/372/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R133-R136) as well to make testing this simpler.

SquareShark:
I have now got all commits signed with a new gpg key, was having major issues but all sorted now. I have also made all the changes you have suggested. Thank you so much for your patience and help with this

tobio:
It looks like [this](https://github.com/elastic/terraform-provider-elasticstack/pull/481/commits/2aa260b994d3d6fcb7661d1487be8bf22b0d1ae5) commit is using a different email to what's published (`.com` rather than `.co.uk`) which is causing the CLA checker to fail. Are you able to rewrite the author on that one to feed the machine?

tobio:
It looks like the lint failure can be fixed by running `make lint` locally and committing the result.

tobio (dismissed):
🎉 this is awesome, thanks for taking the time to make this addition!

SquareShark:
I just done the Lint request to `make fmt` then recommitted the file that it formatted.

SquareShark:
somehow I have royally messed this up, I am so sorry haha

PR #485, opened 2023-11-20
Kibana SLOs: use `spaceID`

afharo:
When creating the SLOs against a custom space ID, I'm getting the following errors:

```
╷
│ 
│ When applying changes to elasticstack_kibana_slo.kibana_http_latency_slo, provider "provider[\"terraform.local/elastic/elasticstack\"]" produced an unexpected new value: Root resource was present, but now absent.
│ 
╵
```

This is caused because TF is correctly creating it in the indicated space, but retrieving it from the `default` one.


tobio (approved):

tobio:
Thanks for taking the time fix this one up 👍

PR #492, opened 2023-11-23
Fix buildkite release pipeline

nkammah:
Fixes the release pipeline:
* uses the proper vault path (there was no value defined previously)
* fixes go version discrepancy
* limit parralelism to avoid build failure
* proper secrets handling

Note: atm, this is just running `make release-no-publish` until we can have the job run only on tag pushes.


dimuon (approved):
Thanks for fixing this

Issue #75, opened 2022-02-22
[Feature] Add the ability to add data views (ex index pattern)


CondormanFr:
When setting up a new instance, needed to create data views (ex index pattern)

https://www.elastic.co/guide/en/kibana/current/data-views-api.html

audig:
this can be very usefull, an exemple of the resource declaration that can be implemented  
```
resource "elasticstack_kibana_dataview" "mydataview" {
  name           = "log"
  pattern = "log-*"
  timeFieldName = "@timestamp"
}
```

SHolzhauer:
This is something we would like to have as well, however if this is implemented we would need the ability to configure the dataview(s) to be shared across all spaces, specifically in 1 or shared across multiple.

nicholas-r-king:
This would be very useful for us. (Enterprise customer on ECK)

mholttech:
As an Enterprise customer I am also looking for this

nicholas-r-king:
Any update on this? I see it was added to the 0.8 milestone but the project is now on 0.10.

Blefish:
My poor man's workaround to this (and dashboards)

The form data is an index-pattern, exported via Kibana's saved objects UI. Would love a native resource though!

```
resource "terraform_data" "object" {
  triggers_replace = filesha1(var.object_path)
  provisioner "local-exec" {
    command = join(" ", [
      "curl --fail -X POST --insecure",
      "--user '${var.kibana_user}:${var.kibana_password}'",
      "--header 'kbn-xsrf: true'",
      "--form file=@${var.object_path}",
      "${var.kibana_url}/api/saved_objects/_import?overwrite=true",
    ])
  }
}
```

bartoszcisek:
[Elastic query alerting rule](https://www.elastic.co/guide/en/kibana/8.10/rule-type-es-query.html) definition require dataView ID as an argument. Without data source for `dataView` we need to hardocde UUID. 

Having dataView resource or at least dataView data source be super helpful.

PR #502, opened 2023-12-06
Add data view resource

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/75

Adds a new resource to manage Kibana data views. This also updates some of the existing generated clients to use the latest codegen version.


Makefile
```diff
@@ -279,17 +279,32 @@ generate-alerting-client: ## generate Kibana alerting client
 	@ rm -rf generated/alerting/go.mod generated/alerting/go.sum generated/alerting/test
 	@ go fmt ./generated/alerting/...
 
+#  https://raw.githubusercontent.com/elastic/kibana/main/src/plugins/data_views/docs/openapi/bundled.yaml \
```
wandergeek: Do we want to keep this comment considering the one below has been removed?

internal/kibana/data_view/schema_test.go
```diff
@@ -0,0 +1,329 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var tfModel tfModelV0
+			diags := tfsdk.ValueFrom(context.Background(), tt.model, getSchema().Type(), &tfModel)
+			require.Nil(t, diags)
+
+			req, diags := tfModel.ToCreateRequest(context.Background())
+
+			require.Equal(t, tt.expectedRequest, req)
+			require.Equal(t, tt.expectedDiags, diags)
```
wandergeek, line 125: Probably looking at this wrong, but do we not want to test cases where this function returns diags? I don't see `expectedDiags` in any of the test cases.

wandergeek (approved):
Looks good. I feel like putting the regenerated clients in another PR might be better, but not essential. People will be stoked about this.

internal/kibana/data_view/schema_test.go
tobio: It would be nice to test some of the type conversions. It's quite difficult to do so, TF doesn't make it easy to define nested TF objects which is what we'd need to do to test the error cases here.

wandergeek (approved):

Issue #501, opened 2023-12-05
[Bug] role mappings being created when name is changed but not replacing or removing what is replaces


SquareShark:
As the title says, Role mappings are just being created if you change the name of them, rather than the old resources being updated or removed. In the [Roles.go](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/security/role.go) inside the name block on Line 30, you have `ForceNew:    true,`. But nothing similar in the [role_mappings.go](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/security/role_mapping.go) name block.

Can we add that little line into the role_mapping.go pretty please.

tobio:
Thanks again for taking the time to improve the provider @SquareShark!

tobio:
> Just curious on when the next planned release is?

@SquareShark I've just released 0.11 including these changes :)

PR #503, opened 2023-12-06
Updating role mapping name field to update or delete role mappings wh…

SquareShark:
https://github.com/elastic/terraform-provider-elasticstack/issues/501
- Update the name field for the role_mappings.go to "ForceNew", I was seeing the issue where if you update the name for a role mapping it would create a new role mapping but not actually remove the one it was replacing. Role.go has this feature in already so figured this was just missed out and wanted to add it in.


tobio (approved):
Nice! Thanks for fixing this one up.

PR #508, opened 2023-12-14
Do not set LastExecutionDate if it is `nil`

gsoldevila:
`NilPointer` can happen if we try to read and update rules which have not been executed.
This PR protects the code in that scenario.


tobio (dismissed):
LGTM. Thanks for taking the time to fix this one up!

tobio (approved):

PR #517, opened 2023-12-19
Secured release pipeline

nkammah:
This change, related to https://github.com/elastic/ci/pull/2506 updates the release script to read the PGP keys from a more secured location shared with less folks. The GPG key information is now in a secured shared location accessible only to members of the GH group `terraform-providers-admins` and to the CI pipelines of this repository and terraform-provider-elasticstack.

Note - I also updated the pipeline to actual release the binary (it was set to `-no-publish` before).

Related change: https://github.com/elastic/terraform-provider-ec/pull/758


dimuon (approved):
LGTM

Issue #506, opened 2023-12-13
[Bug] Import of elasticstack_fleet_output resource fails to import data


glisignoli:
**Describe the bug**
When running `terraform import elasticstack_fleet_output.test_output be7628c0-9967-11ee-9129-e3b2c5beabce`, the resource imports successfully, but fails to import any fleet data (except the id).

**To Reproduce**
1. TF configuration used:
```
resource "elasticstack_fleet_output" "test_output" {
  name                 = "test_output"
  type                 = "elasticsearch"
  hosts                = ["https://test.example.com:443"]
  default_integrations = false
  default_monitoring   = false
}
```
3. TF operations to execute to get the error: `terraform import elasticstack_fleet_output.test_output be7628c0-9967-11ee-9129-e3b2c5beabce`
4. See the error in the output:
```
...
...
elasticstack_fleet_output.test_output: Import prepared!
  Prepared elasticstack_fleet_output for import
...
...

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```
But:
```
terraform state show elasticstack_fleet_output.test_output
# elasticstack_fleet_output.test_output:
resource "elasticstack_fleet_output" "test_output" {
    id = "be7628c0-9967-11ee-9129-e3b2c5beabce"
}
```

**Expected behavior**
The import of the elasticstack_fleet_output resource should contain all data from the fleet output

**Debug output**
Run `terraform` command with `TF_LOG=trace` and provide extended information on TF operations.
```
elasticstack_fleet_output.test_output: Importing from ID "be7628c0-9967-11ee-9129-e3b2c5beabce"...

elasticstack_fleet_output.test_output: Import prepared!
  Prepared elasticstack_fleet_output for import

2023-12-13T16:36:46.716+1300 [DEBUG] provider.terraform-provider-elasticstack_v0.11.0: Fleet API Request Details:
---[ REQUEST ]---------------------------------------
GET /api/fleet/outputs/be7628c0-9967-11ee-9129-e3b2c5beabce HTTP/1.1
Host: kibana.example.com
User-Agent: Go-http-client/1.1
Authorization: ***********************************************
Accept-Encoding: gzip


[...]

PR #522, opened 2024-01-01
Populate all relevant attributes during fleet output import

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/506

Uses the returned output type rather than relying on internal state for parsing the response during read/import.


taylor-swanson (approved):
LGTM

PR #538, opened 2024-01-25
Add downsample to ILMs

xaviervolvo:
There is a parameter for the hot, warm and cold phases called downsample.

According to the description:

"Roll up documents within a fixed interval to a single summary document. Reduces the index footprint by storing time series data at reduced granularity."

As of 0.11.0 there is no support for this parameter in the proivider, this PR adds it.

### DIFF
--- a/internal/elasticsearch/index/ilm.go
+++ b/internal/elasticsearch/index/ilm.go
@@ -46,7 +46,7 @@ func ResourceIlm() *schema.Resource {
 			MaxItems:     1,
 			AtLeastOneOf: []string{"hot", "warm", "cold", "frozen", "delete"},
 			Elem: &schema.Resource{
-				Schema: getSchema("set_priority", "unfollow", "rollover", "readonly", "shrink", "forcemerge", "searchable_snapshot"),
+				Schema: getSchema("set_priority", "unfollow", "rollover", "readonly", "shrink", "forcemerge", "searchable_snapshot", "downsample"),
 			},
 		},
 		"warm": {
@@ -56,7 +56,7 @@ func ResourceIlm() *schema.Resource {
 			MaxItems:     1,
 			AtLeastOneOf: []string{"hot", "warm", "cold", "frozen", "delete"},
 			Elem: &schema.Resource{
-				Schema: getSchema("set_priority", "unfollow", "readonly", "allocate", "migrate", "shrink", "forcemerge"),
+				Schema: getSchema("set_priority", "unfollow", "readonly", "allocate", "migrate", "shrink", "forcemerge", "downsample"),
 			},
 		},
 		"cold": {
@@ -66,7 +66,7 @@ func ResourceIlm() *schema.Resource {
 			MaxItems:     1,
 			AtLeastOneOf: []string{"hot", "warm", "cold", "frozen", "delete"},
 			Elem: &schema.Resource{
-				Schema: getSchema("set_priority", "unfollow", "readonly", "searchable_snapshot", "allocate", "migrate", "freeze"),
+				Schema: getSchema("set_priority", "unfollow", "readonly", "searchable_snapshot", "allocate", "migrate", "freeze", "downsample"),
 			},
 		},
 		"frozen": {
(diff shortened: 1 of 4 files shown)


xaviervolvo:
Hello, I signed the agreement but this seems stuck. Let me know if i need to do anything else.

internal/elasticsearch/index/ilm.go
```diff
@@ -389,6 +389,26 @@ var supportedActions = map[string]*schema.Schema{
 			},
 		},
 	},
+	"downsample": {
+		Description: "Roll up documents within a fixed interval to a single summary document. Reduces the index footprint by storing time series data at reduced granularity.",
+		Type:        schema.TypeList,
+		Optional:    true,
+		MaxItems:    2,
```
tobio: Why 2? IIUC there's no case where a user should specify more than a single `downsample` block right?

internal/elasticsearch/index/ilm_test.go
```diff
@@ -128,6 +130,11 @@ resource "elasticstack_elasticsearch_index_lifecycle" "test" {
       max_age = "1d"
     }
 
+		downsample {
```
tobio: nit: Can we fix up the indentation on this block please.

tobio (commented):
Thanks for adding this, small question on the schema.

tobio:
@xaviervolvo are you able to add an entry to the CHANGELOG.md file for this one too?

xaviervolvo:
I think I addressed all the questions, thank you.

tobio (approved):
LGTM

tobio (dismissed):
Still LGTM, just was regenerated docs.

xaviervolvo:
I see that acceptance tests are failing, is there anything I can do to solve that?

martijnbakkerelastic:
@tobio When briefly discussing the current status of the PR with Volvo Cars, they shared the feedback on the failed checks that this might have to do with downsample not being implemented in version prior to 8.10. Could this maybe be the cause and if so how can we move this PR forward?

tobio:
> they shared the feedback on the failed checks that this might have to do with downsample not being implemented in version prior to 8.10. Could this maybe be the cause and if so how can we move this PR forward?

It looks like there's three cases here:

* [8.4 and lower](https://github.com/elastic/terraform-provider-elasticstack/actions/runs/7719631590/job/21043205348?pr=538): No support for downsample.
* [8.5 - 8.9](https://github.com/elastic/terraform-provider-elasticstack/actions/runs/7719631590/job/21043206307?pr=538): No support for the `wait_timeout` attribute
* 8.10 and higher: Full support

We don't usually restrict provider functionality based on the target cluster version (i.e there's no provider level version checks, the resource will just fail). That's not ideal, nicer errors would of course be nicer, I'm just stating the status quo. If a nice error message was added here then that would be lovely, but not necessary. 

Instead, we usually filter the acceptance tests, the ILM test already does this for the [shards per node](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/index/ilm_test.go#L65) case. It looks like we should do the same here, i.e create a new test (or two given the 8.5-8.9 case above) case explicitly validating the downscale attributes work, and skipping them on unsupported versions.

xaviervolvo:
I understand. I haven't familiarised myself with the test setup to the point to see how to filter out tests depending on version?  Can you take care of it or would you give me some pointers? Thank you.

tobio:
> Can you take care of it or would you give me some pointers

If you have some time to wrap this one up then it'll get merged and released faster :) 

[This](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/index/ilm_test.go#L65) test case shows how we're skipping test cases based on the target stack version. Let me know if you've got any questions for that, but we've just been defining a minimum version and using this skip mechanism elsewhere.

xaviervolvo:
I created skip versions functions and added the test cases for the different versions as mentioned in the comment above.

tobio (approved):
LGTM

tobio:
@xaviervolvo thanks for making the time for this addition!

Issue #529, opened 2024-01-10
[Bug] elasticstack_kibana_data_view doesn't support API keys for authentication


mathias-ioki:
**Describe the bug**
`elasticstack_kibana_data_view` expect a username and password to be set for authentication. All other resources support also API keys for authentication.

**To Reproduce**
1. Set environment variable `ELASTICSEARCH_API_KEY` and don't set `ELASTICSEARCH_USERNAME` and `ELASTICSEARCH_PASSWORD`
2. Create a new `elasticstack_kibana_data_view` resource via `terraform apply`
3. Data view can't be created because of missing authorization `unable to authenticate user []`

**Expected behavior**
API keys can be used to create also data view resources

**Terraform Output**
```
 Error: Unable to create data view

   with elasticstack_kibana_data_view.dataview-test,
   on data-view.tf line 17, in resource "elasticstack_kibana_data_view" "dataview-test":
   17: resource "elasticstack_kibana_data_view" "dataview-test" {

 Failed with: {"statusCode":401,"error":"Unauthorized","message":"[security_exception\n\tRoot causes:\n\t\tsecurity_exception: unable to authenticate user [] for REST request [/_security/_authenticate]]: unable
 to authenticate user [] for REST request [/_security/_authenticate]"}
```

**Versions (please complete the following information):**
 - Terraform Version v1.5.7
 - Provider version v0.11.0
 - Elasticsearch Version 8.11.2

PR #549, opened 2024-02-01
Support API Keys with data_view resource

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/529

Does what it says on the tin, adds support for API key auth into the data view resource.


dimuon (approved):
LGTM

dimuon:
@tobio , is lack of privilege in the API key causing acceptance tests to fail?

tobio:
Looks like the data view api spec is missing details for API key auth 😢. It's also missing support for Kibana spaces so I'll try and get both of those added in and fix this one up.

Issue #548, opened 2024-01-30
[Bug]  Error when setting override  = true for resource elasticstack_kibana_data_view error


dllegru:
**Describe the bug**
When setting `override = true` for the resource `elasticstack_kibana_data_view` and running `terraform apply`, execution halts with `Error: Provider produced inconsistent result after apply`, `This is a bug in the provider, which should be reported in the provider's own issue tracker.`

**To Reproduce**
1. TF configuration used '...'

```
resource "elasticstack_kibana_data_view" "pocdataview2" {
  data_view = {
    name            = "po2c-dataview"
    title           = "poc2-*-dw"
    time_field_name = "@timestamp"
  }

  override = true
}

```

3. TF operations to execute to get the error '...' [e.g `terraform plan`,`terraform apply`, `terraform destroy`]
`terraform apply`

5. See the error in the output '...'

```
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to elasticstack_kibana_data_view.pocdataview2, provider
│ "provider[\"registry.terraform.io/elastic/elasticstack\"]" produced an unexpected new value: .override: was cty.True, but now
│ cty.False.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

**Expected behavior**

Provider is able to create the resource with `override` setting

**Versions (please complete the following information):**
 - OS: [Ubuntu 22.04] 
 - Terraform Version [1.5.1]
 - Provider version [0.11.0]
 - Elasticsearch Version [8.11.1]

PR #550, opened 2024-02-01
Persist override value during data view read

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/548

Ensures that the override value is correctly persisted when managing data views


dimuon (approved):
LGTM

PR #570, opened 2024-02-22
Updating `set_priority` documentation

ASRagab:
👋🏽 The `set_priority` documentation was incorrect and was a bad copy-paste of the `shrink` documentation. 

https://www.elastic.co/guide/en/elasticsearch/reference/8.12/ilm-set-priority.html
> Phases allowed: hot, warm, cold.
>
> Sets the [priority](https://www.elastic.co/guide/en/elasticsearch/reference/8.12/recovery-prioritization.html) of the index as soon as the policy enters the hot, warm, or cold phase. Higher priority indices are recovered before indices with lower priorities following a node restart.
>
> Generally, indexes in the hot phase should have the highest value and indexes in the cold phase should have the lowest values. For example: 100 for the hot phase, 50 for the warm phase, and 0 for the cold phase. Indices that don’t set this value have a default priority of 1.


tobio (approved):
Thanks for taking the time to improve the docs here!

PR #577, opened 2024-02-29
Support `restriction` in `elasticstack_elasticsearch_security_api_key`

sandermvanvliet-stack:
In its current form the provider is unable to provision API keys that have a restriction specified in the role descriptors.
For example, given this terraform:

```terraform
resource "elasticstack_elasticsearch_security_api_key" "api_key_with_restriction" {
  name = "restricted-key"

  role_descriptors = jsonencode({
    role-a = {
      indices = [
        {
          names      = ["index-a*"],
          privileges = ["read"]
        }
      ],
      restriction = {
        workflows = [ "search_application_query" ]
      }
    }
  })
}
```

The provider will happily create an API key named `restricted-key` however the `restriction` field is not persisted in Elasticsearch.

On the next `terraform plan`, this API key will be marked as needing replacement due to the fact that the comparison between the local terraform definition and the response from the Elasicsearch API are different. That is because the local terraform definition serializes to JSON and will contain `restriction` while the JSON for the existing API key does not have that field.

This PR adds the `restriction` to the model structs necessary to provision the API keys.

In order to achieve that, the `Role` struct was duplicated to an API key specific struct `ApiKeyRoleDescriptor`. Previously `Role` was shared between the `elasticstack_elasticsearch_security_api_key` _resource_ and the `elasticstack_elasticsearch_security_role` _data source_.

In Elasticsearch, the `restriction` is only available from 8.9.x and up. To provide some feedback to users, an error will be generated during plan/apply time when the target Elasticsearch is below 8.9 and the user has specified a `restriction` in their local terraform definition:

> Error: Specifying `restriction` on an API key role description is not supported in this version of Elasticsearch. Role descriptor(s) role-a

This is currently making a call directly to the Elasticsearch instance which is probably not ideal, however I could not find a reference to the current version in the code. If that's available that would be a much better option.

### DIFF
--- a/docs/resources/elasticsearch_security_api_key.md
+++ b/docs/resources/elasticsearch_security_api_key.md
@@ -41,10 +41,10 @@ resource "elasticstack_elasticsearch_security_api_key" "api_key" {
   })
 }
 
+# restriction on a role descriptor for an API key is supported since Elastic 8.9
 resource "elasticstack_elasticsearch_security_api_key" "api_key_with_restriction" {
   # Set the name
   name = "My API key"
-
   # Set the role descriptors
   role_descriptors = jsonencode({
     role-a = {
--- a/docs/resources/elasticsearch_security_api_key.md
+++ b/docs/resources/elasticsearch_security_api_key.md
@@ -56,7 +56,7 @@ resource "elasticstack_elasticsearch_security_api_key" "api_key_with_restriction
         }
       ],
       restriction = {
-        workflows = [ "search_application_query" ]
+        workflows = ["search_application_query"]
       }
     }
   })
--- a/docs/resources/elasticsearch_security_api_key.md
+++ b/docs/resources/elasticsearch_security_api_key.md
@@ -41,6 +41,37 @@ resource "elasticstack_elasticsearch_security_api_key" "api_key" {
   })
 }
 
+resource "elasticstack_elasticsearch_security_api_key" "api_key_with_restriction" {
+  # Set the name
+  name = "My API key"
+
+  # Set the role descriptors
+  role_descriptors = jsonencode({
+    role-a = {
+      cluster = ["all"],
+      indices = [
+        {
+          names      = ["index-a*"],
+          privileges = ["read"]
+        }
+      ],
+      restriction = {
+        workflows = [ "search_application_query" ]
+      }
+    }
+  })
+
+  # Set the expiration for the API key
+  expiration = "1d"
+
+  # Set the custom metadata for this user
+  metadata = jsonencode({
+    "env"    = "testing"
+    "open"   = false
+    "number" = 49
+  })
+}
+
 output "api_key" {
   value     = elasticstack_elasticsearch_security_api_key.api_key
   sensitive = true
--- a/internal/elasticsearch/security/api_key.go
+++ b/internal/elasticsearch/security/api_key.go
@@ -3,6 +3,7 @@ package security
 import (
 	"context"
 	"encoding/json"
+	"github.com/elastic/go-elasticsearch/v7/esapi"
 	"regexp"
 	"strings"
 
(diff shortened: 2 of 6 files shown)


docs/resources/elasticsearch_security_api_key.md
```diff
@@ -41,6 +41,37 @@ resource "elasticstack_elasticsearch_security_api_key" "api_key" {
   })
 }
 
+# restriction on a role descriptor for an API key is supported since Elastic 8.9
```
tobio, line 44: This is failing lint because it's not consistent with the result of docs generation. Are you able to run `make docs-generate` locally and get things trued up?

internal/elasticsearch/security/api_key.go
```diff
@@ -155,6 +187,36 @@ func resourceSecurityApiKeyCreate(ctx context.Context, d *schema.ResourceData, m
...
+			return false, err
+		}
+
+		supportedVersion, _ := version.NewVersion("8.9.0") // restriction is supported since 8.9.x
+
+		if currentVersion.LessThan(supportedVersion) {
+			return false, err
+		}
+	}
+	return true, nil
```
tobio: Can replace this with something like

```suggestion
	currentVersion, diags := client.ServerVersion()
	if diags.HasError() {
		return true
	}
	
	return currentVersion.GreaterThanOrEqual(APIKeyWithRestrictionMinVersion)
```

tobio (commented):
Thanks for the addition. Looks like there's something not quite right with the docs and we can re-use the the existing ES version code rather than duplicating it, sorry it wasn't obvious how to access this.

docs/resources/elasticsearch_security_api_key.md
sandermvanvliet-stack: For sure :+1:

internal/elasticsearch/security/api_key.go
sandermvanvliet-stack: Let me give that a go

sandermvanvliet-stack: @tobio I made a slight variation but it now uses the `ServerVersion` call from the client directly instead of making the whole request objects like I had before.

tobio:
Hey @sandermvanvliet-stack are you able to sign the [CLA](https://www.elastic.co/contributor-agreement) so we can continue with this one?

sandermvanvliet-stack:
@tobio I'm still waiting on some internal comms before I do that. Trying to figure out if we have/need/want a corporate CLA for this.

sandermvanvliet-stack:
@tobio the CLA should now be signed as a company wide one by our teams, would you be able to check so we can get on the merge train?

tobio:
> the CLA should now be signed as a company wide one by our teams

I can't see any signed agreements from your company or email domain. I'm not familiar with the company wide agreement process, I'll ask around and see if there's something we need to do.

tobio:
@sandermvanvliet-stack we can't find any recent CLA activity on anyone related to Stack Overflow or Stack Exchange. Are you able to get some more details about how the CLA was signed so we can try and track this down?

sandermvanvliet-stack:
@tobio I was told it was handled but let me prod some people here. Thanks for checking 👍

sandermvanvliet-stack:
@tobio the CLA was signed 26th of June. The contact listed is David Whitt.
I've got a copy here of the signed CLA, let me know if you need one

tobio:
> the CLA was signed 26th of June. The contact listed is David Whitt.

Ok, weird. I can't find anything under that name where we'd expect it to be (our internal tooling or in the Docusign account used for the CLA agreements). Are you able to forward me a copy to toby.brain@elastic.co, or I can send you a secure file sharing link. Once I've got a copy I can try and find out what's happened here. 

At the very least, IIUC we should be able to manually add your accounts with a copy of the agreement.

tobio:
@sandermvanvliet-stack sorry about the delay here. I'm not clear why it took a while for this agreement to show up. It looks like everything's good to go here. 

Thanks for taking the time to add this into the provider, and dealing with the box ticking :)

tobio (approved):

Issue #582, opened 2024-03-06
[Bug] Kibana `ca_certs` connection option not working


unclestacks:
**Describe the bug**
With the latest release, we tried to use the `ca_certs` connection parameter with kibana, but we receive tls connection errors

**To Reproduce**
1. TF configuration used 
```
provider "elasticstack" {
  kibana {
    endpoints = ["https://kibana-stage.company.com"]
    username  = var.admin_user
    password  = var.admin_creds
    ca_certs  = ["/etc/ssl/trust/IssuingCA.pem"]
  }
}
```
2. TF operations to execute to get the error: `terraform apply`
3. See the error in the output:
```
│ Error: Failed to create data view
│
│   with module.elasticstack.elasticstack_kibana_data_view.data_views["default-filebeat"],
│   on ../../../modules/elasticstack/kibana_data_view.tf line 33, in resource "elasticstack_kibana_data_view" "data_views":
│   33: resource "elasticstack_kibana_data_view" "data_views" {
│
│ Post "https://kibana-stage.company.com/s/default/api/data_views/data_view": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

**Expected behavior**
provider should be able to establish a secure connection to Kibana

**Versions (please complete the following information):**
 - OS: Linux
 - Terraform Version: v1.3.7
 - Provider version: v0.11.1
 - Elasticsearch Version: v8.12.1

**Additional context**
I don't believe kibana `ca_certs` is present in the release despite being in the release notes

AdilKhan1117:
Is there, any update for this, also seeing the same issue? Thanks, would be great to have a more stable version for this provider!

tobio:
@AdilKhan1117 are you able to confirm which resources you're seeing this error on?

AdilKhan1117:
> @AdilKhan1117 are you able to confirm which resources you're seeing this error on? 

Thanks 

It's for the below, 


Nested Schema for Kibana:

ca_certs (List of String) A list of paths to CA certificates to validate the certificate presented by the Kibana server.

insecure (Boolean) Disable TLS certificate validation

unclestacks:
We've also seen this issue trying to manage `elasticstack_kibana_slo` resources.

tobio:
@unclestacks @AdilKhan1117 this should be fixed in 0.11.2

PR #585, opened 2024-03-11
Re-use the http client from the kibana client in the openapi generated clients

tobio:
This ensures that logging and tls options are consistent across all Kibana endpoints. This mass of openapi clients is a huge pain, I wonder if it would be nicer to bundle the individual openapi specs together and have a single mass client 🤷 

Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/582

### DIFF
--- a/internal/clients/api_client.go
+++ b/internal/clients/api_client.go
@@ -95,7 +95,9 @@ func NewAcceptanceTestingClient() (*ApiClient, error) {
 		return nil, err
 	}
 
-	actionConnectors, err := buildConnectorsClient(cfg)
+	kibanaHttpClient := kib.Client.GetClient()
+
+	actionConnectors, err := buildConnectorsClient(cfg, kibanaHttpClient)
 	if err != nil {
 		return nil, fmt.Errorf("cannot create Kibana action connectors client: [%w]", err)
 	}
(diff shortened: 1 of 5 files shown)


internal/clients/api_client.go
```diff
@@ -431,30 +435,22 @@ func buildConnectorsClient(cfg config.Client) (*connectors.Client, error) {
 		authInterceptor = connectors.WithRequestEditorFn(basicAuthProvider.Intercept)
 	}
 
-	httpClient := &http.Client{}
```
tobio, line 434: Instead of all the logging related changes I could also just create a new http client, with a debug transport which wraps `httpClient.Transport` which would keeps things more consistent with the current release. 

I'm not a huge fan of that since I feel like it leaves the door open to similar bugs in the future, where we actually do just want to use the http client which was built for the Kibana client.

internal/clients/api_client.go
```diff
@@ -379,39 +381,41 @@ func buildKibanaClient(cfg config.Client) (*kibana.Client, error) {
 	}
 
 	if logging.IsDebugOrHigher() {
-		kib.Client.SetDebug(true)
+		// Don't use kib.Client.SetDebug() here as we re-use the http client within the OpenAPI generated clients
```
dimuon, line 384: Sorry, I don't get it - I guess what we are dong in the following 2 lines has the same result - setting debug mode for Kibana client that is shared between different generated clients (that is totally fine IMO).

tobio: Yep, the logs will look a little different from what they are now, but ultimately we'll be logging the same things (request/response pairs).

dimuon (dismissed):
LGTM

Issue #439, opened 2023-10-04
[Bug] elasticstack_elasticsearch_snapshot_lifecycle: Problem with indicies


Tim-herbie:
**Describe the bug**
When I tried to import the elasticstack_elasticsearch_snapshot_lifecycle resource into my .tfstate, I got the following error message:
```
Error: json: cannot unmarshal string into Go struct field SnapshotPolicyConfig.policy.config.indices of type []string
```

I run a GET against the API (/_slm/policy) and got the following result for the indicies:
```
"config": {
        "include_global_state": true,
        "indices": ".a*,.k*,.r*,.t*,.s*",
        "ignore_unavailable": true
      },
```

**To Reproduce**
Create a Snapshot Lifefycle Policy with the indices above and try to import the resource

**Expected behavior**
No error messages while importing the resource

**Versions (please complete the following information):**
 - OS: Mac
 - Terraform Version: v1.5.7
 - Provider version: 0.8.0
 - Elasticsearch Version: v 8.9.2

Tim-herbie:
It´s still not possible with the newer provider versions 0.9.0 or 0.10.0

PR #593, opened 2024-03-21
Support CSV indices in SLM policy definitions

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/439

Elasticsearch supports both a string array (what the provider has been expecting) and a CSV list for the `indices` field on an SLM policy. This PR customises the JSON marshaller to ensure both formats are parsed into a Go slice.


dimuon (approved):
LGTM 👍 (just changelog)

Issue #417, opened 2023-09-06
[Bug] Panic from provider on empty template block in elasticstack_elasticsearch_index_template resource


nkvoll:
**Describe the bug**

If `template {}` is present in an `elasticstack_elasticsearch_index_template` resource, the provider plugin crashes.

**To Reproduce**

Apply the following resource:

```
    resource "elasticstack_elasticsearch_index_template" "my_index_template" {
        name           = "my_index_template"
        priority       = 500

        index_patterns = [
            "my_index_template-*",
        ]
        template {}
    }
```

**Expected behavior**

Success

** Actual behavior **

```
Terraform will perform the following actions:

  # elasticstack_elasticsearch_index_template.my_index_template will be created
  + resource "elasticstack_elasticsearch_index_template" "my_index_template" {
      + composed_of    = (known after apply)
      + id             = (known after apply)
      + index_patterns = [
          + "my_index_template*",
        ]
      + name           = "my_index_template"
      + priority       = 500

      + template {}
    }

Plan: 1 to add, 0 to change, 0 to destroy.
elasticstack_elasticsearch_index_template.my_index_template: Creating...
╷
│ Error: Plugin did not respond
│
│   with elasticstack_elasticsearch_index_template.my_index_template,
│   on index_templates.tf line 19, in resource "elasticstack_elasticsearch_index_template" "my_index_template":
│   19: resource "elasticstack_elasticsearch_index_template" "my_index_template" {
│
│ The plugin encountered an error, and failed to respond to the
│ plugin.(*GRPCProvider).ApplyResourceChange call. The plugin logs may
│ contain more details.
╵

Stack trace from the terraform-provider-elasticstack_v0.7.0 plugin:

panic: interface conversion: interface {} is nil, not map[string]interface {}

goroutine 58 [running]:
github.com/elastic/terraform-provider-elasticstack/internal/elasticsearch/index.resourceIndexTemplatePut({0xfa1c08, 0x40001ca030}, 0x0?, {0xd69940, 0x40002cae70})
	github.com/elastic/terraform-provider-elasticstack/internal/elasticsearch/index/template.go:262 +0xda4
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).create(0x4000365dc0, {0xfa1c40, 0x40004e9a10}, 0xd?, {0xd69940, 0x40002cae70})
[...]

PR #598, opened 2024-03-27
Don't panic when the template is not defined

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/417

Adds a checked cast to catch the panic when the `template` block is nil. Also removes the duplicate code from the `component_template` resource to fix the same panic there.


dimuon (dismissed):
LGTM (does it make sense to add a test to cover the use case?)

PR #607, opened 2024-04-16
Add alerting connector data source

wandergeek:
This PR adds a datasource that allows users to retrieve alerting connectors.


internal/kibana/connector_data_source.go
```diff
@@ -0,0 +1,30 @@
+package kibana
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSourceConnector() *schema.Resource {
```
tobio, line 12: We need an acceptance test for this data source.

docs/data-sources/kibana_action_connector.md
```diff
@@ -0,0 +1,53 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "elasticstack_kibana_action_connector Data Source - terraform-provider-elasticstack"
+subcategory: ""
```
tobio: You'll need to create a template ([example](https://github.com/elastic/terraform-provider-elasticstack/blob/main/templates/data-sources/kibana_security_role.md.tmpl)) for this data source doc so we can set the subcategory.

internal/clients/kibana/connector.go
```diff
@@ -141,6 +141,67 @@ func GetConnector(ctx context.Context, apiClient *clients.ApiClient, connectorID
...
+			return nil, diag.Errorf("cannot unmarshal connector: %v", err)
+		}
+
+		c, err := connectorResponseToModel(spaceID, respProps)
+		if err != nil {
+			return nil, diag.Errorf("unable to convert response to model: %v", err)
+		}
+		if c.Name == connectorName {
+			foundConnectors = append(foundConnectors, c)
+		}
```
tobio: Why not put the inverse of this at the top of the loop and avoid the wasted conversions, e.g:

```
if connector.Name != connectorName {
    continue
}
```

internal/clients/kibana/connector_test.go
```diff
@@ -130,3 +135,95 @@ func Test_connectorResponseToModel(t *testing.T) {
...
+		  "is_preconfigured": false,
+		  "is_deprecated": false,
+		  "is_missing_secrets": false,
+		  "referenced_by_count": 3
+		},
+		{
+			"id": "d55b6eb0-6bad-11eb-9f3b-611eebc6c3ad",
+			"connector_type_id": ".index",
+			"name": "doubledup-connector",
+			"config": {
```
tobio: Lets fix up the indentation between these two. I guess one's tabbed and the other isn't.

internal/clients/kibana/connector_test.go
```diff
@@ -130,3 +135,95 @@ func Test_connectorResponseToModel(t *testing.T) {
...
+	  ]`
+
+	var requests []*http.Request
+	var mockResponses []string
+	server := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		requests = append(requests, req)
+
+		if len(mockResponses) > 0 {
+			r := []byte(mockResponses[0])
+			// t.Logf("Responding with %s", r)
```
tobio:
```suggestion
```

internal/kibana/connector_data_source.go
```diff
@@ -0,0 +1,30 @@
...
+func dataSourceConnectorRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	connectorName := d.Get("name").(string)
+	if err := d.Set("name", connectorName); err != nil {
+		return diag.FromErr(err)
+	}
+
+	spaceId := d.Get("space_id").(string)
+	if err := d.Set("space_id", spaceId); err != nil {
+		return diag.FromErr(err)
+	}
```
tobio: Why do we need the `Get/Set` dance here, the set bit feels wrong reading through this.

tobio (commented):
Couple of comments, but the main one is the acceptance test.

internal/kibana/connector_data_source.go
wandergeek: Can you suggest an instance of what it should look like? Was just following the pattern from here: https://github.com/wandergeek/terraform-provider-elasticstack/blob/connector-data-source/internal/kibana/role_data_source.go#L163-L169.

internal/clients/kibana/connector.go
wandergeek: I don't have access to the name until I get to the bottom of the loop so I'm not sure how to accomplish this.

tobio: Sorry, I'm not following, what side of the comparison doesn't exist at the top of the loop? The response body includes the name, and `connectorName` is a function argument.

internal/kibana/connector_data_source.go
tobio: That code gets a `name` property and then sets the ID. In this case, you're just getting `name` and then setting `name` on the same resource, as in a noop just writing the same value back to the same place it was read from. 

> of what it should look like?

```
func dataSourceConnectorRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
	return resourceConnectorsRead(ctx, d, meta)
}
```

or even just use `resourceConnectorsRead` [here](https://github.com/elastic/terraform-provider-elasticstack/pull/607/files#diff-c42f369829e0231d33f049572ebaf354e9a1d51d1608531b4b6a2f122d6e7f68R13)

internal/clients/kibana/connector.go
wandergeek: Ah cool, didn't realize the response had it accessible as a field. Fixed.

internal/kibana/connector_data_source.go
wandergeek: Sick one.

docs/data-sources/kibana_action_connector.md
wandergeek: done

internal/kibana/connector_data_source.go
wandergeek: done

internal/kibana/connector_data_source_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+const testAccDataSourceConnector = `
+provider "elasticstack" {
+  elasticsearch {}
+  kibana {}
+}
+
+resource "elasticstack_kibana_action_connector" "slack" {
+	name              = "myconnector"
+	connector_type_id = ".slack"
```
dimuon, line 36: Does it make sense to cover other connector types as well (similar to `internal/kibana/connector_test.go`)?

dimuon: E.g., I guess, we can amend `connector_test.go` to cover the new datasource.

dimuon (dismissed):
LGTM. 
Covering more connector types in acc tests will be nice.

templates/data-sources/kibana_action_connector.tmpl
```diff
@@ -0,0 +1,17 @@
+---
```
tobio: Should be `.md.tmpl`

internal/kibana/connector_data_source.go
```diff
@@ -0,0 +1,57 @@
...
+		},
+		"name": {
+			Description: "The name of the connector. While this name does not have to be unique, a distinctive name can help you identify a connector.",
+			Type:        schema.TypeString,
+			Required:    true,
+		},
+		"connector_type_id": {
+			Description: "The ID of the connector type, e.g. `.index`.",
+			Type:        schema.TypeString,
+			Optional:    true,
```
tobio, line 33: This field should be read-only, we're not actually using it to filter out the connector. 

```suggestion
			Computed:    true,
```

internal/kibana/connector.go
```diff
@@ -198,6 +197,26 @@ func resourceConnectorRead(ctx context.Context, d *schema.ResourceData, meta int
 	return flattenActionConnector(connector, d)
 }
 
+func resourceConnectorsRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
tobio: This looks like it's only used by the data source now. Can we move it into the ds file and rename it to datasourceConnectorRead or something.

internal/kibana/connector.go
```diff
@@ -198,6 +197,26 @@ func resourceConnectorRead(ctx context.Context, d *schema.ResourceData, meta int
...
+		return diags
+	}
+	name := d.Get("name").(string)
+	spaceId := d.Get("space_id").(string)
+
+	connector, diags := kibana.GetConnectorByName(ctx, client, name, spaceId)
+	if diags.HasError() {
+		return diags
+	}
+	id, diags := client.ID(ctx, connector.ConnectorID)
```
tobio: This id should be the connector id and space id like were building on line 142.

tobio (changes requested):

internal/kibana/connector_data_source.go
wandergeek: Changed this so we can search for the connector type now

internal/kibana/connector.go
wandergeek: sure

internal/kibana/connector_data_source.go
dimuon: Does it make sense to search using both name and connector type? IMO `name` should be enough.

internal/clients/kibana/connector.go
```diff
@@ -141,6 +141,74 @@ func GetConnector(ctx context.Context, apiClient *clients.ApiClient, connectorID
...
+
+	if len(foundConnectors) == 1 {
+		return foundConnectors[0], nil
+	}
+
+	if len(foundConnectors) > 1 {
+		return nil, diag.Errorf("error while creating elasticstack_kibana_action_connector datasource: multiple connectors with name [%s/%s] found", spaceID, connectorName)
+	}
+
+	return nil, diag.Errorf("error while creating elasticstack_kibana_action_connector datasource: connector [%s/%s] not found", spaceID, connectorName)
```
tobio: IMO this func should return a slice of matching connectors and these errors should be in the data source. This API client code shouldn't be concerned about data source specifics.

wandergeek: Yeah, good point. Fixing.

internal/kibana/connector_data_source.go
```diff
@@ -0,0 +1,90 @@
...
+	foundConnectors, diags := kibana.SearchConnectors(ctx, client, connectorName, spaceId, connectorType)
+	if diags.HasError() {
+		return diags
+	}
+
+	if len(foundConnectors) == 0 {
+		diag.Errorf("error while creating elasticstack_kibana_action_connector datasource: connector with name [%s/%s] and type [%s] not found", spaceId, connectorName, connectorType)
+	}
+
+	if len(foundConnectors) > 1 {
```
tobio, line 82: I do wonder if it would have been better for this DS to output a list of matching connectors TBH.

tobio (approved):

internal/kibana/connector_data_source.go
wandergeek: For the use case that I want to use this with, I think this would cause confusion. What I'm trying to facilitate here is the ability to grab a connector by name (eg `o11y-alertbot`) and use it in an alert without having to create a new connector in the module. If the DS returns >1 connector then its ambiguous what connector is actually being used, particularly if it has been created erroneously.

tobio: Yea, I understand. You could still achieve that by having a [check](https://developer.hashicorp.com/terraform/language/checks) block do the assertion on a single result, whilst the DS would be more flexible for other consumers.

PR #627, opened 2024-05-13
Add component template data source

lukaspj:
We find that having this data source will make it immensely easier to create Index Templates because we can avoid having to copy-paste the built-in mappings all over the place.


tobio:
Closing this one down. Definitely keen to add this in if you're able to get the CLA signed.

Issue #496, opened 2023-11-30
[Bug] Import of Fleet Agent Policy does not capture and populate  policy_id field


dejongm:
**Describe the bug**
When importing an existing  Fleet Agent Policy into Terraform State, it does not capture the policy_id of the policy. I believe this is causing issues with the `elasticstack_fleet_enrollment_tokens` data object, in that it collects every token for every Agent policy in the query associated with the malformed fleet agent policy resource.

This is an example of an imported fleet_agent_policy resource. **Note the missing `policy_id` field.**

```
# module.fleet_agent_policy["Default Policy"].elasticstack_fleet_agent_policy.main:
resource "elasticstack_fleet_agent_policy" "main" {
    description     = "Default agent policy"
    id              = "7254ece0-8d5c-11ee-8101-5972d5470162"
    monitor_logs    = true
    monitor_metrics = true
    name            = "Default Policy"
    namespace       = "default"
    sys_monitoring  = true
}
```

The following is an example of a fleet policy resource created by Terraform with the policy_id field intact. 
```
# module.fleet_agent_policy["AWS-EKS Pod Policy"].elasticstack_fleet_agent_policy.main:
resource "elasticstack_fleet_agent_policy" "main" {
    description     = "AWS-EKS Pod Policy"
    id              = "8c3e2370-8ec8-11ee-8101-5972d5470162"
    monitor_logs    = true
    monitor_metrics = true
    name            = "AWS-EKS Pod Policy"
    namespace       = "default"
    policy_id       = "8c3e2370-8ec8-11ee-8101-5972d5470162"
    sys_monitoring  = true
}
```

**To Reproduce**
1. terraform import module.fleet_agent_policy[\"Default\ Policy\"].elasticstack_fleet_agent_policy.main <policy_id>

**Expected behavior**
Importing the resource should collect all the relevant fields to the resource. 

**Debug output**
N/A

**Screenshots**
N/A
**Versions (please complete the following information):**
 - OS: Linux
 - Terraform Version v1.3.9
 - Provider version v.0.10.0
 - Elasticsearch Version 8.11

**Additional context**

dejongm:
As a side note,  I just confirmed that manually adding the poilcy_id into the statefile fixes the agent policy token query.

GeorgeGkinis:
We come against this issue also when importing policies with an import block.
While we set the policy_id in our resource, terraform plan says that the policy needs to be destroyed:

```
locals {
[...]

PR #645, opened 2024-05-27
Populate policy_id when importing fleet policies and integrations

GeorgeGkinis:
The policy_id field in the state file now gets populated when importing Fleet integrations and policies based on the ID field of the remote source.

Solves https://github.com/elastic/terraform-provider-elasticstack/issues/496


GeorgeGkinis:
I signed the Contributor Agreement but this does not get reflected here.

GeorgeGkinis:
I will create this PR anew to check is the CLA passes

Issue #647, opened 2024-05-28
[Bug] Issue when creating kibana action webhook connector


JackHurley2:
**Describe the bug**
Issue when creating kibana action webhook connector, in particular the config blocks url and body field.
What I'm attempting to do is create an kibana alerting rule that will trigger when its reached a certain threshold has reached and trigger a connected webhook connector which contains the body contents of relevant fields of the log/document that caused the initial rule to trigger

**To Reproduce**
1. TF configuration used '...'

terraform {
  required_providers {
    elasticstack = {
      source  = "elastic/elasticstack"
      version = "~>0.11.3"
    }
  }
}

3. TF operations to execute to get the error '...' [e.g `terraform plan`,`terraform apply`, `terraform destroy`]
4. See the error in the output '...'

elasticstack_kibana_alerting_rule.example: Modifying... []
╷
│ Error: Plugin did not respond
│
│ The plugin encountered an error, and failed to respond to the plugin6.(*GRPCProvider).ApplyResourceChange call. The plugin logs may contain more details.
╵

Stack trace from the terraform-provider-elasticstack_v0.11.3.exe plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x88 pc=0x19882d4]

goroutine 96 [running]:
github.com/elastic/terraform-provider-elasticstack/internal/clients/kibana.UpdateAlertingRule({0x1ee28c0, 0xc00025f180}, 0xc0004500d0, {{0xc00038b080, 0x24}, {0xc000581908, 0x7}, {0xc00038b050, 0x2f}, {0xc0005818a0, ...}, ...})
        github.com/elastic/terraform-provider-elasticstack/internal/clients/kibana/alerting.go:132 +0x814
github.com/elastic/terraform-provider-elasticstack/internal/kibana.resourceRuleUpdate({0x1ee28c0, 0xc00025f180}, 0xc0002a5e80, {0x1c6aac0, 0xc0004500d0})
        github.com/elastic/terraform-provider-elasticstack/internal/kibana/alerting.go:257 +0x179
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).update(0xc00037d6c0, {0x1ee2818, 0xc00061a150}, 0xc0002a5e80, {0x1c6aac0, 0xc0004500d0})     
[...]

PR #649, opened 2024-05-30
Fix alerting rule update crash for 4xx HTTP errors.

dimuon:
[This](https://github.com/elastic/terraform-provider-elasticstack/pull/633) fixed the problem for `CreateRule`. 
This PR should fix it for `UpdateRule`.
Also it adds unit tests to cover the case.

Rel: https://github.com/elastic/terraform-provider-elasticstack/issues/647


internal/clients/kibana/alerting_api_stub.go
```diff
@@ -0,0 +1,278 @@
...
+
+type FakeAlertingAPI struct {
+	RuleResponseProperties *alerting.RuleResponseProperties
+	HttpResponse           *http.Response
+	Error                  error
+}
+
+// The method stubs are generated initially by the VS Code Quick Fix for the below Blank Identifier definition.
+// https://stackoverflow.com/a/77393824
+var _ alerting.AlertingAPI = (*FakeAlertingAPI)(nil)
```
tobio: Can we use [gomock](https://github.com/uber-go/mock) or similar to generate a proper mock implementation rather than this?

internal/clients/kibana/alerting.go
```diff
@@ -124,18 +130,26 @@ func UpdateAlertingRule(ctx context.Context, apiClient *clients.ApiClient, rule
...
+
 	defer res.Body.Close()
+
 	if diags := utils.CheckHttpError(res, "Unable to update alerting rule"); diags.HasError() {
 		return nil, diags
 	}
 
 	shouldBeEnabled := rule.Enabled != nil && *rule.Enabled
+
 	if shouldBeEnabled && !ruleRes.Enabled {
```
tobio, line 173: We're assuming here that anything where `ruleRes == nil` will be caught by `utils.CheckHttpError` above? I wonder if it's worth having an explicit check as well?

dimuon: > We're assuming here that anything where ruleRes == nil will be caught by utils.CheckHttpError above?

I'm not sure actually - `CheckHttpError` just checks HTTP status code.

dimuon: I've added logic to explicitly handle an empty response with HTTP 200.

dimuon: Sorry - the new logic just returns an error in case of an empty response for Create/Update call, HTTP status code doesn't make any diff in this case. 
Still in reality, `CheckHttpError` should handle the cases when something goes wrong based on the HTTP status code. The new logic just makes sure that crash won't happen.

dimuon:
@tobio , I addressed the comments, could you take a look at it again when you have time?

tobio (approved):
LGTM 🎉

PR #653, opened 2024-06-04
Fix examples to support elastic/ec from v0.6.0

aaharu:


tobio (approved):
Ooof, thanks for fixing this up!

Are you able to sign the CLA so we can get this one merged?

aaharu:
I have signed, but still failing.

tobio:
Ah right, these docs are generated. If you update [this](https://github.com/elastic/terraform-provider-elasticstack/blob/main/examples/cloud/deployment.tf) example file as well everything should be happy (`make docs-generate` should not result in a dirty git state).

I can take a look at fixing this up shortly if you don't get to it.

tobio (approved):

PR #663, opened 2024-06-12
Fixup data views in 8.14

tobio:
Includes the following changes:
* Makes the `title` required. It was being treated as required in the code, and really didn't work if it wasn't defined so in reality no-one is using the resource without setting the title now. 
* More correctly maps the TF plan to the API body. Fields which are unknown in the plan shouldn't be specified in the body, however they were being incorrectly set to the Go default (e.g `""` for strings). This causes the resource to fail when the `id` isn't set in 8.14


dimuon (dismissed):
👍

Issue #610, opened 2024-04-19
[Feature] Support Synthetics APIs as Terraform resources


BenB196:
**Is your feature request related to a problem? Please describe.**

I would like to be able to programmatically manage Elastic Synthetics via Terraform.

**Describe the resource you would like to have implemented.**

https://github.com/elastic/kibana/issues/169547 recently introduced public APIs for these resources.

**Describe the solution you'd like**

This Terraform provider to be able to manage Synthetics resources.

**Describe alternatives you've considered**

Hacking something together via https://registry.terraform.io/providers/Mastercard/restapi/latest/docs

**Additional context**

https://github.com/elastic/cloud-on-k8s/issues/7724

russellproud:
Coming here to add some support to this request.

Utilising the mastercard/restapi provider works for create and update, but is unable to manage the delete lifecycle as a result of  the DELETE endpoint for Synthetics requires the Id to be included in the request body, not the uri. 

@BenB196 have you found a method to achieve the delete via mastercard/restapi?

tobio:
Duplicates https://github.com/elastic/terraform-provider-elasticstack/issues/335

doom160:
@russellproud 
I used this code (quite an amount of transformation) for elastic cloud 8.12.2, so far it works for most of the monitors, i think this endpoint is different for new version
```terraform
resource "restapi_object" "main" {
  for_each  = { for idx, synthetic in local.synthetic_request_body : synthetic.name => synthetic }
  path      = "/internal/synthetics/service/monitors"
  read_path = "/internal/synthetics/service/monitor/{id}?decrypted=true"
  data      = jsonencode(each.value)
}

locals {
  synthetic_request_body = [
    for synthetic_url in var.synthetic_monitors : merge({
      name : (var.prefix_name == "") ? "${synthetic_url.url}-${synthetic_url.env}" : "${var.prefix_name}-${synthetic_url.url}-${synthetic_url.env}",
      urls : synthetic_url.url
      type : can(synthetic_url.form_monitor_type) ? ((synthetic_url.form_monitor_type == "single" || synthetic_url.form_monitor_type == "multistep") ? "browser" : synthetic_url.form_monitor_type) : "http",
      form_monitor_type : can(synthetic_url.form_monitor_type) ? synthetic_url.form_monitor_type : "http",

      enabled : can(synthetic_url.enable) ? synthetic_url.enable : true,
      alert : {
        status : {
[...]

PR #692, opened 2024-07-31
feat(synthetics): add synthetics monitor and private location API support to go-kibana-rest

biscout42:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/610

This PR adds support for synthetic `http` Monitor and PrivateLocation APIs. 

Main doc: https://www.elastic.co/guide/en/kibana/current/synthetics-apis.html

Notes: 
1. new `uuid` dependency for testing purposes
2. proposed way to model monitor API. Different monitor kinds has different set of parameters. 
3. updated to `8.14.3`
4. added kibana config for docker compose

### DIFF
--- a/libs/go-kibana-rest/docker-compose.yml
+++ b/libs/go-kibana-rest/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '2.3'
 services:
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:8.14.1
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.14.3
     environment:
       cluster.name: test
       discovery.type: single-node
@@ -23,11 +23,14 @@ services:
       elasticsearch:
         condition: service_started
   kibana:
-    image: docker.elastic.co/kibana/kibana:8.14.1
+    image: docker.elastic.co/kibana/kibana:8.14.3
+    volumes:
+      - ./kibana.yml:/usr/share/kibana/config/kibana.yml
     environment:
       ELASTICSEARCH_HOSTS: http://es:9200
       ELASTICSEARCH_USERNAME: kibana_system
       ELASTICSEARCH_PASSWORD: changeme
+      xpack.security.http.ssl.enabled: false
       xpack.license.self_generated.type: trial
     links:
       - elasticsearch:es
--- a/libs/go-kibana-rest/docker-compose.yml
+++ b/libs/go-kibana-rest/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     ports:
       - "9200:9200/tcp"
   set-kibana-password:
-    image: docker.elastic.co/kibana/kibana:8.14.1
+    image: docker.elastic.co/kibana/kibana:8.14.3
     restart: on-failure
     links:
       - elasticsearch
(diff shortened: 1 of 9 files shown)


libs/go-kibana-rest/kbapi/api.kibana_synthetics.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+type KibanaSyntheticsPrivateLocationAPI struct {
+	Create KibanaSyntheticsPrivateLocationCreate
+	Delete KibanaSyntheticsPrivateLocationDelete
+	Get    KibanaSyntheticsPrivateLocationGet
+}
+
+type MonitorAlertConfig struct {
+	Status struct {
```
tobio: Do we ever need to create these structs in code or will they only be used for serialisation? I'm not opposed to the anonymous struct, but if we expect to reference them in code it might be easier to define them as their own type.

libs/go-kibana-rest/kbapi/api_test.go
```diff
@@ -73,7 +75,7 @@ func (s *KBAPITestSuite) SetupSuite() {
 
 func (s *KBAPITestSuite) SetupTest() {
 
-	// Do somethink before each test
+	// Do something before each test
```
tobio, line 78: 👏 best change in this PR.

libs/go-kibana-rest/docker-compose.yml
```diff
@@ -23,11 +23,14 @@ services:
       elasticsearch:
         condition: service_started
   kibana:
-    image: docker.elastic.co/kibana/kibana:8.14.1
+    image: docker.elastic.co/kibana/kibana:8.14.3
+    volumes:
+      - ./kibana.yml:/usr/share/kibana/config/kibana.yml
```
tobio: IIUC everything in this config file could be supplied through the env, IMO it would be nicer to keep all the config here if possible.

tobio (approved):
LGTM, couple of small comments.

libs/go-kibana-rest/kbapi/api_test.go
biscout42: I'm sorry, couldn't resist to fix typo. Please, let me know if I should not do it 🙈

libs/go-kibana-rest/docker-compose.yml
biscout42: Good idea, I have added `XPACK_ENCRYPTEDSAVEDOBJECTS_ENCRYPTIONKEY` and removed this yaml.

libs/go-kibana-rest/kbapi/api.kibana_synthetics.go
biscout42: That is a good question. tbh: I don't know yet, how will customer use this API. I expect to refactor this part after adding terraform changes. 

For scope of this PR: I have moved nested types to separate structures.

libs/go-kibana-rest/kbapi/api_test.go
biscout42: Okey, googled it and reverted.

tobio: > I'm sorry, couldn't resist to fix typo. Please, let me know if I should not do it 🙈

You did absolutely nothing wrong, sorry for the confusion. I'm 100% on board with fixing this typo, I assume we inherited it when this code was vendored in here.

tobio (dismissed):
LGTM

tobio (approved):

Issue #238, opened 2023-01-05
[Bug] Cannot create index with number_of_replicas=0


paivin-dn:
**Describe the bug**
I've tried to create an index with the parameter `number_of_replicas=0`, but in an elasticsearch instance, I've got an index with `number_of_replicas=1`.

**To Reproduce**
1. Define index in a terraform code:
```hcl
resource "elasticstack_elasticsearch_index" "example" {
  name               = "example"
  number_of_shards   = 1
  number_of_replicas = 0
}
```
2. Run the terraform plan & terraform plan commands
3. Check index settings and make sure index.number_of_replicas = 1.

**Expected behavior**
The created index should have  `index.number_of_replicas = 0` instead of `index.number_of_replicas = 0`

**Versions (please complete the following information):**
 - OS: macOS
 - Terraform Version: v1.1.7
 - Provider version: 0.5.0
 - Elasticsearch Version: 8.3.2

k-yomo:
The root cause would be that `if raw, ok := d.GetOk(tfFieldKey); ok { ` will only be true when the value is not zero value.
It seems that's [known issue](https://discuss.hashicorp.com/t/check-if-optional-attribute-is-set-in-tf-file/8735/5) and can be fixed by migrating to new sdk(terraform-provider-framework) since they have `IsNull` method to check if not set, or is explicitly set to null.
https://github.com/elastic/terraform-provider-elasticstack/blob/c5e409539eeec2ed46d38eb749ff8d760a9d87d3/internal/utils/utils.go#L148-L163

Workaround would be 
1. create with value > 0 and update `number_of_replicas` to 0 after.
```
resource "elasticstack_elasticsearch_index" "example" {
  name               = "example"
  number_of_shards   = 1
  number_of_replicas = 1
}
```
2. use settings field (it's now deprecated though)
```
resource "elasticstack_elasticsearch_index" "example" {
  name               = "example"
  number_of_shards   = 1

  settings {
    setting {
      name  = "number_of_replicas"
      value = "0"
    }
  }
}
```

paivin-dn:
[...]

PR #698, opened 2024-08-06
Migrate index resource to plugin framework

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/238

Partly meant as an evaluation of effort in doing these migrations. Partly solving actual customer issues. This PR migrates the existing SDK based index resource to a plugin framework based implementation. 

There are no schema changes introduced in the migration, and the change is intended to be fully BWC with current definitions. 

It adds an acceptance test explicitly verifying the migration behaviour (i.e creating an index with an SDK based provider version before verifying that the index continues to work with the latest code).  

Needs a changelog entry. I can also split this one up more, it's been a thing.


go.mod
```diff
@@ -8,10 +8,12 @@ require (
 	github.com/deepmap/oapi-codegen/v2 v2.2.0
 	github.com/disaster37/go-kibana-rest/v8 v8.5.0
 	github.com/elastic/go-elasticsearch/v7 v7.17.10
+	github.com/google/gofuzz v1.2.0
 	github.com/google/uuid v1.6.0
 	github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320
 	github.com/hashicorp/go-version v1.7.0
 	github.com/hashicorp/terraform-plugin-framework v1.10.0
+	github.com/hashicorp/terraform-plugin-framework-jsontypes v0.1.0
```
biscout42: 👍

internal/clients/elasticsearch/index.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
 	"github.com/elastic/terraform-provider-elasticstack/internal/models"
 	"github.com/elastic/terraform-provider-elasticstack/internal/utils"
+	fwdiags "github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
```
dimuon, line 16: nit-pick: It seems the same file contains functions for both ILM (still SDKv2) and index (Framework) resources. Does it make sense to split them?

tobio: This file has functions for:

* ILM
* Indices
* Templates
* Component templates
* Ingest processing
...

I'm happy to split it up, but I don't think it should be an SDK vs FW thing. WDYT?

dimuon: I mean that having one file per resource (e.g. indices) would allow us not to mix both SDK and FW during migration, FWIW (and besides other benefits). However, if we cannot split it like that then obviously there is no sense to do it based on SDK vs FW thing.

internal/utils/schema.go
```diff
@@ -13,6 +13,6 @@ func ExpandStringSet(set *schema.Set) []string {
 	return strs
 }
 
-func IsKnown(val attr.Value) bool {
-	return !(val.IsNull() || val.IsUnknown())
+func IsKnown[T attr.Value](value T) bool {
```
dimuon: Just curios - what are the benefits of converting the function to generic in this case? Is it just about performance?

internal/elasticsearch/index/index/mapping_modifier.go
```diff
@@ -0,0 +1,110 @@
...
+		return
+	}
+
+	stateStr := req.StateValue.ValueString()
+	cfgStr := req.ConfigValue.ValueString()
+
+	var stateMappings map[string]interface{}
+	var cfgMappings map[string]interface{}
+
+	// No error checking, schema validation ensures this is valid json
```
dimuon, line 33: 👍 thanks for commenting this

internal/utils/schema.go
tobio: I don't think there is one. IIRC this function was merged into main after I'd started work on this PR. I'm happy to revert this change if you think that makes sense?

internal/elasticsearch/index/index/models.go
```diff
@@ -0,0 +1,642 @@
...
+		return basetypes.SetValue{}, diags
+	}
+
+	return modelAliases, nil
+}
+
+func setSettingsFromAPI(ctx context.Context, model *tfModel, apiModel models.Index) diag.Diagnostics {
+	modelType := reflect.TypeOf(*model)
+
+	for _, key := range dynamicSettingsKeys {
```
dimuon, line 228: Out of curiosity: did you consider using struct instead of map of interfaces for `Index.settings`? The below code resembles low-level code from the FW. I understand that ES client doesn't provide a model for index but what if we define it instead? Then we can rely on golang types and conversions from the FW.

internal/utils/customtypes/duration_type_test.go
```diff
@@ -0,0 +1,142 @@
...
+			} else {
+				require.Nil(t, err)
+			}
+
+			require.Equal(t, tt.expectedValue, val)
+		})
+	}
+}
+
+func TestDuration_StringSemanticEquals(t *testing.T) {
```
dimuon: nit-pick: the function is defined in `duration_value.go`,  shall we move it to `duration_value_test.go`?

internal/utils/schema.go
dimuon: Up to you. I don't have any preference. Having said this I came across an [investigation](https://blog.stackademic.com/golang-design-generics-vs-interfaces-how-it-really-works-under-the-hood-c1e5057a7ae6) regarding the topic, still IMO the perf effect if any doesn't make any diff in our case.

internal/elasticsearch/index/index/models.go
tobio: I didn't, but probably because I was keeping this analogous to the tf -> api conversion logic. It's an interesting point though.

I had a quick look at this, and I'm concerned that there's a bunch of inconsistencies around how ES handles settings which make this tricky. For example, `number_of_shards` is defined in the schema as an int64, we pass it to ES as an int, but it gets returned as a string which then breaks decoding (since it's an int in the schema). 

I'm sure there's some logic that we could build to make this stuff work, but I'm not sure it will be much nicer than whats in this PR. I can dig in here more if you think it's worthwhile, but I'm not convinced it's worth the time.

dimuon: I trust your judgement. Let's keep it as is.

dimuon (approved):

PR #699, opened 2024-08-06
Add resource for kibana synthetics monitors: http and tcp

biscout42:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/610 

TODOs

- [x] Implement CRUD
- [x] Implement `tfModelV0` and unit test for API <-> tf model conversion
- [x] Implement `tcp` in kibana API client
- [x] Add acc tests
- [x] Add resource documentation, sample templates and readme entry


libs/go-kibana-rest/kibana.go
```diff
@@ -35,6 +35,7 @@ func NewClient(cfg Config) (*Client, error) {
 	}
 
 	restyClient := resty.New().
+		SetDebug(true).
```
biscout42: @biscout42 , TODO

.github/workflows/test.yml
```diff
@@ -47,6 +47,12 @@ jobs:
       - name: Lint
         run: make lint
 
+      - name: Check env
```
biscout42: @biscout42 , TODO

.github/workflows/test.yml
```diff
@@ -78,30 +86,32 @@ jobs:
           ELASTICSEARCH_USERNAME: ${{ env.KIBANA_SYSTEM_USERNAME }}
           ELASTICSEARCH_PASSWORD: ${{ env.KIBANA_SYSTEM_PASSWORD }}
           XPACK_ENCRYPTEDSAVEDOBJECTS_ENCRYPTIONKEY: a7a6311933d3503b89bc2dbc36572c33a6c10925682e591bffcab6911c06786d
-          xpack.fleet.enabled: true
+          LOGGING_ROOT_LEVEL: debug
+          XPACK_FLEET_ENABLED: true
```
biscout42: @biscout42 , TODO - remove. It is not supported as env var config

.github/workflows/test.yml
```diff
@@ -78,30 +86,32 @@ jobs:
...
         ports:
           - 5601:5601
+        options: --health-cmd="curl http://localhost:5601/api/status" --health-interval=10s --health-timeout=5s --health-retries=10
```
biscout42, line 86: @biscout42 , TODO - check if works with all stack versions

Makefile
```diff
@@ -132,7 +132,8 @@ docker-kibana: docker-network docker-elasticsearch set-kibana-password ## Start
 		-e ELASTICSEARCH_USERNAME=$(KIBANA_SYSTEM_USERNAME) \
 		-e ELASTICSEARCH_PASSWORD=$(KIBANA_SYSTEM_PASSWORD) \
 		-e XPACK_ENCRYPTEDSAVEDOBJECTS_ENCRYPTIONKEY=a7a6311933d3503b89bc2dbc36572c33a6c10925682e591bffcab6911c06786d \
-		-e "logging.root.level=debug" \
+		-e "xpack.fleet.enabled=true" \
```
biscout42: @biscout42 , TODO - remove and add kibana docker log env var

Makefile
```diff
@@ -132,7 +132,8 @@ docker-kibana: docker-network docker-elasticsearch set-kibana-password ## Start
...
+		-v ./kibana-config/:/usr/share/kibana/config/ \
```
biscout42: TODO: remove

libs/go-kibana-rest/docker-compose.yml
```diff
@@ -25,16 +25,16 @@ services:
...
     image: docker.elastic.co/kibana/kibana:8.14.3
     environment:
+      SERVER_NAME: kibana
       ELASTICSEARCH_HOSTS: http://es:9200
       ELASTICSEARCH_USERNAME: kibana_system
       ELASTICSEARCH_PASSWORD: changeme
-      xpack.security.http.ssl.enabled: false
-      xpack.license.self_generated.type: trial
       XPACK_ENCRYPTEDSAVEDOBJECTS_ENCRYPTIONKEY: min-32-byte-long-strong-encryption-key
+      xpack.fleet.enabled: true
```
biscout42: TODO: remove

internal/kibana/synthetics/acc_test.go
```diff
@@ -0,0 +1,318 @@
...
+  	elasticsearch {}
+	kibana {}
+	fleet{}
+}
+`
+
+	privateLocationConfig = `
+
+resource "elasticstack_fleet_agent_policy" "test" {
+	name            = "TestMonitorResource Agent Policy - test"
```
tobio: It would be good to add a random suffix to these names to avoid any conflicts during parallel tests, or prior failed tests which failed to get cleaned up properly.

internal/kibana/synthetics/acc_test.go
```diff
@@ -0,0 +1,318 @@
...
+		ssl_supported_protocols = ["TLSv1.2"]
+		proxy_url = "http://localhost"
+		proxy_use_local_resolver = false
+	}
+}
+
+`
+)
+
+func TestSyntheticMonitorResource(t *testing.T) {
```
tobio: IMO it would be preferable to split this into a test for each TCP and HTTP monitors.

internal/kibana/synthetics/acc_test.go
```diff
@@ -0,0 +1,318 @@
...
+	agent_policy_id = elasticstack_fleet_agent_policy.test.policy_id
+}
+
+`
+
+	httpMonitorConfig = `
+
+resource "elasticstack_kibana_synthetics_monitor" "http-monitor" {
+	name = "TestHttpMonitorResource"
+	space_id = "default"
```
tobio: Are you able to verify the resource behaviour (specifically imports) when the monitor is in a non-default namespace? 

In other [resources](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/kibana/data_view/schema.go#L284) we encode both an ID (whatever that means for the specific resource) and space into the TF resource ID attribute. Originally this was to support importing since that (at least used to) only have access to the ID. The [docs](https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-resource-import#implement-import-functionality) don't seem to suggest there's a change here.

internal/kibana/synthetics/schema.go
```diff
@@ -1,15 +1,299 @@
...
+			"ssl_verification_mode": schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "Controls the verification of server certificates. ",
+			},
+			"ssl_supported_protocols": schema.ListAttribute{
+				ElementType:         types.StringType,
+				Optional:            true,
+				MarkdownDescription: "List of allowed SSL/TLS versions.",
+			},
+			"max_redirects": schema.StringAttribute{ //TODO: make int64??
```
tobio: Can this be int64 now? If not can we add the reasoning behind this being a string as a comment?

internal/kibana/synthetics/schema.go
```diff
@@ -1,15 +1,299 @@
...
+	MaxRedirects          types.String   `tfsdk:"max_redirects"`
+	Mode                  types.String   `tfsdk:"mode"`
+	IPv4                  types.Bool     `tfsdk:"ipv4"`
+	IPv6                  types.Bool     `tfsdk:"ipv6"`
+	ProxyURL              types.String   `tfsdk:"proxy_url"`
+	// commented out due-to https://github.com/elastic/kibana/issues/189906
+	// it leads to terraform  Error: Provider produced inconsistent result after apply
+	//ProxyHeader           jsontypes.Normalized `tfsdk:"proxy_header"`
+	//Username              types.String         `tfsdk:"username"`
+	//Password              types.String         `tfsdk:"password"`
```
tobio: These three in particular feel fairly important to support. IIUC these fields are just not present in a GET. IMO it would be preferable to support these attributes during creation only, than not supporting them at all. To be clear, I'm suggesting we let the user configure them, Terraform just won't detect drift from the desired state since the current values are not available during the read cycle.

biscout42: > to be clear, I'm suggesting we let the user configure them, Terraform just won't detect drift from the desired state since the current values are not available during the read cycle.

Terraform detects drift because PUT / POST returns this values back. Do you suggest to remove them from PUT / POST response mapping as well?

tobio: > Terraform detects drift because PUT / POST returns this values back.

IIUC the current workflow is something like (assuming these fields still existed):
1. TF creates the monitor. We set state based on the response (including user/pass)
2. TF refresh reads the existing monitor. These fields aren't included in the response and so they're updated to be empty
3. TF apply notices empty != configured and causes and updates the monitor
4. Update sets the fields in state again (since they're returned in the response)
5. Loop back to 2.

Is that right, or is there something else about PUT/POST I'm missing here?

There's a couple of changes we can make here:
* Most (all?) of the other resources discard the PUT/POST responses tbh. Instead they get the state model via the read. In some cases there's technical reasons for that, in other cases it's mostly to have a consistent read path. 
* Instead of building an entirely new state model from the response (like [here](https://github.com/elastic/terraform-provider-elasticstack/pull/699/files#diff-1cfb31628c813a5448fbf4e10535e71f3233bcb07692b41eb7316f67c54613aaR42), we update the existing state, preserving the user/pass fields if they're unset in the response. 

WDYT?

biscout42: > Instead they get the state model via the read.

not sure, if that will work (I don't know how TF will handle that fields become empty). Currently, if `TF apply notices empty != configured` it fails the test. If I allow state be based on read, it would eventually set not-returned fields to empty, but they will stay configured on the server, because of [partial update](https://www.elastic.co/guide/en/kibana/current/update-monitor-api.html#_partial_updates) API behaviour. 

> we update the existing state, preserving the user/pass fields if they're unset in the response.

so merge. That might be a good idea, let me try, if I could implement it.

internal/kibana/synthetics/acc_test.go
biscout42: I see what you mean. Let me try to test API, if get by ID without namespace works for not `default` namespace.

biscout42: Interesting, that [import test](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/kibana/synthetics/private_location/acc_test.go#L66) for `private_location` worked with `testacc`  namespace.

Does it answer on your question for `private_location` resource import or shall I test something else?

biscout42: Okey, tested APIs - private location works, but monitor. 

This works
```
POST  /s/testacc/api/synthetics/private_locations # creates a private location with id 97a1d865-5d52-4e52-838f-9af6a186d97e
GET  /api/synthetics/private_locations/97a1d865-5d52-4e52-838f-9af6a186d97e # return correct result

```

This doesn't work:
```
POST  /s/testacc/api/synthetics/monitors #creates a monitor with id 510bedd2-58a9-47aa-921b-bb9c2bc4ff34
GET  /api/synthetics/monitors/510bedd2-58a9-47aa-921b-bb9c2bc4ff34 # 404
```

Good catch, @tobio .

Last question from my side: is the import acceptance test, as in the example above enough or shall I test import somehow else?

internal/kibana/synthetics/schema.go
biscout42: Yep, will do. API returns it as a string, although documents it as a number. I'll do conversion.

biscout42: done.

internal/kibana/synthetics/acc_test.go
biscout42: IMHO import test with `testacc` namespace is good enough. I was able to reproduce the issue moving http monitor to `testacc` 

```
Step 2/6 error running import: exit status 1

        Error: Failed to get monitor `eeae1afb-014e-475c-a826-799f829e9fe1`, namespace

        Monitor id eeae1afb-014e-475c-a826-799f829e9fe1 not found!%!(EXTRA string=Not
        Found)
```

@tobio , let me know if I need to add different way to verify imports.

biscout42: Done. I have migrated private location to composite id as well.

internal/kibana/synthetics/schema.go
biscout42: Done. Please, double check the update.

internal/kibana/synthetics/acc_test.go
biscout42: should be done

internal/kibana/synthetics/acc_test.go
biscout42: done

internal/kibana/synthetics/private_location/read.go
```diff
@@ -25,17 +23,22 @@ func (r *Resource) Read(ctx context.Context, request resource.ReadRequest, respo
 		return
 	}
 
-	label := state.Label.ValueString()
-	namespace := state.SpaceID.ValueString()
-	result, err := kibanaClient.KibanaSynthetics.PrivateLocation.Get(label, namespace)
+	compositeId, dg := synthetics.GetCompositeId(state.ID.ValueString())
```
tobio: This resource was released as part of 0.11.5, so it's possible customers will have existing resources using the original schema, and attempting to get the composite ID from the old format will fail (it doesn't contain a `/`). 

You could handle it here (i.e use the `Label`/`SpaceID` path if `ID` doesn't contain `/`) or implement a [state upgrade](https://developer.hashicorp.com/terraform/plugin/framework/resources/state-upgrade) which would mean duplicating the schema, but is more explicit.

tobio (commented):
LGTM, we're going to need to deal with the state transition in the private_location resource unfortunately.

tobio:
We'll also need a CHANGELOG entry too

internal/kibana/synthetics/private_location/read.go
biscout42: I though `0.11.5` release failed and we could get this change in due-to this: https://buildkite.com/elastic/terraform-provider-elasticstack-release/builds/26

Have I misunderstood it?

biscout42: Done

biscout42: I have implemented the change, but imho it is not necessary. 0.11.5 has import documentation for private locations based on ID, but implementation based on `label`. IMHO it is bug and would not work at all. 

What I try to say, I assume there is no working import functionality for private locations.

tobio: It's not just imports though, a private location created by 0.11.5 would have an ID in the current format, which would fail to be parsed by the code in this PR.

biscout42: Only for delete, right? There is no update, create doesn't parse ID and I have changed read. And IIUC new code provides new ID during the read, which should replace resource according to schema. Please, correct me, if I miss something. Will fix delete soon.

biscout42: @tobio , I updated delete. Please, advice, what else do I miss.

tobio: Thanks, read/delete should be enough. Anything else will first read the resource to get the current state so the ID should be updated to the new format before anything else is called anyway.

tobio (dismissed):
LGTM

biscout42:
@tobio , could you ack it again 🙏 ? I have split resourced in monitor test to prevent cross-resource deletion.

biscout42:
FYI: I enabled auto-merge.

tobio (approved):

PR #712, opened 2024-08-15
Use ephemeral github token for build.

gigerdo:
The `VAULT_GITHUB_TOKEN` env var contains the token with permissions to create releases. Goreleaser on the other hand will only use `GITHUB_TOKEN` for the release.

### DIFF
--- a/.buildkite/scripts/release.sh
+++ b/.buildkite/scripts/release.sh
@@ -12,4 +12,5 @@ echo "--- Caching GPG passphrase"
 echo "$GPG_PASSPHRASE_SECRET" | gpg --armor --detach-sign --passphrase-fd 0 --pinentry-mode loopback
 
 echo "--- Release the binaries"
+export GITHUB_TOKEN="${VAULT_GITHUB_TOKEN}"
 make release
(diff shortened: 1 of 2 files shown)


.buildkite/scripts/release.sh
```diff
@@ -12,4 +12,5 @@ echo "--- Caching GPG passphrase"
 echo "$GPG_PASSPHRASE_SECRET" | gpg --armor --detach-sign --passphrase-fd 0 --pinentry-mode loopback
 
 echo "--- Release the binaries"
+export GITHUB_TOKEN="${VAULT_GITHUB_TOKEN}"
```
tobio: I think it would be worth only doing this if `VAULT_GITHUB_TOKEN` is set so this script still works locally.

tobio (approved):

.buildkite/scripts/release.sh
gigerdo: The one problem is that I suspect `GITHUB_TOKEN` may be set as well in buildkite. Otherwise the goreleaser would have complained about the token not being set.

gigerdo: We could use the logic from the pre-command script:

```
if [[ "$BUILDKITE_PIPELINE_SLUG" == "terraform-provider-elasticstack-release" ]]; then
  export GITHUB_TOKEN="${VAULT_GITHUB_TOKEN}"
fi
```

gigerdo: Actually I noticed now that `GITHUB_TOKEN` is set in `pre-command` as well. So I can just replace it there.

tobio (approved):

PR #715, opened 2024-08-19
 Add support for the `alert_delay` param in the Create Rule API

adcoelho:
related to [kibana/#186963](https://github.com/elastic/kibana/issues/186963)

## Summary

I am updating the Terraform Provider to use the latest version of the Create Rule API.

This is the first of a few PRs in which I intend to add support for new parameters in this API.

This PR adds support for `alert_delay`.

I am still testing the changes manually but updated the tests and they are all passing. I think the review can start.

## What was done

I am not super familiar with this code so I'll share my steps and maybe you guys could doublecheck to see if I forgot something?

1. Generated a new `bundled.yaml` following [these instructions](https://github.com/elastic/kibana//tree/8.14/x-pack/plugins/alerting/docs/openapi).
2. Moved `bundled.yaml` to `generated/alerting/bundled.yaml`
3. `make generate-alerting-client`
4. Fixed errors in alerting.go etc. Not much changed in the types so I think there were no BWC issues. The only relevant thing was that some fields seem to have become optional.
5. Re-generate mocks and fixed the tests
6. Updated the documentation.

PS: Am I missing something in `resourceRuleRead` in `internal/kibana/alerting.go`?


tobio:
It looks like you need to add some version checks to make sure `alert_delay` is only included in the request when it's supported.

adcoelho:
@tobio Thanks for your comment! I added one now to ignore the alert delay field if the version doesn't support it but I see in other places where we do version checks an error is thrown.

Is ignoring the field a suitable solution?

tobio:
> but I see in other places where we do version checks an error is thrown.

We should be consistent with the other resources and throw an error if `alert_delay` is set but not supported by the server.

internal/kibana/alerting.go
```diff
@@ -181,6 +189,22 @@ func getAlertingRuleFromResourceData(d *schema.ResourceData) (models.AlertingRul
...
 	}
 
+	if v, ok := d.GetOk("notify_when"); ok {
+		t := v.(string)
+		rule.NotifyWhen = &t
+	}
+
+	if v, ok := d.GetOk("alert_delay"); ok {
+		if serverVersion.LessThanOrEqual(alertDelayMinSupportedVersion) {
+			return models.AlertingRule{}, diag.FromErr(fmt.Errorf("'alert_delay' field is supported only for elasticsearch v8.13.x"))
```
dimuon:
```suggestion
			return models.AlertingRule{}, diag.FromErr(fmt.Errorf("'alert_delay' field is supported only starting from elasticsearch v8.13"))
```

dimuon:
@adcoelho, please add changelog similar to https://github.com/elastic/terraform-provider-elasticstack/pull/719/files

internal/kibana/alerting.go
```diff
@@ -181,6 +189,22 @@ func getAlertingRuleFromResourceData(d *schema.ResourceData) (models.AlertingRul
 		rule.Throttle = &t
 	}
 
+	if v, ok := d.GetOk("notify_when"); ok {
+		t := v.(string)
+		rule.NotifyWhen = &t
```
dimuon:
```suggestion
		rule.NotifyWhen = utils.Pointer(v.(string))
```

internal/kibana/alerting.go
```diff
@@ -181,6 +189,22 @@ func getAlertingRuleFromResourceData(d *schema.ResourceData) (models.AlertingRul
...
+
+	if v, ok := d.GetOk("alert_delay"); ok {
+		if serverVersion.LessThanOrEqual(alertDelayMinSupportedVersion) {
+			return models.AlertingRule{}, diag.FromErr(fmt.Errorf("'alert_delay' field is supported only for elasticsearch v8.13.x"))
+
+		}
+		t := v.(float64)
+		rule.AlertDelay = &models.AlertingRuleAlertDelay{
+			Active: (float32)(t),
+		}
```
dimuon:
```suggestion
		rule.AlertDelay = &models.AlertingRuleAlertDelay{
			Active: v.(float32),
		}
```

internal/models/alert_rule.go
```diff
@@ -18,19 +18,31 @@ type AlertingRule struct {
 
 	ScheduledTaskID *string
 	ExecutionStatus AlertingRuleExecutionStatus
+	AlertDelay      *AlertingRuleAlertDelay
```
dimuon: Why do we need a pointer to struct? Can we just use something like `AlertDelay *float32`?

internal/models/alert_rule.go
```diff
@@ -18,19 +18,31 @@ type AlertingRule struct {
...
+}
+
 type AlertingRuleAction struct {
-	Group  string
-	ID     string
-	Params map[string]interface{}
+	Group     string
+	ID        string
+	Params    map[string]interface{}
+	Frequency AlertingRuleActionFrequency
```
dimuon: It seems we don't use it.

dimuon:
@adcoelho 
> PS: Am I missing something in `resourceRuleRead` in `internal/kibana/alerting.go`?

We need to read `alert_delay` from API similar to other fields. To check this in an IT test, you may want to add a test for resource import, e.g. refer to internal/elasticsearch/cluster/slm_test.go#L47 (look for `ImportState:`), and check that `alert_delay` contains the required value.

dimuon:
> @adcoelho
> 
> > PS: Am I missing something in `resourceRuleRead` in `internal/kibana/alerting.go`?
> 
> We need to read `alert_delay` from API similar to other fields. To check this in an IT test, you may want to add a test for resource import, e.g. refer to internal/elasticsearch/cluster/slm_test.go#L47 (look for `ImportState:`), and check that `alert_delay` contains the required value.

Please disregard this (while import test can still be handy). `alert_delay` is read in `ruleResponseToModel`.

internal/kibana/alerting.go
tobio:
```suggestion
			return models.AlertingRule{}, diag.Diagnostics{
				diag.Diagnostic{
					Severity:      diag.Error,
					Summary:       "alert_delay is only supported for Elasticsearch v8.13 or higher",
					Detail:        "alert_delay is only supported for Elasticsearch v8.13 or higher",
					AttributePath: cty.GetAttrPath("alert_delay"),
				},
			}
```

Will highlight the failing attribute better in the output. I don't feel strongly here though.

internal/kibana/alerting.go
```diff
@@ -181,6 +189,22 @@ func getAlertingRuleFromResourceData(d *schema.ResourceData) (models.AlertingRul
 		rule.Throttle = &t
 	}
 
+	if v, ok := d.GetOk("notify_when"); ok {
+		t := v.(string)
+		rule.NotifyWhen = &t
+	}
+
+	if v, ok := d.GetOk("alert_delay"); ok {
+		if serverVersion.LessThanOrEqual(alertDelayMinSupportedVersion) {
```
tobio:
```suggestion
		if serverVersion.LessThan(alertDelayMinSupportedVersion) {
```

internal/models/alert_rule.go
adcoelho: This will come on a follow up PR where I add more fields. I can remove it for now 👍

internal/models/alert_rule.go
adcoelho: I did it, this made more sense than what I had ;)

dimuon (approved):

Issue #700, opened 2024-08-07
[Bug] Error: json: cannot unmarshal array into Go value of type map[string]interface {}


Alfaxomo:
**Describe the bug**
I am trying to create elasticsearch index templates using the elasticstack_elasticsearch_index_template resource. When running terraform plan, it reports that it will create the index templates correctly. However, when running apply I get the following error for all index templates:

```
│ Error: json: cannot unmarshal array into Go value of type map[string]interface {}
│
│   with elasticstack_elasticsearch_index_template.index_templates["the_index_template_name"],
│   on IndexTemplates.tf line 13, in resource "elasticstack_elasticsearch_index_template" "index_templates":
│   13: resource "elasticstack_elasticsearch_index_template" "index_templates" {
│
```

**To Reproduce**
1. TF configuration used '...'

## locals.tf
```
locals {
  features = {
    feature0 = {
      component_template = "true"
      contextkey         = "feature0"
      ilm_policy        = "policy0"
    }
    feature1 = {
      component_template = "true"
      contextkey         = "feature1"
      ilm_policy        = "policy0"
    }
    feature2 = {
      component_template = "true"
      contextkey         = "feature2"
      ilm_policy        = "policy2"
    }
    feature3 = {
      component_template = "false"
      contextkey         = "feature3"
      ilm_policy        = "policy3"
    }
    feature4 = {
      component_template = "true"
      contextkey         = "feature4"
      ilm_policy        = "policy4"
    }
    feature5 = {
      component_template = "false"
      contextkey         = "feature5"
      ilm_policy        = "policy5"
    }
  }
}
```

## index_templates.tf

```
resource "elasticstack_elasticsearch_component_template" "default_policy_component" {

  name = "policy0"
  template {
    settings = file("component_templates/policy0_settings.json")
    mappings = file("component_templates/policy0_mappings.json")
  }
  metadata = jsonencode({
    description = "Terraform managed resource by DevOps Team"
  })
}

resource "elasticstack_elasticsearch_index_template" "index_templates" {

  for_each = local.features
  
      composed_of = each.value.component_template ? [elasticstack_elasticsearch_component_template.default_policy_component.name] : []
    index_patterns = ["${each.value.contextkey}_*"]
    name           = each.value.contextkey
    template {
      settings = jsonencode({
        index = {
          lifecycle = {
            name = "${each.value.ilm_policy}"
            rollover_alias = "${each.value.contextkey}_write"
          }
        }
      })
[...]

PR #719, opened 2024-08-20
Validate that mappings are a JSON object, not just valid json

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/700

Validates that these attributes are specifically objects, hopefully pointing users to the correct place to look.

### DIFF
--- a/internal/elasticsearch/index/validation.go
+++ b/internal/elasticsearch/index/validation.go
@@ -1,0 +1,22 @@
+package index
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+func stringIsJSONObject(i interface{}, s string) (warnings []string, errors []error) {
+	iStr, ok := i.(string)
+	if !ok {
+		errors = append(errors, fmt.Errorf("expected type of %s to be string", s))
+		return warnings, errors
+	}
+
+	m := map[string]interface{}{}
+	if err := json.Unmarshal([]byte(iStr), &m); err != nil {
+		errors = append(errors, fmt.Errorf("expected %s to be a JSON object. Check the documentation for the expected format. %w", s, err))
+		return
+	}
+
+	return
+}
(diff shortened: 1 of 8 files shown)


internal/elasticsearch/index/validation_test.go
```diff
@@ -0,0 +1,46 @@
...
+func Test_stringIsJSONObject(t *testing.T) {
+	tests := []struct {
+		name                  string
+		fieldVal              interface{}
+		expectedErrsToContain []string
+	}{
+		{
+			name:     "should not return an error for a valid json object",
+			fieldVal: "{}",
+		},
```
biscout42, line 18:
```suggestion
		{
			name:     "should not return an error for a valid json object",
			fieldVal: "{}",
		},
		{
			name:     "should not return an error for a null",
			fieldVal: "null",
		},

```

biscout42 (commented):
Left a suggestion to make `null` explicitly valid json object.

biscout42 (dismissed):

internal/elasticsearch/index/validation.go
```diff
@@ -0,0 +1,22 @@
+package index
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+func stringIsJSONObject(i interface{}, s string) (warnings []string, errors []error) {
```
dimuon, line 8: Does it make sense to use TF's `StringIsValidRegExp` instead?

dimuon: As we use it in combination with `StringIsJSON` in any case...

dimuon: Actually, disregard the comment.

dimuon (approved):

biscout42 (approved):

PR #723, opened 2024-08-21
Add support for remote_indicies to elasticstack_elasticsearch_security_role & elasticstack_kibana_security_role

### LINKED ISSUE #722: [Feature] Add support for remote_indices in elasticstack_elasticsearch_security_role
In order to leverage remote clusters configured with the API key based model, remote_indices permissions must be configured. elasticstack_elasticsearch_security_role does not currently expose this which is preventing us from rolling out cross cluster search.

mholttech:
resolves #722 by adding remote_indices to elasticstack_kibana_security_role & elasticstack_elasticsearch_security_role


internal/kibana/role.go
```diff
@@ -370,6 +436,66 @@ func expandKibanaRoleElasticsearch(v interface{}) *kbapi.KibanaRoleElasticsearch
...
+					}
+
+					newRemoteIndex := kbapi.KibanaRoleElasticsearchRemoteIndice{
+						Names:      names,
+						Clusters:   clusters,
+						Privileges: privileges,
+					}
+
+					if query := index["query"].(string); query != "" {
+						newRemoteIndex.Query = &query
```
tobio, line 485: I think we want to unmarshal this string into `map[string]interface{}` here since the underlying API will be expecting a JSON object, not a plain string.

tobio:
It looks like the build is failing because the committed docs don't match what would be generated. Are you able to run `make docs-generate` and commit the result?

internal/kibana/role.go
```diff
@@ -370,6 +436,66 @@ func expandKibanaRoleElasticsearch(v interface{}) *kbapi.KibanaRoleElasticsearch
 				elasticConfig.Indices = indices
 			}
 
+			if v, ok := userElasticConfig["remote_indices"]; ok {
```
tobio, line 451: We'll want to throw an error if this attribute is configured, but not supported by the underlying cluster. The [index](https://github.com/elastic/terraform-provider-elasticstack/blob/1c23072930f15b03c8548a4ba215e6d16bf98b24/internal/elasticsearch/index/index.go#L776-L781) resource has an example of doing that. 

We'll also need to update the acceptance tests to only configure `remote_indices` when supported.

tobio (commented):
The linked issue talks about the elasticsearch role resource (which is [here](https://github.com/elastic/terraform-provider-elasticstack/blob/1c23072930f15b03c8548a4ba215e6d16bf98b24/internal/elasticsearch/security/role.go)). I'm not sure if adding it to the Kibana role is sufficient for you though?

mholttech:
Good catch. I didn't realize there was two seperate resources. Looks like the api for elasticstack_kibana_security_role is in preview but the work done is still valid. I'll expand the PR to include the intended resource and regenerate the docs.

internal/kibana/role.go
mholttech: This is configured the same way it is for Indicies, https://github.com/elastic/terraform-provider-elasticstack/blob/0118d48ecf7015bc6b441c144c29a70072b58dfc/internal/kibana/role.go#L341-L343

I don't know enough about the underlying structures

mholttech:
Everything should be in place for the elasticstack_kibana_security_role now. Still working on the elasticstack_elasticsearch_security_role

mholttech:
forgot about TestAccDataSourceKibanaSecurityRole, i've commented out the remote_indices for now, will add the version constraints tomorrow.

mholttech:
@tobio This is ready for another review

internal/elasticsearch/security/role.go
```diff
@@ -286,6 +359,71 @@ func resourceSecurityRolePut(ctx context.Context, d *schema.ResourceData, meta i
 		role.Indices = indices
 	}
 
+	if v, ok := d.GetOk("remote_indices"); ok {
+		definedRemoteIndices := v.(*schema.Set)
+		if definedRemoteIndices.Len() > 0 {
+			if serverVersion.LessThan(minSupportedRemoteIndicesVersion) {
+				return diag.FromErr(fmt.Errorf("'remote_indices' is supported only for Kibana v%s and above", minSupportedRemoteIndicesVersion.String()))
+			}
+		}
```
tobio:
```suggestion
		if definedRemoteIndices.Len() > 0 && serverVersion.LessThan(minSupportedRemoteIndicesVersion) {
			return diag.FromErr(fmt.Errorf("'remote_indices' is supported only for Elasticsearch v%s and above", minSupportedRemoteIndicesVersion.String()))
		}
```

internal/elasticsearch/security/role.go
```diff
@@ -418,6 +561,29 @@ func flattenIndicesData(indices *[]models.IndexPerms) []interface{} {
 	}
 	return make([]interface{}, 0)
 }
+func flattenRemoteIndicesData(remoteIndices *[]models.RemoteIndexPerms) []interface{} {
```
tobio: Why `remoteIndices *[]models.RemoteIndexPerms` here? It looks like the only usage takes the address of a plain slice. 

We can remove the if block and drop a level of nesting without the pointer typed variable here.

tobio: Ok, it looks like you're just mimicking the current code. IMO we should change this to just take a direct slice reference (which can be nil anyway). It also means we don't have the `remoteIndexes`/`remoteIndices` naming above which confused me for longer than it should have.

internal/kibana/role.go
```diff
@@ -454,13 +597,42 @@ func flattenKibanaRoleIndicesData(indices *[]kbapi.KibanaRoleElasticsearchIndice
 	return make([]interface{}, 0)
 }
 
+func flattenKibanaRoleRemoteIndicesData(indices *[]kbapi.KibanaRoleElasticsearchRemoteIndice) []interface{} {
```
tobio: Same thing here, I don't think we're getting any value out of the pointer typed param.

internal/models/models.go
```diff
@@ -73,24 +73,26 @@ type UserPassword struct {
...
-	Indices      []IndexPerms           `json:"indices,omitempty"`
-	Metadata     map[string]interface{} `json:"metadata,omitempty"`
-	RusAs        []string               `json:"run_as,omitempty"`
-	Restriction  *Restriction           `json:"restriction,omitempty"`
+	Name          string                 `json:"-"`
+	Applications  []Application          `json:"applications,omitempty"`
+	Global        map[string]interface{} `json:"global,omitempty"`
+	Cluster       []string               `json:"cluster,omitempty"`
+	Indices       []IndexPerms           `json:"indices,omitempty"`
+	RemoteIndices []RemoteIndexPerms     `json:"remote_indices,omitempty"`
```
tobio: IIUC this would also mean we add support for `remote_indices` in the API Key resource (🎉). Can we mention that in the changelog as well please.

tobio (commented):
Couple of minor comments. Generally this LGTM. Thanks so much for adding this! It's a decent chunk of code here. ❤️

mholttech:
> Couple of minor comments. Generally this LGTM. Thanks so much for adding this! It's a decent chunk of code here. ❤️ 

Thanks! I've reviewed your feedback and now that i've learned how everything works and it's in a working state I'll attempt some of the suggested code cleanup in indices and remote_indices before we merge this.

internal/models/models.go
mholttech: This actually shouldn't be here so I'm removing it

mholttech:
@tobio I just pushed some code improvements based on your feedback.

tobio (approved):
LGTM. This is a big chunk of code, thanks for adding this to the provider!

mholttech:
@tobio when will the next version be released so we can begin leveraging this?

Issue #726, opened 2024-08-23
[Bug] Inconsistent state for elasticstack_kibana_synthetics_private_location space_id


doom160:
**Describe the bug**
I was testing out the new resource `elasticstack_kibana_synthetics_private_location` 
Seems like the `elasticstack_fleet_agent_policy.namespace` has to match `elasticstack_kibana_synthetics_private_location.space_id`

if not you will get an error like this
```
│ When applying changes to module.agent_policies.elasticstack_kibana_synthetics_private_location.main["dev"], provider
│ "provider[\"registry.terraform.io/elastic/elasticstack\"]" produced an unexpected new value: .space_id: was null, but now
│ cty.StringVal("dev").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
```

Trying to understand these 2 values should be matching, it throws the same error if you set unset space_id as null because agent_policy namespace would have been "default"

**To Reproduce**

Creating a sample code
```
resource "elasticstack_fleet_agent_policy" "main" {
  for_each        = toset(var.environments)
  name            = each.value
  namespace       = each.value
  description     = each.value

  sys_monitoring  = true
  monitor_logs    = true
  monitor_metrics = true
}

resource "elasticstack_kibana_synthetics_private_location" "main" {
  for_each        = toset(var.environments)
  label           = each.value
  agent_policy_id = elasticstack_fleet_agent_policy.main[each.value].id
  space_id = "default"
  tags            = [each.value]
  geo = {
    lat = 0
    lon = 0
  }
}
```

**Expected behavior**
My understanding is space_id is a different thing as namespace, so i dont think it should be matching

**Versions (please complete the following information):**
 - OS: mac
 - Terraform Version v1.6.4
 - Provider version v0.11.6
 - Elasticsearch Version 8.12.2

**Additional context**

tobio:
@biscout42 do you have a chance to take a look at this?

biscout42:
Let me have a look. Indeed, `space_id` is a namespace, or space in kibana terminology, but there might be additional restriction by kibana between policy and location (one example - private location has same id as policy).

@doom160 , does it work for you if you set `space_id = each.value` for private location?

biscout42:
I have done a test to verify, if I can create private location in other space, than policy and it doesn't work.

[...]

PR #733, opened 2024-08-28
remove space_id parameter from private locations

biscout42:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/726


tobio (commented):
IIUC, the space id parameter in the API is more or less ignored, and instead the namespace/space_id is inherited from the agent policy? Assuming that's correct, LGTM

Can we get a changelog entry for this one calling out the breaking change.

biscout42:
Thanks, @tobio , 

> IIUC, the space id parameter in the API is more or less ignored, and instead the namespace/space_id is inherited from the agent policy?

Yes, that is correct.

tobio (approved):

PR #753, opened 2024-09-05
Add support for the`frequency` field in the Create Rule API

adcoelho:
related to [kibana/#186963](https://github.com/elastic/kibana/issues/186963)

## Summary

As a follow-up to what was done in https://github.com/elastic/terraform-provider-elasticstack/pull/715 I am now adding support for another field.

## Testing notes

I could not update the `TestAccResourceAlertingRule` tests in `internal/kibana/alerting_test.go` because this new field needs the `actions` block that requires the ID of an existing connector.

(We already supported the `actions` field but those were not included in the tests in the initial PR and I think the reasoning was the same. Maybe @tobio knows?).

To test I started Kibana locally and configured an MS Teams connector manually.

I then updated the test resources with the new fields and my connector ID and ran `make testacc TESTARGS='-run ^TestAccResourceAlertingRule$$'`, which worked as expected.

Additionally, I added the file `resource_rule_action_frequency.tf` and tested it with Terraform CLI. (I substituted `elasticstack_kibana_action_connector.index_example.connector_id` for my connector ID).

I also updated the tests in `internal/clients/kibana/alerting_test.go` to include all possible combinations of the new fields in the Rule API responses.


internal/kibana/alerting.go
```diff
@@ -47,7 +48,7 @@ func ResourceAlertingRule() *schema.Resource {
 		"notify_when": {
 			Description:  "Defines how often alerts generate actions. Valid values include: `onActionGroupChange`: Actions run when the alert status changes; `onActiveAlert`: Actions run when the alert becomes active and at each check interval while the rule conditions are met; `onThrottleInterval`: Actions run when the alert becomes active and at the interval specified in the throttle property while the rule conditions are met. NOTE: This is a rule level property; if you update the rule in Kibana, it is automatically changed to use action-specific `notify_when` values.",
 			Type:         schema.TypeString,
-			Required:     true,
+			Optional:     true,
```
adcoelho, line 54: This became optional because the preferred way is to pass the value inside the `frequency` in `actions`.

examples/resources/elasticstack_kibana_alerting_rule/resource_rule_action_frequency.tf
```diff
@@ -0,0 +1,38 @@
...
+    timeField           = "@timestamp"
+    aggField            = "version"
+    termField           = "name"
+  })
+  rule_type_id = ".index-threshold"
+  interval     = "1m"
+  enabled      = true
+
+  actions {
+    # Should be the id of a MS Teams connector
```
cnasikas, line 26: nit: Of any connector, right?

docs/resources/kibana_alerting_rule.md
```diff
@@ -64,6 +63,7 @@ Could not create API key - Unsupported scheme "ApiKey" for granting API Key
 - `actions` (Block List) An action that runs under defined conditions. (see [below for nested schema](#nestedblock--actions))
 - `alert_delay` (Number) A number that indicates how many consecutive runs need to meet the rule conditions for an alert to occur.
 - `enabled` (Boolean) Indicates if you want to run the rule on an interval basis.
+- `notify_when` (String) Defines how often alerts generate actions. Valid values include: `onActionGroupChange`: Actions run when the alert status changes; `onActiveAlert`: Actions run when the alert becomes active and at each check interval while the rule conditions are met; `onThrottleInterval`: Actions run when the alert becomes active and at the interval specified in the throttle property while the rule conditions are met. NOTE: This is a rule level property; if you update the rule in Kibana, it is automatically changed to use action-specific `notify_when` values.
```
cnasikas: Should we add a note that this is [deprecated](https://www.elastic.co/docs/api/doc/kibana/v8/operation/operation-createruleid#operation-createruleid-body-application-json-elastic-api-version-2023-10-31-notify_when)?

docs/resources/kibana_alerting_rule.md
```diff
@@ -86,8 +86,21 @@ Required:
 
 Optional:
 
+- `frequency` (Block List, Max: 1) The parameters for the action, which are sent to the connector. (see [below for nested schema](#nestedblock--actions--frequency))
```
cnasikas: How do you feel about using the text from [here](https://www.elastic.co/docs/api/doc/kibana/v8/operation/operation-createruleid#operation-createruleid-body-application-json-elastic-api-version-2023-10-31-actions-frequency)?

examples/resources/elasticstack_kibana_alerting_rule/resource_rule_action_frequency.tf
adcoelho: In this particular example the action `params` are for a ms teams connector.

internal/clients/kibana/alerting.go
```diff
@@ -20,11 +20,26 @@ func ruleResponseToModel(spaceID string, res *alerting.RuleResponseProperties) *
...
+		a := models.AlertingRuleAction{
 			Group:  action.Group,
 			ID:     action.Id,
 			Params: action.Params,
-		})
+		}
+
+		if alerting.IsNil(action.Frequency) {
+			a.Frequency = nil
+		} else {
```
tobio:
```suggestion
		if !alerting.IsNil(action.Frequency) {
```

The field will be nil by default no?

internal/clients/kibana/alerting.go
```diff
@@ -68,11 +83,28 @@ func ruleActionsToActionsInner(ruleActions []models.AlertingRuleAction) []alerti
...
+		actionToAppend := alerting.ActionsInner{
 			Group:  action.Group,
 			Id:     action.ID,
 			Params: action.Params,
-		})
+		}
+
+		if alerting.IsNil(action.Frequency) {
+			actionToAppend.Frequency = nil
+		} else {
```
tobio:
```suggestion
		if !alerting.IsNil(action.Frequency) {
```

Similarly

internal/kibana/alerting.go
```diff
@@ -235,11 +263,38 @@ func getActionsFromResourceData(d *schema.ResourceData) ([]models.AlertingRuleAc
...
+			currentAction := fmt.Sprintf("actions.%d", i)
+
+			if _, ok := d.GetOk(currentAction + ".frequency"); ok {
+				if serverVersion.LessThan(alertDelayMinSupportedVersion) {
+					return []models.AlertingRuleAction{}, diag.Diagnostics{
+						diag.Diagnostic{
+							Severity: diag.Error,
+							Summary:  "actions.frequency is only supported for Elasticsearch v8.13 or higher",
+							Detail:   "actions.frequency is only supported for Elasticsearch v8.13 or higher",
+						},
```
tobio:
```suggestion
						diag.Errorf("actions.frequency is only supported for Elasticsearch v8.13 or higher"),
```

tobio (commented):
> I could not update the TestAccResourceAlertingRule tests in internal/kibana/alerting_test.go because this new field needs the actions block that requires the ID of an existing connector.

IIRC alert management was implemented before connector management. Now that we have the connector resource I think we should update the acceptance tests to create say a server_log connector so we can reference that in the actions. WDYT?

internal/kibana/alerting.go
cnasikas: Should we do the same for the `throttle`? It seems both became optional in 8.6+ in this [PR](https://github.com/elastic/kibana/pull/144130). Also, what will happen to <=8.5 if the fields are optional in TF but required in Kibana?

internal/kibana/alerting.go
```diff
@@ -235,11 +263,38 @@ func getActionsFromResourceData(d *schema.ResourceData) ([]models.AlertingRuleAc
...
 				Group:  action["group"].(string),
 				ID:     action["id"].(string),
 				Params: params,
-			})
+			}
+
+			currentAction := fmt.Sprintf("actions.%d", i)
+
+			if _, ok := d.GetOk(currentAction + ".frequency"); ok {
+				if serverVersion.LessThan(alertDelayMinSupportedVersion) {
```
cnasikas: It seems that the `frequency` was added in [8.6+](https://github.com/elastic/kibana/pull/144130).

internal/kibana/alerting.go
adcoelho: > Also, what will happen to <=8.5 if the fields are optional in TF but required in Kibana?

I'm not sure how to handle that here, I can specify they are either optional or not — not depending on the version.

Maybe the documentation can specify that they are optional only from `8.6`?

internal/kibana/alerting.go
adcoelho: Good catch, I added this in the beginning and forgot to update 🙏

adcoelho:
@tobio 

> Now that we have the connector resource I think we should update the acceptance tests to create say a server_log connector so we can reference that in the actions. WDYT?

Ah nice, I hadn't noticed the `connector.go` file! I can update `TestAccResourceAlertingRule` to include the actions but is it ok if I do that in my next PR? I have one last field to add to this resource and I can do the tests together then.

internal/kibana/alerting.go
tobio: > Maybe the documentation can specify that they are optional only from 8.6

There's similar notes in the docs for other resources. 

It would also be a good idea to apply this validation within the resource code itself, i.e throw an error if create/update is called against ES<=8.5 and `notify_when` is not defined.

tobio (approved):
LGTM

Let's add a note about `notify_when` being required in 8.5 and lower and add some validation in the resource code itself enforcing that.

internal/kibana/alerting.go
cnasikas: I would also suggest throwing an error if possible on the resource level. @tobio Do you have an example of another resource that achieves the same?

adcoelho: > I would also suggest throwing an error if possible on the resource level.

I already do something similar for `alert_delay` [here](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/kibana/alerting.go#L195).

I'll add one for `notify_when`(and `throttle`, I will double check the versions).

tobio (approved):

PR #772, opened 2024-09-14
synthetics icmp and browser monitors support

biscout42:
implelents https://github.com/elastic/terraform-provider-elasticstack/issues/610


biscout42:
Error was from `TestAccSpacesDataSource`, trying to re-run.

```
    data_source_test.go:11: Step 1/1 error: Error running post-apply plan: exit status 1
        
        Error: unable to list spaces
        
          with data.elasticstack_kibana_spaces.all_spaces,
          on terraform_plugin_test.tf line 7, in data "elasticstack_kibana_spaces" "all_spaces":
           7: data "elasticstack_kibana_spaces" "all_spaces" {
        
        400 Bad Request
--- FAIL: TestAccSpacesDataSource
```

internal/kibana/synthetics/schema.go
```diff
@@ -176,6 +196,61 @@ func monitorConfigSchema() schema.Schema {
...
+				MarkdownDescription: "The inline script.",
+			},
+			"screenshots": schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "Controls the behavior of the screenshots feature.",
+				Validators: []validator.String{
+					stringvalidator.OneOf("on", "off", "only-on-failure"),
+				},
+				Computed: true,
+				Default:  stringdefault.StaticString("on"),
```
tobio: From the [docs](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework@v1.11.0/resource/schema#BoolAttribute):

> If the attribute is computed and the value could be altered by other changes then a default should be avoided and a plan modifier should be used instead.

The same goes for the other `Default` values in the schema. 

I'm not sure if these _need_ to be computed, but if they do we likely want something like:

```
PlanModifiers: []{stringplanmodifiers.UseStateForUnknown(), defaultModifier("on")}
```

internal/kibana/synthetics/schema.go
```diff
@@ -488,6 +581,47 @@ func (v *tfTCPMonitorFieldsV0) toTfTCPMonitorFieldsV0(api *kbapi.SyntheticsMonit
...
+	if err != nil {
+		return nil, err
+	}
+	return &tfICMPMonitorFieldsV0{
+		Host: types.StringValue(api.Host),
+		Wait: types.Int64Value(wait),
+	}, nil
+}
+
+func (v *tfBrowserMonitorFieldsV0) toTfIBrowserMonitorFieldsV0(api *kbapi.SyntheticsMonitor) (*tfBrowserMonitorFieldsV0, error) {
```
tobio:
```suggestion
func (v *tfBrowserMonitorFieldsV0) toTfBrowserMonitorFieldsV0(api *kbapi.SyntheticsMonitor) (*tfBrowserMonitorFieldsV0, error) {
```

I think this is a typo?

internal/kibana/synthetics/schema.go
biscout42: Thanks a lot, let me check. The issue I had - plan was different after creation, because API returns default value for this attribute, but I didn't set it.

biscout42: I"m sorry, it doesn't work. TF complains 
```
Schema Using
        Attribute Default For Non-Computed Attribute: Attribute "browser.screenshots"
        must be computed when using default.
```

biscout42: Ok, I'll leave only computed then, as a sign that API defines values, if not set.

biscout42: FYI: according to  https://github.com/hashicorp/terraform-plugin-framework/issues/285 there is no default for plan modifier and `UseStateForUnknown` should be enough. At least, I wasn't able to find default modifier in plugin framework. Please, let me know if I need to add it and how.

biscout42: @tobio , should be done, could you re-check?

biscout42:
FYI: auto-merge is on.

tobio (approved):

PR #774, opened 2024-09-16
Add support for the `alerts_filter` param in the Create Rule API

adcoelho:
Related to [kibana/#186963](https://github.com/elastic/kibana/issues/186963)

## Summary

Follow-up to https://github.com/elastic/terraform-provider-elasticstack/pull/715 and https://github.com/elastic/terraform-provider-elasticstack/pull/753. This should be the last PR 🎉 

I will update the documentation next.

Since we now support the connector resource I also added tests for the `actions` and the `frequency`(as discussed [here](https://github.com/elastic/terraform-provider-elasticstack/pull/753#pullrequestreview-2286058165)).


adcoelho:
The failing tests are because the new field wasn't available initially for all rule types.

I am trying to rewrite the tests with a rule type that works in every version.

It shouldn't affect the review 🙌

lcawl:
The text fields LGTM

internal/utils/validation.go
```diff
@@ -40,3 +40,23 @@ func StringIsElasticDuration(i interface{}, k string) (warnings []string, errors
...
+	v, ok := i.(string)
+	if !ok {
+		return nil, []error{fmt.Errorf("expected type of %s to be string", k)}
+	}
+
+	if v == "" {
+		return nil, []error{fmt.Errorf("%q string is not a valid time in HH:mm format: [empty]", k)}
+	}
+
+	r := regexp.MustCompile(`^([0-9]|0[0-9]|1[0-9]|2[0-3]):[0-5][0-9]$`)
```
tobio, line 55: This will accept `1:10` but not `10:1` which feels a bit inconsistent. I don't feel strongly here, but maybe we should just enforce two digit hours as the simplest option?

tobio (commented):
LGTM, can you add a changelog entry for this as well.

internal/utils/validation.go
adcoelho: This is a good point but if anything I don't like that we allow 1:10 😅 So I could alternatively do:

```suggestion
	r := regexp.MustCompile(`^(0[0-9]|1[0-9]|2[0-3]):[0-5][0-9]$`)
```

But this is exactly the same validation we have in the rule API so I'm not sure it is worth it.

cnasikas (approved):
Great work!!

CHANGELOG.md
```diff
@@ -1,11 +1,12 @@
 ## [Unreleased]
 
+- Add the `alerts_filter` field to the `actions` in the Create Rule API ([#774](https://github.com/elastic/terraform-provider-elasticstack/pull/774))
 - Add the `alert_delay` field to the Create Rule API ([#715](https://github.com/elastic/terraform-provider-elasticstack/pull/715))
 - Add support for data_stream `lifecycle` template settings ([#724](https://github.com/elastic/terraform-provider-elasticstack/pull/724))
 - Fix a provider panic when `elasticstack_kibana_action_connector` reads a non-existant connector ([#729](https://github.com/elastic/terraform-provider-elasticstack/pull/729))
 - Add support for `remote_indicies` to `elasticstack_elasticsearch_security_role` & `elasticstack_kibana_security_role` (#723)[https://github.com/elastic/terraform-provider-elasticstack/pull/723]
 - Fix error handling in `elasticstack_kibana_import_saved_objects` ([#738](https://github.com/elastic/terraform-provider-elasticstack/pull/738))
-- Remove `space_id` parameter from private locations to fix inconsistent state for `elasticstack_kibana_synthetics_private_location` `space_id` ([#733](https://github.com/elastic/terraform-provider-elasticstack/pull/733))
```
tobio, line 8: I genuinely like seeing other people have their editors configured to trim trailing whitepace.

tobio (approved):
🚢

PR #778, opened 2024-09-18
impl framework data_source.fleet_enrollment_tokens

daemitus:
- Migrated the fleet enrollment_tokens data_source to the plugin framework. 
- removed resource.isReady from several places, added the same check to ConvertProviderData. The error will propagate up when resource.Configure is called, no need to keep it in the CRUD methods. 
- converted the enrollment_token fleet APIs to return fwdiags.


internal/utils/utils.go
```diff
@@ -244,3 +244,21 @@ func FlipMap[K comparable, V comparable](m map[K]V) map[V]K {
...
+}
+
+func FwDiagsAsError(diags fwdiag.Diagnostics) error {
+	for _, diag := range diags {
+		if diag.Severity() == fwdiag.SeverityError {
+			return fmt.Errorf("%s: %s", diag.Summary(), diag.Detail())
+		}
+	}
+	return nil
+}
```
tobio:
```suggestion
```

This appears to be unused?

tobio (commented):
Thanks for converting this! Are you able to add an entry to CHANGELOG.md for this?

internal/utils/tfsdk.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// SliceToListType converts a tfsdk naive []T1 into an types.List of []T2.
+// This handles both structs and simple types to attr.Values.
+func SliceToListType[T1 any, T2 any](ctx context.Context, value []T1, elemType attr.Type, path path.Path, diags diag.Diagnostics, iteratee func(item T1) T2) types.List {
```
daemitus, line 14: What is your hot take on helpers like this? While its not much of a problem in this data_source, I've had a ton of issues with naive slices as compared to types.List members. In the more complicated nested structs, these let you just inline a lot of boilerplate.

tobio: 👍 I think there's likely a lot of room for more helpers like this in this codebase.

tobio (approved):
🎉 love it. Thanks again for this!

Issue #767, opened 2024-09-11
[Bug] elasticsearch_index.alias and import regressions


daemitus:
Regression due to #698

The alias on my existing index wants to flip the `is_write_index` field
Additionally, import is no longer supported.

Elasticsearch version:  8.15.1
Provider version: main/head

```
  # module.elastic.module.signals.elasticstack_elasticsearch_index.signals_defender will be updated in-place
  ~ resource "elasticstack_elasticsearch_index" "index" {
      - default_pipeline       = "something" -> null
        id                     = "my-cluster-id/my-index-name-000001"
        name                   = "my-index-name-000001"
      ~ settings_raw           = jsonencode(
            {
               ... everything is deleted...
            }
        ) -> (known after apply)
        # (7 unchanged attributes hidden)

      - alias {
          - is_hidden      = false -> null
          - is_write_index = true -> null
          - name           = "my-alias" -> null
            # (3 unchanged attributes hidden)
        }
      + alias {
          + is_hidden      = false
          + is_write_index = false
          + name           = "my-alias"
            # (3 unchanged attributes hidden)
        }
    }
```

tobio:
@daemitus are you able to provide the output of `GET /my-index-name-000001?flat_settings=true` and the actual TF definition?

daemitus:
```
terraform {
  required_version = "1.8.2"
  required_providers {
    elasticstack = {
      source  = "elastic/elasticstack"
      version = "= 0.11.6"
    }
  }
}

provider "elasticstack" {
  elasticsearch {
    username  = "elastic"
    password  = "password"
    endpoints = ["https://localhost:9200"]
    insecure  = true
  }
}

resource "elasticstack_elasticsearch_index_template" "test" {
  name           = "test"
  index_patterns = ["test*"]
  template {
    settings = jsonencode({
      default_pipeline = ".fleet_final_pipeline-1"
      lifecycle        = { name = ".monitoring-8-ilm-policy" }
    })
  }
}

resource "elasticstack_elasticsearch_index" "test" {
  name                = "test-index-000001"
  deletion_protection = false
  alias {
    name           = "test-alias"
    is_write_index = true
  }
  lifecycle {
    ignore_changes = [mappings]
  }
  depends_on = [elasticstack_elasticsearch_index_template.test]
}
```

- Apply resources as typical
- Add ~/.terraformrc with 
```
provider_installation {
    dev_overrides {
[...]

PR #779, opened 2024-09-19
Don't populate settings during read

tobio:
There's a fair bit of complexity around handling computed settings. I'll look at adding this back in a follow up PR, but want to get this resource in a working state so we can make a new release.

Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/767


dimuon:
@tobio , some other fields also use `Default` together with `Computed` in the same scheme (e.g. `mappings`, `deletion_protection` and few others). Shall we consider removing `Default` for them as well?

dimuon (approved):

dimuon:
Just as a side note - kibana data view also uses that combination of `Default` and `Computed`.

PR #785, opened 2024-09-20
more fleet framework migrations

daemitus:
- add a fleen gen transform to de-pointer `*[]sometype` and `*map[string]sometype` fields
- migrate resource fleet_agent_policy
- migrate resource fleet_integration
- migrate data_source fleet_integration
- migrate resource fleet_server_host


generated/fleet/getschema.go
tobio, line 1: IMO we should revert this change. The more we manipulate the published spec, the more difficult it is to maintain this code moving forward. AFAICT this doesn't fix anything, but instead makes the generated code slightly simpler to use.

internal/fleet/agent_policy/resource_test.go
tobio, line 1: Can we add an acceptance test here (and for the other resources) verifying state compatibility between the SDK and FW resources ([example](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/index/index/acc_test.go#L80-L175)).

internal/fleet/integration/update.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"context"
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients/fleet"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func (r *integrationResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
```
tobio, line 9: Can we share the `Create` implementation here instead?

internal/fleet/agent_policy/resource_test.go
```diff
@@ -57,7 +57,7 @@ func TestAccResourceAgentPolicy(t *testing.T) {
 				ResourceName:            "elasticstack_fleet_agent_policy.test_policy",
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"skip_destroy"},
+				ImportStateVerifyIgnore: []string{"monitor_metrics", "monitor_logs", "skip_destroy"},
```
tobio: What's happening here? This PR populates these fields on `Read`, why are we not setting them to computed?

daemitus:
I was having problems earlier testing those, I think it has to do with on
read, not knowing if they should be false or null. I'll try that bit again
with a computerd/default false, and on read set false if not set.

Regarding the x-pointers flag, I can revert that. Will just need to change
up the helpers to do pointers instead. As you said, it doesn't change much.

As you've brought up manipulation of the published spec, whats your take
then on all of the nested anonymous structs? Fleet isn't too bad, but
Kibana's is a disaster for generation at the moment.

daemitus:
ok. 
- x-go-pointer flag is dropped.
- fixed agent policy monitoring test ignores finally
- shared integration.create with update
- added FromSDK tests to all 3 migrated resources

tobio:
> whats your take then on all of the nested anonymous structs?

Yea, these are a total pain. Ultimately we need to fix these in the source though so anyone using the specs can benefit.

tobio (approved):
❤️ thanks for doing this work. It simplifies ongoing development loads!

daemitus:
And here I couldn't sleep this morning. Looks like I'll be rebasing the
other pr now.

Issue #676, opened 2024-07-09
System integration not added although "sys_monitoring" is set to true [Bug]


Kronusus:
**Describe the bug**
I'm trying to create and configure an elastic agent policy according to the current documenation:
// Definition of agent policy
resource "elasticstack_fleet_agent_policy" "test_policy" {
  name            = "Test Policy"
  namespace       = "default"
  description     = "Test Agent Policy"
  sys_monitoring  = true
  monitor_logs    = true
  monitor_metrics = true
}

According to the documentation, I would expect that the system integration is automatically added to my created "test_policy", since I set sys_monitoring to true.

When I execute terraform apply, this also runs without an error. However, the created configuration doesn't contain the systems integration.

**To Reproduce**
1. After the necessary base set-up of the elasticstack provider, try to create an agent policy according to the code above.

**Expected behavior**
Created agent policy should contain system integration.

**Versions (please complete the following information):**
 - OS: [e.g. Windows 11]
 - Terraform Version [e.g. 1.8.1]
 - Provider version [e.g. v0.11.4]
 - Elasticsearch Version [e.g. 8.14.2]

thanks for any assistance on this behaviour.

RainrainWu:
It seems like the `sys_monitoring` has to be passed via URL param, which is not quite aligned with the current implementation based on the request body.

https://www.elastic.co/guide/en/fleet/current/fleet-api-docs.html#create-agent-policy-api

I can take this issue and try to propose a workaround these days via request editor functions and see if it is feasible 😃 , what do u think?

Kronusus:
I have not enough insides but to me it sounds good :)

PR #792, opened 2024-09-24
Apply the sys_monitoring attribute during creation

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/676

This attribute was being entirely ignored by the resource.


dimuon (approved):
Changelog entry is needed.

PR #797, opened 2024-09-26
migrate fleet integration_policy to framework, fix secret churn

### LINKED ISSUE #689: [Bug] elasticstack_fleet_integration_policy render consistent diff for policy secret
**Describe the bug**

When configuring [policy secret](https://www.elastic.co/guide/en/fleet/8.14/agent-policy.html#agent-policy-secret-values) via integration policy resource, the provider and terraform always renders a diff even though the value is not changed.

I'm trying to configure the universal profiler integration.

**To Reproduce**

```terraform
resource "elasticstack_fleet_integration_policy" "profiling" {
  name      = "profiling"

  agent_policy_id     = elasticstack_fleet_agent_policy.this.policy_id
  integration_name    = data.elasticstack_fleet_integration.profiling.name
  integration_version = data.elasticstack_fleet_integration.profiling.version

  input {
    input_id     = "universal_profiling_agent-pf-host-agent"
    streams_json = jsonencode({})
    vars_json = jsonencode({
      "profiler.collector"    = "cloud-es.io"
      "profiler.secret_token" = "Hunter2"
      "profiler.disable_tls"  = false
    })
  }
}
```

Apply once, then make a plan.

**Expected behavior**
There should not be any diff in the second plan.

 - OS: MacOS
 - Terraform Version: v1.7.5
 - Provider version 0.1.4
 - Elasticsearch Version 8.14.3

daemitus:
Dependent on #785 (will fix this up after merge to clean up the diff)
Solves #689 

- Migrate fleet_integration_policy to framework
- Migrate fleet_output to framework
- Fix #689, churn on integration policy secrets by storing a map of `secretID -> actual` in the private resource state.  

Actually testing the secrets required implementing an actual fleet server. So theres a bunch of changes in the makefile and github actions setting that up. 

- Change from previously: for integration policies, sortInputs would leave `new` in the original order if `old` was not present, e.g. during import. Since only the TCP integration was ever tested, the go maps are unordered/randomized issue never cropped up. Now `new` is sorted by inputID if `old` is not present. If both are present, existing functionality is preserved.


breml:
@daemitus Thanks for the work you have put into this.

if I understand correctly, this PR is based on #785. In order to make the diff in this PR only containing the changes necessary to fix the issue with the secret ref (#689), I suggest to use `daemitus:feat` as merge target instead of `main`, at least for the time until `daemitus:feat` is merged. What do you think?

daemitus:
Oh sure

.github/workflows/test.yml
```diff
@@ -63,7 +63,6 @@ jobs:
           xpack.security.enabled: true
           xpack.security.authc.api_key.enabled: true
           xpack.security.authc.token.enabled: true
-          xpack.security.http.ssl.enabled: false
```
daemitus, line 66: defaults to false, not included in the makefile

.github/workflows/test.yml
```diff
@@ -80,10 +79,28 @@ jobs:
...
+          FLEET_SERVER_POLICY_ID: fleet-server
+          FLEET_SERVER_ELASTICSEARCH_HOST: http://elasticsearch:9200
+          FLEET_SERVER_ELASTICSEARCH_INSECURE: "true"
+          FLEET_SERVER_INSECURE_HTTP: "true"
+          KIBANA_HOST: http://kibana:5601
+          KIBANA_FLEET_SETUP: "1"
+          KIBANA_FLEET_PASSWORD: ${{ env.ELASTIC_PASSWORD }}
+        ports:
+          - 8220:8220
+        options: --restart="unless-stopped"
```
daemitus, line 103: To test for secret storage, an actual fleet server is necessary. It'll fail until the kibana password is set, so keep restarting it.

.github/workflows/test.yml
```diff
@@ -123,8 +140,6 @@ jobs:
       - name: Setup Kibana user
         run: make set-kibana-password
         env:
-          ELASTICSEARCH_ENDPOINTS: "http://localhost:9200"
-          ELASTICSEARCH_USERNAME: "elastic"
```
daemitus, line 127: unused arguments in this makefile directive

.github/workflows/test.yml
```diff
@@ -134,10 +149,17 @@ jobs:
         run: |-
           echo "apikey=$(make create-es-api-key | jq -r .encoded)" >> "$GITHUB_OUTPUT"
         env:
-          ELASTICSEARCH_ENDPOINTS: "http://localhost:9200"
-          ELASTICSEARCH_USERNAME: "elastic"
```
daemitus, line 138: unused arguments in this makefile directive

Makefile
```diff
@@ -31,7 +31,11 @@ KIBANA_SYSTEM_USERNAME ?= kibana_system
 KIBANA_SYSTEM_PASSWORD ?= password
 KIBANA_API_KEY_NAME ?= kibana-api-key
 
+FLEET_NAME ?= terraform-elasticstack-fleet
+FLEET_ENDPOINT ?= https://$(FLEET_NAME):8220
+
 SOURCE_LOCATION ?= $(shell pwd)
+, := ,
```
daemitus, line 38: escaping JSON commas in the setup fleet payloads. This makes me sad, but not much of a better way to do it when wrapping in the retry func

internal/fleet/integration_policy/secrets.go
```diff
@@ -0,0 +1,165 @@
...
+						if original, ok := secrets[refID]; ok {
+							vars[key] = original
+						}
+					}
+				}
+			}
+		}
+	}
+
+	handleVars(utils.Deref(resp.Vars))
```
daemitus, line 97: fwiw, without using the x-go-remove-unnecessary-pointers flag, you have to sprinkle these deref methods everywhere. Being able to range a nil map is nice.

internal/utils/tfsdk.go
```diff
@@ -2,77 +2,165 @@ package utils
...
+	Index int
+	Path  path.Path
+	Diags diag.Diagnostics
+}
+
+type MapMeta struct {
+	Key   string
+	Path  path.Path
+	Diags diag.Diagnostics
+}
```
daemitus, line 24: The idea with using these, is to make the iter functions in each of these helpers a little less wordy. If you don't inline the function, you'll still have access to the parent metadata.

internal/fleet/integration_policy/models.go
```diff
@@ -0,0 +1,149 @@
...
+
+	return body, diags
+}
+
+// sortInputs will sort the 'incoming' list of input definitions based on
+// the order of inputs defined in the 'existing' list. Inputs not present in
+// 'existing' will be placed at the end of the list. Inputs are identified by
+// their ID ('input_id'). The 'incoming' slice will be sorted in-place.
+func sortInputs(incoming []integrationPolicyInputModel, existing []integrationPolicyInputModel) {
+	if len(existing) == 0 {
```
daemitus, line 123: This is a change from previous that I think addresses a bug that was not noticed before. When importing, existing will be empty/null. And if you have more than 1 import, the order will get randomized (thanks go-maps), so the import test will randomly fail.

generated/fleet/getschema.go
tobio, line 1: Can we just use a newer version of the spec?

internal/fleet/integration_policy/models.go
```diff
@@ -0,0 +1,149 @@
...
+		} else {
+			oldInputs := utils.ListTypeAs[integrationPolicyInputModel](ctx, model.Input, path.Root("input"), diags)
+			sortInputs(newInputs, oldInputs)
+
+			inputList, d := types.ListValueFrom(ctx, getInputType(), newInputs)
+			diags.Append(d...)
+
+			model.Input = inputList
+		}
+	}
```
tobio:
```suggestion
	newInputs := utils.TransformMapToSlice(data.Inputs, path.Root("input"), diags,
		func(inputData fleetapi.PackagePolicyInput, meta utils.MapMeta) integrationPolicyInputModel {
			return integrationPolicyInputModel{
				InputID:     types.StringValue(meta.Key),
				Enabled:     types.BoolValue(inputData.Enabled),
				StreamsJson: utils.MapToNormalizedType(utils.Deref(inputData.Streams), meta.Path.AtName("streams_json"), diags),
				VarsJson:    utils.MapToNormalizedType(utils.Deref(inputData.Vars), meta.Path.AtName("vars_json"), diags),
			}
		})
	if newInputs == nil {
		model.Input = types.ListNull(getInputType())
	} else {
		oldInputs := utils.ListTypeAs[integrationPolicyInputModel](ctx, model.Input, path.Root("input"), diags)
		sortInputs(newInputs, oldInputs)

		inputList, d := types.ListValueFrom(ctx, getInputType(), newInputs)
		diags.Append(d...)

		model.Input = inputList
	}
```

We don't need the extra block here. IMO it would be nice to just extract this block into it's own function.

internal/fleet/integration_policy/secrets.go
```diff
@@ -0,0 +1,165 @@
...
+
+// newSecretStore creates a new secretStore from the resource privateData.
+func newSecretStore(ctx context.Context, private privateData) (store secretStore, diags diag.Diagnostics) {
+	bytes, diags := private.GetKey(ctx, "secrets")
+	if diags != nil {
+		return
+	}
+	if bytes == nil {
+		bytes = []byte("{}")
+	}
```
tobio:
```suggestion
	if bytes == nil {
		return
	}
```

Can we just return directly here instead?

internal/fleet/integration_policy/secrets.go
```diff
@@ -0,0 +1,165 @@
...
+	for _, r := range utils.Deref(resp.SecretReferences) {
+		refs[*r.Id] = nil
+	}
+
+	for id := range secrets {
+		if _, ok := refs[id]; !ok {
+			delete(secrets, id)
+		}
+	}
+}
```
tobio: I'd probably prefer returning a new `secretStore` containing only the refs in the response rather than manipulating the existing map. WDYT? 

```suggestion
func (secrets secretStore) pruneRefsFromResponse(resp *fleetapi.PackagePolicy) secretStore {
	refs := make(map[string]any)
	for _, r := range utils.Deref(resp.SecretReferences) {
		refs[*r.Id] = nil
	}

	newSecrets := secretStore{}
	for id, secret := range secrets {
		if _, ok := refs[id]; ok {
			newSecrets[id] = secret
		}
	}
}
```

internal/fleet/output/models.go
```diff
@@ -292,41 +292,6 @@ func (model outputModel) toAPICreateModel(ctx context.Context) (union fleetapi.O
 			diags.AddError(err.Error(), "")
 		}
 
-	case "kafka":
```
tobio: What's the plan here?

tobio (commented):
Can we split this PR into one for the integration_policy resource and one for the output resource please?

generated/fleet/getschema.go
daemitus: most of the fleet changes in getschema (including secret_references) are still missing on the main branch. i was playing around with it yesterday, seeing what would need doing.

internal/fleet/output/models.go
daemitus: gone

internal/fleet/integration_policy/models.go
daemitus: extracted

internal/fleet/integration_policy/secrets.go
daemitus: We can init an empty map and return, sure.

generated/fleet/getschema.go
daemitus: I can look into bumping the spec once this is done. I wanted to incorporate some of the transforms upstream anyways. Based on what I've seen, its likely going to require some extra work on getschema anyways.

internal/fleet/integration_policy/secrets.go
daemitus: since the entire process is only called once each CRUD method, I'll simplify it by pulling in the entire load/save mess into the handle methods. The resp will be available there as well, so we can merge the secrets filter directly into the contructor at that point. 

Ultimately, all read/create/update will see, is `diags = handleReqRespSecrets(ctx, body, policy, resp.Private)` and the rest will all be encapsulated.

tobio (approved):
This looks great, thanks for making this change!

Issue #800, opened 2024-09-27
[Bug] synthetics http/tcp monitor produces inconsistent result after apply


biscout42:
**Describe the bug**

For a `http` monitor there are fields populated not as they are set in tf config. ES stack: `8.15.1`

```
produced an unexpected new value: 

.locations[0]: was cty.StringVal("us_east"), but now cty.StringVal("North America - US East").
.http.proxy_url: was null, but now cty.StringVal("")
.http.ssl_supported_protocols: was null, but now cty.ListVal([]cty.Value{cty.StringVal("TLSv1.1"), cty.StringVal("TLSv1.2"), cty.StringVal("TLSv1.3")}).
.http.ssl_verification_mode: was null, but now cty.StringVal("full").

```

biscout42:
Same for `tcp`

GuillermoRosello:
Similar issue with 

- ES Stack 8.14.3
- Provider version v0.11.8
- Terraform v1.5.3

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # elasticstack_kibana_synthetics_monitor.test_uptime is tainted, so must be replaced
-/+ resource "elasticstack_kibana_synthetics_monitor" "test_uptime" {
      ~ id        = "ops/7c31be2c-6cdc-4d7a-a8d7-06e1e1bf2aca" -> (known after apply)
      ~ locations = [
          - "us-west1-a",
          - "us-east4-a",
          + "us_west",
          + "us_east",
        ]
        name      = "test Uptime: enlightedinc.com"
        tags      = [
            "service:test",
            "terraform:yes",
        ]
        # (6 unchanged attributes hidden)
    }

Plan: 1 to add, 0 to change, 1 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

elasticstack_kibana_synthetics_monitor.test_uptime: Destroying... [id=ops/7c31be2c-6cdc-4d7a-a8d7-06e1e1bf2aca]
elasticstack_kibana_synthetics_monitor.test_uptime: Destruction complete after 2s
elasticstack_kibana_synthetics_monitor.test_uptime: Creating...
╷
│ Error: Provider produced inconsistent result after apply
│ 
[...]

PR #801, opened 2024-09-27
fix for http/tcp monitor produces inconsistent result after apply

biscout42:
## fix: synthetics http monitor creations produced inconsistent result after apply

for https://github.com/elastic/terraform-provider-elasticstack/issues/800

fix for the fileds: http.locations, http.proxy_url, http.ssl_supported_protocols, http.ssl_verification_mode, tcp.proxy_url, tcp.ssl_supported_protocols, tcp.ssl_verification_mode


internal/kibana/synthetics/schema.go
```diff
@@ -55,13 +58,13 @@ type tfHTTPMonitorFieldsV0 struct {
...
-	Host                  types.String   `tfsdk:"host"`
-	SslVerificationMode   types.String   `tfsdk:"ssl_verification_mode"`
-	SslSupportedProtocols []types.String `tfsdk:"ssl_supported_protocols"`
-	CheckSend             types.String   `tfsdk:"check_send"`
-	CheckReceive          types.String   `tfsdk:"check_receive"`
-	ProxyURL              types.String   `tfsdk:"proxy_url"`
-	ProxyUseLocalResolver types.Bool     `tfsdk:"proxy_use_local_resolver"`
+	Host                  types.String `tfsdk:"host"`
+	SslVerificationMode   types.String `tfsdk:"ssl_verification_mode"`
+	SslSupportedProtocols types.List   `tfsdk:"ssl_supported_protocols"`
```
devon-kim, line 64: This type change for the field potentially breaks existing users of the Terrafor module. Are these resources marked such that we can change this? I'm going to presume that we can do this specifically because anyone that used these affected resources before have inconsistent states anyway and will need to update to this version.

tobio: Both these types would be represented as a json list of strings in state either way so the change should be compatible.

@biscout42 iiuc fully specifying these fields in the TF module would avoid the error. Are we able to verify state compatibility in an acceptance test which first creates the resource with provider v0.11.7 and the reapplies with the current version ([example](https://github.com/elastic/terraform-provider-elasticstack/blob/b6434baddcfd07bec06653f380d4795e36bf99df/internal/elasticsearch/index/index/acc_test.go#L80))

internal/kibana/synthetics/schema.go
```diff
@@ -435,6 +450,19 @@ func StringSliceValue(v []string) []types.String {
 	return res
 }
 
+func StringSliceToListAttrValue(v []string) (types.List, diag.Diagnostics) {
```
tobio: https://github.com/elastic/terraform-provider-elasticstack/blob/b6434baddcfd07bec06653f380d4795e36bf99df/internal/utils/tfsdk.go#L14

internal/kibana/synthetics/schema.go
biscout42: @devon-kim , this type change doesn't affect schema. There are TF acceptance tests, which show that config for ssl protocols list stays the same.

internal/kibana/synthetics/schema.go
biscout42: Nice idea, thanks. Let me try.

biscout42: done

tobio (approved):

PR #812, opened 2024-09-30
Add URL support to data_view field_formats

kjwardy:
This PR adds support for the `urlTemplate` and `labelTemplate` attributes when specifying a field in a Data View as type URL, as well as adding functionality to account for when any available `field_formats.params.attributes` don't exist.

By specifying that all fields in `apiFieldFormatParams` are optional, other custom field format types (colour, key/value, etc.) can be added much easier.

### DIFF
--- a/internal/kibana/data_view/schema.go
+++ b/internal/kibana/data_view/schema.go
@@ -147,6 +147,12 @@ func getSchema() schema.Schema {
 										"pattern": schema.StringAttribute{
 											Optional: true,
 										},
+										"urltemplate": schema.StringAttribute{
+											Optional: true,
+										},
+										"labeltemplate": schema.StringAttribute{
+											Optional: true,
+										},
 									},
 								},
 							},
@@ -334,8 +340,15 @@ func dataViewFromResponse(resp data_views.DataViewResponseObjectDataView) apiDat
 
 		if params, ok := formatMap["params"]; ok {
 			if paramsMap, ok := params.(map[string]interface{}); ok {
+				fieldFormatParams := apiFieldFormatParams{}
 				if pattern, ok := paramsMap["pattern"]; ok {
-					apiFormat.Params = &apiFieldFormatParams{Pattern: pattern.(string)}
+					fieldFormatParams.Pattern = pattern.(string)
+				}
+				if urltemplate, ok := paramsMap["urlTemplate"]; ok {
+					fieldFormatParams.Urltemplate = urltemplate.(string)
+				}
+				if labeltemplate, ok := paramsMap["labelTemplate"]; ok {
+					fieldFormatParams.Labeltemplate = labeltemplate.(string)
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


internal/kibana/data_view/schema.go
```diff
@@ -334,8 +340,15 @@ func dataViewFromResponse(resp data_views.DataViewResponseObjectDataView) apiDat
 
 		if params, ok := formatMap["params"]; ok {
 			if paramsMap, ok := params.(map[string]interface{}); ok {
+				fieldFormatParams := apiFieldFormatParams{}
```
tobio: You're never setting `apiFormat.Params = &fieldFormatParams` here and so the params are always nil (this is causing the current test failures).

internal/kibana/data_view/schema.go
```diff
@@ -658,9 +673,13 @@ type apiFieldFormat struct {
...
+	Pattern       types.String `tfsdk:"pattern"`
+	Urltemplate   types.String `tfsdk:"urltemplate"`
+	Labeltemplate types.String `tfsdk:"labeltemplate"`
 }
 
 type apiFieldFormatParams struct {
-	Pattern string `tfsdk:"pattern" json:"pattern"`
+	Pattern       string `tfsdk:"pattern" json:"pattern,omitempty"`
+	Urltemplate   string `tfsdk:"urltemplate" json:"urlTemplate,omitempty"`
+	Labeltemplate string `tfsdk:"labeltemplate" json:"labelTemplate,omitempty"`
```
tobio: These should potentially be `*string` since they're optional in the schema.

internal/kibana/data_view/schema.go
```diff
@@ -658,9 +673,13 @@ type apiFieldFormat struct {
 }
 
 type tfFieldFormatParamsV0 struct {
-	Pattern types.String `tfsdk:"pattern"`
+	Pattern       types.String `tfsdk:"pattern"`
+	Urltemplate   types.String `tfsdk:"urltemplate"`
+	Labeltemplate types.String `tfsdk:"labeltemplate"`
```
tobio:
```suggestion
	UrlTemplate   types.String `tfsdk:"urltemplate"`
	LabelTemplate types.String `tfsdk:"labeltemplate"`
```

internal/kibana/data_view/schema.go
tobio:
```suggestion
	UrlTemplate   string `tfsdk:"urltemplate" json:"urlTemplate,omitempty"`
	LabelTemplate string `tfsdk:"labeltemplate" json:"labelTemplate,omitempty"`
```

kjwardy: That's exactly what I was missing! Thanks for the assist. Changes made and pushed.

tobio (approved):
LGTM. Thanks for taking the time to add this in to the provider!

PR #821, opened 2024-10-02
fix package policy secrets

daemitus:
I migrated work to 0.11.8, and any secret refs were removed in the state after apply.

During testing, it looked like responses either came in the format of 
```{"k1": "v1", ...}``` or ```{"k1": {"type": "string", "value": "v2"}, ...}```. Any time a secretRef was seen, it was inside the latter format. Apparently they can also show up in the former. This fixes the oversight, unit test didn't catch it :(

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## [Unreleased]
 
+- Fix secret handling `elasticstack_fleet_integration_policy` resource. ([#821](https://github.com/elastic/terraform-provider-elasticstack/pull/821))
+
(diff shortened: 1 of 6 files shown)


tobio (commented):
Can we add some unit tests around these two functions to capture both these formats?

daemitus:
can do

daemitus:
I'm not even sure if all these test combinations will even show up, but they pass :D Even found an undesirable edge case that pops up with imports/0.11.8 migrations and unknown secrets.

tobio (approved):
Thanks for following up with this :)

PR #825, opened 2024-10-04
Add aliases to transform resource

### LINKED ISSUE #817: [Bug] Missing field `aliases` for `elasticstack_elasticsearch_transform` resource
The resource `elasticstack_elasticsearch_transform` is missing a field `aliases` which is possible via the transform API https://www.elastic.co/guide/en/elasticsearch/reference/current/put-transform.html

**Versions:**
 - Provider version [e.g. v0.11.7]
 - Elasticsearch Version [e.g. 8.15.0]

ottramst:
Add `aliases` to `elasticstack_elasticsearch_transform` resource.

Resolves #817


internal/elasticsearch/transform/transform.go
```diff
@@ -789,18 +829,18 @@ func flattenSync(sync *models.TransformSync) []interface{} {
...
 	}
 
 	if sync.Time.Field != "" {
-		time["field"] = sync.Time.Field
+		t["field"] = sync.Time.Field
 	}
 
 	s := make(map[string]interface{})
-	s["time"] = []interface{}{time}
+	s["t"] = []interface{}{t}
```
tobio:
```suggestion
	s["time"] = []interface{}{t}
```

The underlying schema isn't changing here.

tobio (commented):
Hey thanks for making both of these additions. They're great updates!

internal/elasticsearch/transform/transform.go
```diff
@@ -810,18 +850,18 @@ func flattenRetentionPolicy(retention *models.TransformRetentionPolicy) []interf
...
 	}
 
 	if retention.Time.Field != "" {
-		time["field"] = retention.Time.Field
+		t["field"] = retention.Time.Field
 	}
 
 	r := make(map[string]interface{})
-	r["time"] = []interface{}{time}
+	r["t"] = []interface{}{t}
```
tobio:
```suggestion
	r["time"] = []interface{}{t}
```

Similarly

tobio:
It looks like the committed docs don't match the generated output. You should be able to just run `make docs-generate` and commit the result.

ottramst:
Updated documentation as well :)

tobio (approved):
Nice one! Thanks for taking the time with both these contributions.

PR #828, opened 2024-10-04
feat: Add `elasticsearch_index_template` data source

fatmcgav:
This commit adds a new data source `elasticsearch_index_template`.


fatmcgav:
Hmm, I'm confused why the integration tests are failing for new ds?

Unless the `ilm-history-7` template doesn't exist in earlier versions maybe?

fatmcgav:
Right, I've switched to checking the `logs` template, as that seems to exist in both `v7 `and `v8`. 🤞

tobio:
@fatmcgav maybe it would make sense for the acceptance test to first create a template, and then check that the data source can retrieve it instead of relying on default templates being present. WDYT?

tobio:
Are you able to add an entry to the [CHANGELOG](https://github.com/elastic/terraform-provider-elasticstack/blob/main/CHANGELOG.md) for this one too please?

tobio (commented):
LGTM pending the changelog and acc test improvements.

fatmcgav:
Thanks @tobio... Tweaks pushed 👍

fatmcgav:
Damn, something it still doesn't like there 😢 ... I'm struggling to run the integration tests locally due to environmental issues... will keep plugging away though...

tobio:
@fatmcgav I believe https://github.com/elastic/terraform-provider-elasticstack/pull/828/commits/9e0f7da905f2c28b99ea1b0459c1425532fe77fa should fix the tests. Debugging these locally had the data source being executed before the template had been created.

fatmcgav:
> @fatmcgav I believe [9e0f7da](https://github.com/elastic/terraform-provider-elasticstack/commit/9e0f7da905f2c28b99ea1b0459c1425532fe77fa) should fix the tests. Debugging these locally had the data source being executed before the template had been created.

Ahh, that makes sense... Thanks for pushing the fix there...

tobio (approved):
Thanks for adding this!

fatmcgav:
@tobio Could I be cheeky and ask for a release? This ds will be useful in some work we're currently undertaking to manage all our ILM policies and index templates via Terraform... And with Kubernetes there's _quite a few_ index templates that need to be copied for https://www.elastic.co/guide/en/fleet/current/data-streams-ilm-tutorial.html

👍

tobio:
> Could I be cheeky and ask for a release?

Not cheeky at all :) We'll aim to get a release out this week. 

I'm not a big fan of leaving community contributions sitting unreleased for ages, you're generally not doing it just for fun.

PR #841, opened 2024-10-12
Fix output inconsistent result (default_integrations/monitoring)

### LINKED ISSUE #840: [Bug] elasticstack_fleet_output inconsistent default_integrations and default_monitoring
```
╷
│
│ When applying changes to elasticstack_fleet_output.test1, provider "provider[\"registry.terraform.io/elastic/elasticstack\"]" produced an unexpected new value: .default_integrations: was null, but now cty.False.
│
╵
╷
│
│ When applying changes to elasticstack_fleet_output.test1, provider "provider[\"registry.terraform.io/elastic/elasticstack\"]" produced an unexpected new value: .default_monitoring: was null, but now cty.False.
│
╵
```

daemitus:
Fixes #840 
Add default: False to default_integrations and default_monitoring.

This was not found by the tests because both fields are set in each test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Changes
 
+- Fix inconsistent output errors in `elasticstack_fleet_output` for `default_integrations` and `default_monitoring`. ([#841](https://github.com/elastic/terraform-provider-elasticstack/pull/841))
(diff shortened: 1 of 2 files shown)


tobio (commented):
Can we get a changelog entry for this one please?

daemitus:
Sure, I'll get it later tonight

tobio (approved):

Issue #793, opened 2024-09-24
[Bug] Changing `role_descriptors` in `elasticstack_elasticsearch_security_api_key` forces


Davasny:
**Describe the bug**
Changing value of `role_descriptors` forces api key replacement.
Editing api key roles via kibana doesn't change the api key value.

**To Reproduce**

1. Create api key with some role descriptor

```hcl
resource "elasticstack_elasticsearch_security_api_key" "test" {
  name = "test"

  role_descriptors = jsonencode({
    read_only = {
      indices = [
        {
          names = ["*"],
          privileges = ["read", "view_index_metadata"]
          allow_restricted_indices = false
        }
      ]
    }
  })
}
```

2. `terraform apply`

3. Change this role to anything else, I removed `view_index_metadata` in the following example

```hcl
resource "elasticstack_elasticsearch_security_api_key" "test" {
  name = "test"

  role_descriptors = jsonencode({
    read_only = {
      indices = [
        {
          names = ["*"],
          privileges = ["read"]
          allow_restricted_indices = false
        }
      ]
    }
  })
}
```

4. `terraform apply` will show "forces replacement"

```
Terraform will perform the following actions:

  # elasticstack_elasticsearch_security_api_key.test must be replaced
-/+ resource "elasticstack_elasticsearch_security_api_key" "test" {
      ~ api_key              = (sensitive value)
      ~ encoded              = (sensitive value)
      ~ expiration_timestamp = 0 -> (known after apply)
      ~ id                   = "BqFsxqFzRn--CUY9V3wk7w/FlCVI5IBVbU1E1HFWlqk" -> (known after apply)
      ~ metadata             = jsonencode({}) -> (known after apply)
        name                 = "test"
      ~ role_descriptors     = jsonencode(
          ~ {
              ~ read_only = {
                  ~ indices = [
                      ~ {
                          ~ privileges               = [
                                "read",
                              - "view_index_metadata",
                            ]
                            # (2 unchanged attributes hidden)
                        },
                    ]
                }
            } # forces replacement
        )
        # (1 unchanged attribute hidden)
    }

Plan: 1 to add, 0 to change, 1 to destroy.

```

**Expected behavior**
I expect terraform module to only update role descriptor the same way as kibana does.

**Debug output**
N/A

**Screenshots**
N/A

**Versions (please complete the following information):**

- OS: macos 15
- terraform 1.8.0
- provider registry.terraform.io/elastic/elasticstack v0.11.7
[...]

PR #843, opened 2024-10-14
Support updating API keys

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/793

Migrates the API Key resource to the plugin framework. This is required to support conditionally setting `RequiresReplace` during the planning phase based on whether the target cluster supports updating API keys.


internal/clients/elasticsearch/security.go
```diff
@@ -292,83 +293,101 @@ func DeleteRoleMapping(ctx context.Context, apiClient *clients.ApiClient, roleMa
...
 }
 
-func GetApiKey(apiClient *clients.ApiClient, id string) (*models.ApiKeyResponse, diag.Diagnostics) {
-	var diags diag.Diagnostics
+func UpdateApiKey(apiClient *clients.ApiClient, apikey models.ApiKey) fwdiag.Diagnostics {
+	id := apikey.ID
+
+	apikey.Expiration = ""
+	apikey.Name = ""
+	apikey.ID = ""
```
dimuon, line 329: Does it mean that both `Name` and `Expiration` cannot be updated after creation?

tobio: Yea, correct. They're marked as RequiresReplace in the schema as well.

dimuon (approved):

PR #845, opened 2024-10-15
Add `max_primary_shard_docs` condition to ILM rollover

### LINKED ISSUE #794: [Bug] max_primary_shard_docs variable is not supported in elasticstack_elasticsearch_index_lifecycle.hot.rollover
**Describe the bug**
`max_primary_shard_docs` variable is not supported in elasticstack_elasticsearch_index_lifecycle.hot.rollover.

**To Reproduce**
1. In resource `elasticstack_elasticsearch_index_lifecycle`, trying to add `hot.roller.max_primary_shard_docs` configurations.
2. See the error that this attribute is not support in the provider, but do support in lifecycle API and UI.

**Expected behavior**
Support `max_primary_shard_docs` variable in Nested Schema for `hot.rollover`

**Debug output**
**Screenshots**

 - OS: [e.g. Linux]
 - Terraform Version [1.3.6]
 - Provider version [v0.11.6]
 - Elasticsearch Version [8.14.3]

**Additional context**
https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-rollover.html

arenard:
* Add `max_primary_shard_docs` condition to rollover (closes  #794)

* Update test for rollover `max_primary_shard_docs` condition

* Specify min version in the description

* Update CHANGELOG.md

### DIFF
--- a/internal/elasticsearch/index/ilm.go
+++ b/internal/elasticsearch/index/ilm.go
@@ -567,7 +567,11 @@ func expandPhase(p map[string]interface{}, serverVersion *version.Version) (*mod
 	return &phase, diags
 }
 
-var RolloverMinConditionsMinSupportedVersion = version.Must(version.NewVersion("8.4.0"))
+var (
+	RolloverMinConditionsMinSupportedVersion = version.Must(version.NewVersion("8.4.0"))
+	MaxPrimaryShardDocsMinSupportedVersion   = version.Must(version.NewVersion("8.2.0"))
+)
+
 var ilmActionSettingOptions = map[string]struct {
 	skipEmptyCheck bool
 	def            interface{}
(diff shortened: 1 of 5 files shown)


internal/elasticsearch/index/ilm.go
```diff
@@ -568,15 +573,16 @@ var ilmActionSettingOptions = map[string]struct {
 	def            interface{}
 	minVersion     *version.Version
 }{
+	"allow_write_after_shrink": {def: false, minVersion: version.Must(version.NewVersion("8.14.0"))},
 	"number_of_replicas":       {skipEmptyCheck: true},
-	"total_shards_per_node":    {skipEmptyCheck: true, def: -1, minVersion: version.Must(version.NewVersion("7.16.0"))},
 	"priority":                 {skipEmptyCheck: true},
+	"max_primary_shard_docs":   {minVersion: version.Must(version.NewVersion("8.2.0"))},
```
arenard: I am not sure about what should be put there as this property is optional. Maybe a skipEmptyCheck should be added there.

tobio: I think this should be fine, `skipEmptyCheck` is around dealing with how the TF SDK represents unset attributes as Golang zero values.

internal/elasticsearch/index/ilm_test.go
```diff
@@ -273,12 +274,13 @@ resource "elasticstack_elasticsearch_index_lifecycle" "test_rollover" {
       max_age = "7d"
       max_docs = 10000
       max_size = "100gb"
+      max_primary_shard_docs = 5000
```
tobio, line 285: We'll want to update the tests to only specify this when it's supported by the backing cluster. That means adding a case covering this attribute which is skipped when it's not supported ([example](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/index/ilm_test.go#L66-L79))

tobio (commented):
Thanks for adding this in, we'll want to update the tests to only cover this attribute when it's supported.

tobio (approved):

Issue #829, opened 2024-10-04
[Bug] Kibana alert action "custom action interval" not working with unit "d" = days in frequency "throttle" settings


ASH895-N:
**Describe the bug**
When using "d" as a unit of time for "custom action interval" that is configured in Terraform via "throttle" in "frequency" field, the terraform plan is failing (here: https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/kibana_alerting_rule#nested-schema-for-actionsfrequency) 

But when I update to use "m" for "minutes" or "h" for "hours", it is working. 

Notes: Also works when I import existing alert created from Kibana UI to Terraform. The below configuration was first taken from an alert created with UI and then imported to Terraform. After that, I removed fields so I can test alert creation. 

**To Reproduce**

Terraform configuration :  **this is to create the alert** 

```
resource "elasticstack_kibana_alerting_rule" "custom_intervals" {
    consumer              = "stackAlerts"
    enabled               = true
    interval              = "1m"
    name                  = "test-custom-interval-tf"
    params                = jsonencode(
        ....
    )
    rule_type_id          = ".es-query"
    space_id              = "default"

    actions {
        group  = "query matched"
        id     = "xxxxxx"
        params = jsonencode(
            ....
        )
        ...

        frequency {
            notify_when = "onThrottleInterval"
            summary     = false
            throttle    = "1d" ------> this setting
        }
    }
    actions {
        group  = "query matched"
        id     = "xxxxxxxx"
        params = ...

        frequency {
            notify_when = "onActionGroupChange"
            summary     = false
        }
    }
}
```

**Expected behavior**
Per documentation here - https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/kibana_alerting_rule#throttle , this should be a valid value. When I imported an alert rule that was created with UI, it gave me the same settings . This is from `terraform state show`

```
  frequency {
            notify_when = "onThrottleInterval"
            summary     = false
            throttle    = "1d"
        }
```

Also referred to : https://github.com/elastic/kibana/blob/7a8d328cfa0f59516124a760710d0d5831680abf/x-pack/plugins/alerting/docs/openapi/components/schemas/throttle.yaml 

**Debug output**
I'm observing this error : 

[...]

Issue #770, opened 2024-09-13
[Bug] elasticstack_kibana_alerting_rule interval doesn't accept days


reakaleek:
**Describe the bug**
Using days in the `elasticstack_kibana_alerting_rule` `interval` option leads to an error.

**To Reproduce**

I imported an existing Kibana alert rule where the check interval is  7 days using `terraform import`

When run `terraform plan`

I get an expected diff with:

```
...
# elasticstack_kibana_alerting_rule.test-reporter will be updated in-place
  ~ resource "elasticstack_kibana_alerting_rule" "test-reporter" {
        id                    = "default/b6679cfe-e09a-4114-82ba-f1684563ee4a"
      ~ interval              = "7d" -> "1h"
...
```

But when I change the `interval` to `"7d"` in the tf file, I get the following error.

```
╷
│ Error: "interval" contains an invalid duration: time: unknown unit "d" in duration "7d"
│ 
│   with elasticstack_kibana_alerting_rule.test-reporter,
│   on 02-kibana-alert-rules.tf line 55, in resource "elasticstack_kibana_alerting_rule" "test-reporter":
│   55:   interval     = "7d"
│ 
╵
```

**Expected behavior**
I should be able to plan or apply with the interval `"7d"`

**Debug output**
Run `terraform` command with `TF_LOG=trace` and provide extended information on TF operations. **Please ensure you redact any base64 encoded credentials from your output**. 
eg
```
[DEBUG] provider.terraform-provider-elasticstack_v0.11.0: Authorization: Basic xxxxx==
```

**Screenshots**

**Versions (please complete the following information):**
 - OS: macos 14.6.1
 - Terraform Version: v1.9.3
 - Provider version 0.11.6
 - Elasticsearch Version 8.15.1

**Additional context**

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/829

PR #846, opened 2024-10-15
Alerting fields should be Elastic durations, not Golang durations

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/829
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/770

Updates validation to check attributes are Elastic durations, not Golang durations.

### DIFF
--- a/internal/kibana/alerting.go
+++ b/internal/kibana/alerting.go
@@ -26,7 +26,7 @@ var alertDelayMinSupportedVersion = version.Must(version.NewVersion("8.13.0"))
 //
 //nolint:staticcheck
 func stringIsAlertingDuration() schema.SchemaValidateFunc {
-	r := regexp.MustCompile(`^[0-9]+(?:\.[0-9]+)?(?:d|h|m|s)$`)
+	r := regexp.MustCompile(`^[1-9][0-9]*(?:d|h|m|s)$`)
 	return validation.StringMatch(r, "string is not a valid Alerting duration in seconds (s), minutes (m), hours (h), or days (d)")
 }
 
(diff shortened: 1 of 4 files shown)


adcoelho (dismissed):
LGTM but the regex in `utils.StringIsElasticDuration` allows a bit more than the API.

`^[0-9]+(?:\.[0-9]+)?(?:d|h|m|s|ms|micros|nanos)$` vs only seconds/minutes/hours/days in the Rules API.

internal/kibana/alerting.go
```diff
@@ -21,6 +22,14 @@ var frequencyMinSupportedVersion = version.Must(version.NewVersion("8.6.0"))
 var alertsFilterMinSupportedVersion = version.Must(version.NewVersion("8.9.0"))
 var alertDelayMinSupportedVersion = version.Must(version.NewVersion("8.13.0"))
 
+// Avoid lint error on deprecated SchemaValidateFunc usage.
+//
+//nolint:staticcheck
+func stringIsAlertingDuration() schema.SchemaValidateFunc {
+	r := regexp.MustCompile(`^[0-9]+(?:\.[0-9]+)?(?:d|h|m|s)$`)
```
adcoelho: Thanks for the change!

If I'm interpreting the regex correctly this block `(?:\.[0-9]+)?` is for decimals? That won't be necessary, the suggestion below [is more in tune with what we use as validation in our API](x-pack/plugins/alerting/common/routes/rule/validation/validate_duration/v1.ts).

```suggestion
	r := regexp.MustCompile(`^[1-9][0-9]*(?:d|h|m|s)$`)
```

PR #852, opened 2024-10-17
add resource for datastream lifecycle

### LINKED ISSUE #838: [Feature] Datastream lifecycle
In my previous PR #724, I added support for the DS template lifecycle. However, this is insufficient to update the existing DS lifecycle.

https://www.elastic.co/guide/en/elasticsearch/reference/current/tutorial-manage-existing-data-stream.html

I can implement new resources for these APIs

https://www.elastic.co/guide/en/elasticsearch/reference/current/data-streams-get-lifecycle.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/data-streams-put-lifecycle.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/data-streams-delete-lifecycle.html

usage would be:

```
resource "elasticstack_elasticsearch_data_stream_lifecycle" "ds_lifecycle" {
  name = "ds-name"
  data_retention = "5d"
  enabled = true
  downsampling=[...]
}
```

Is it OK for you to do it this way?

aizerin:
resolves #838


internal/elasticsearch/index/data_stream_lifecycle_test.go
```diff
@@ -0,0 +1,239 @@
...
+	sdkacctest "github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestAccResourceDataStreamLifecycle(t *testing.T) {
+	// generate renadom name
+	dsName := sdkacctest.RandStringFromCharSet(22, sdkacctest.CharSetAlpha)
+
+	resource.Test(t, resource.TestCase{
```
tobio: We'll want to skip these when the backing cluster doesn't support DS lifecycles ([example](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/kibana/alerting_test.go#L34))

internal/elasticsearch/index/data_stream_lifecycle_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/elastic/terraform-provider-elasticstack/internal/acctest"
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
+	"github.com/elastic/terraform-provider-elasticstack/internal/models"
+	sdkacctest "github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestAccResourceDataStreamLifecycle(t *testing.T) {
+	// generate renadom name
```
tobio:
```suggestion
	// generate random name
```

templates/resources/elasticsearch_data_stream_lifecycle.md.tmpl
```diff
@@ -0,0 +1,17 @@
...
+
+# Resource: elasticstack_elasticsearch_data_stream
+
+Configures the data stream lifecycle for the targeted data streams, see: https://www.elastic.co/guide/en/elasticsearch/reference/current/data-stream-apis.html
+
+## Example Usage
+
+{{ tffile "examples/resources/elasticstack_elasticsearch_data_stream_lifecycle/resource.tf" }}
+
+{{ .SchemaMarkdown | trimspace }}
```
tobio: Can we include an [import example](https://github.com/elastic/terraform-provider-elasticstack/blob/main/templates/resources/elasticsearch_data_stream.md.tmpl#L19-L23) here as well

internal/elasticsearch/index/data_stream_lifecycle.go
```diff
@@ -0,0 +1,232 @@
...
+					},
+					"fixed_interval": {
+						Description: "The interval at which to aggregate the original time series index. For example, 60m produces a document for each 60 minute (hourly) interval. This follows standard time formatting syntax as used elsewhere in Elasticsearch.",
+						Type:        schema.TypeString,
+						Required:    true,
+					},
+				},
+			},
+		},
+		"lifecycles": {
```
tobio: It's not clear what value we get with this block. 

Since we're not writing these values back to the top level attributes TF drift detection won't pickup out of band changes. Choosing which value to use when the response includes multiple lifecycles is a challenge, but something we should solve IMO (a simple heuristic would be to use any value which differed from the configured value so drift detection worked, and if no diff was found just use the first). 

In reality, I suspect most users will likely target a single data stream.

tobio (commented):
Thanks for taking the time to add this in. 

Is it possible to have this resource built on the Plugin Framework instead? There's a strong preference here for new code to utilise that framework over the old SDK v2 and we're not re-using any existing SDK code here. [This](https://github.com/elastic/terraform-provider-elasticstack/pull/843/commits/241ebe667bb583a5091de7b931546e574b66f37f) commit is an example of converting to the plugin framework if that helps.

It looks like lint is failing because the generated docs don't match what's committed, which can be fixed by committing the result of `make docs-generate`

aizerin:
> Thanks for taking the time to add this in.
> 
> Is it possible to have this resource built on the Plugin Framework instead? There's a strong preference here for new code to utilise that framework over the old SDK v2 and we're not re-using any existing SDK code here. [This](https://github.com/elastic/terraform-provider-elasticstack/pull/843/commits/241ebe667bb583a5091de7b931546e574b66f37f) commit is an example of converting to the plugin framework if that helps.
> 
> It looks like lint is failing because the generated docs don't match what's committed, which can be fixed by committing the result of `make docs-generate`

ok, not a problem.

internal/elasticsearch/index/data_stream_lifecycle.go
aizerin: Originally I had it so that I only detected the value that changed, then I thought it would be nice to have everything in state, but I didn't know that TF couldn't work with it that way.

I will change it back :)

aizerin:
updated

tobio (commented):
Thanks for adding this resource, it's really useful! 

One small comment on de-duplicating the `Create`/`Update` implementations but otherwise LGTM.

internal/elasticsearch/index/data_stream_lifecycle/update.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients/elasticsearch"
+	"github.com/elastic/terraform-provider-elasticstack/internal/utils"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func (r *Resource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
```
tobio, line 9: This looks identical to the `Create` method. Can we pull this out to a common `write` (or some other name if you prefer) method and re-use it in both `Create` and `Update`?

aizerin: OK, I did the same as here:

internal/fleet/integration/create.go
internal/fleet/integration/update.go

tobio (approved):
Love it, thanks again for taking the time to get this one added, and for working through the review changes.

PR #854, opened 2024-10-17
EPM fixes

daemitus:
So there's a number of problems that cropped up.

- When an integration is uninstalled outside of terraform, the change is not picked up. A 404 is not returned. Instead, a 200 item.status="not_installed" is. Requiring a switch to ReadPackageWithResponse and updating getschema (again).
- The schema was wrong (again). The entirety of `allOf.1` needs to be under `.item` which is present in `allOf.0`. All the more reason to get the #834 to a satisfactory state.
- Uninstall does not return a 404 like most other delete methods, but a 400 message="name is not installed". So switch to DeletePackageWithResponse.
- Since there's only one non-WithResponse method left at this point in all of fleet, switch InstallPackage to it as well. Does all the body reading for us.
- fleet_integration create and update pass respDiags by value instead of pointer, so if a diag is actually added, it doesn't end up back in the caller. 
Actual:
```
    resource_test.go:76: Step 1/2 error: Error running apply: exit status 1
        Error: Missing Resource State After Create
```
Expected (using ti_abusech on one of the older stack versions):
```
    resource_test.go:76: Step 1/2 error: Error running apply: exit status 1
        Error: Unexpected status code from server: got HTTP 500
```

### DIFF
--- a/internal/clients/fleet/fleet.go
+++ b/internal/clients/fleet/fleet.go
@@ -390,7 +390,7 @@ func Uninstall(ctx context.Context, client *Client, name, version string, force 
 	case http.StatusBadRequest:
 		msg := resp.JSON400.Message
 		// {"statusCode":400,"error":"Bad Request","message":"some-integration is not installed"}
-		if msg != nil && strings.Contains(*msg, "is not installed") {
+		if msg != nil && *msg == fmt.Sprintf("%s is not installed", name) {
 			return nil
 		} else {
 			return reportUnknownError(resp.StatusCode(), resp.Body)
--- a/internal/clients/fleet/fleet.go
+++ b/internal/clients/fleet/fleet.go
@@ -389,7 +389,6 @@ func Uninstall(ctx context.Context, client *Client, name, version string, force 
 		return nil
 	case http.StatusBadRequest:
 		msg := resp.JSON400.Message
-		// {"statusCode":400,"error":"Bad Request","message":"some-integration is not installed"}
 		if msg != nil && *msg == fmt.Sprintf("%s is not installed", name) {
 			return nil
 		} else {
--- a/internal/clients/fleet/fleet.go
+++ b/internal/clients/fleet/fleet.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"strings"
 
 	fleetapi "github.com/elastic/terraform-provider-elasticstack/generated/fleet"
 	"github.com/elastic/terraform-provider-elasticstack/internal/utils"
--- a/internal/clients/fleet/fleet.go
+++ b/internal/clients/fleet/fleet.go
@@ -4,8 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"net/http"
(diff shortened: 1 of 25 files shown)


daemitus:
Additionally, I need to check if the diags by value issue is also in all the utils/tfsdk.go helper methods.
```
type ListMeta struct {
	Index int
	Path  path.Path
	Diags diag.Diagnostics
}

type MapMeta struct {
	Key   string
	Path  path.Path
	Diags diag.Diagnostics
}
```

internal/clients/fleet/fleet.go
```diff
@@ -399,6 +387,14 @@ func Uninstall(ctx context.Context, client *Client, name, version string, force
 	switch resp.StatusCode() {
 	case http.StatusOK:
 		return nil
+	case http.StatusBadRequest:
+		msg := resp.JSON400.Message
+		// {"statusCode":400,"error":"Bad Request","message":"some-integration is not installed"}
+		if msg != nil && strings.Contains(*msg, "is not installed") {
```
tobio: IIUC this could be this? I don't actually feel strongly here, just checking my understanding. 

```suggestion
		if msg != nil && *msg == fmt.Sprintf("%s is not installed", name) {
```

internal/fleet/integration/resource_test.go
```diff
@@ -68,6 +72,47 @@ func TestAccResourceIntegration(t *testing.T) {
...
+			},
+			{
+				SkipFunc: versionutils.CheckIfVersionIsUnsupported(minVersionIntegration),
+				Config:   testAccResourceIntegrationDeleted,
+				// Force uninstall the integration
+				PreConfig: func() {
+					client, err := clients.NewAcceptanceTestingClient()
+					if err != nil {
+						panic(err)
+					}
```
tobio:
```suggestion
					require.NoError(t, err)
```

internal/fleet/integration/resource_test.go
```diff
@@ -68,6 +72,47 @@ func TestAccResourceIntegration(t *testing.T) {
...
+					fleetClient, err := client.GetFleetClient()
+					if err != nil {
+						panic(err)
+					}
```
tobio:
```suggestion
					require.NoError(t, err)
```

internal/fleet/integration/resource_test.go
```diff
@@ -68,6 +72,47 @@ func TestAccResourceIntegration(t *testing.T) {
...
+
+					ctx := context.Background()
+					diags := fleet.Uninstall(ctx, fleetClient, "sysmon_linux", "1.7.0", true)
+					if diags.HasError() {
+						panic(utils.FwDiagsAsError(diags))
+					}
```
tobio:
```suggestion
					require.Empty(t, diags)
```

tobio (commented):
👍 couple of minor comments on the code.

internal/clients/fleet/fleet.go
daemitus: agreed, this works better. must not have realized name was already accessible.

tobio (approved):
Love it, 🎉

PR #855, opened 2024-10-17
update tfsdk helpers to use *diags

daemitus:
As mentioned [here](https://github.com/elastic/terraform-provider-elasticstack/pull/854#issuecomment-2419856734), the idea behind the utils/tfsdk.go meta parameters is to pass around response.Diagnostics, the key/index, path, etc. Anything added to meta.diags doesn't actually get returned because it is passed by value and not pointer. This never cropped up in the other tests, mostly because the tests pass, but secondly all the usage is inline, so they reference the outer diags, and not meta.diags.


daemitus:
This is actually causing `TestAccResourceIntegrationPolicySecretsFromSDK` to fail now, due to input.vars_json="" in 0.11.7 when the actual value was null.

An empty string isn't valid JSON and is causing the jsontypes.Normalized vars_json to add an error diagnostic.

```
(basetypes.ListValue) [{"enabled":true,"input_id":"aws_logs-aws-cloudwatch","streams_json":"...","vars_json":""}, ...}]
```

daemitus:
Implemented an upgrade for fleet_integration_policy, Any streams_json or vars_json as empty strings get reassigned as nulls, so the jsontypes.Normalized wont complain anymore.

PR #859, opened 2024-10-18
Fix `alert_delay` update in Update Rule API

adcoelho:
Fixes https://github.com/elastic/kibana/issues/196797

## Description

The `alert_delay` field was being created properly but not updated.


jcger (approved):
Tested locally, working as expected 🎉

tobio (approved):
🎉

Issue #872, opened 2024-10-23
[Bug] SLO resource contains a type mismatch in the optional group_by field


jasonrhodes:
**Describe the bug**
[The SLO resource defines a "group_by" key](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/kibana/slo.go#L411-L416) whose type is set to string, but the API accepts string OR a list of strings. Because of this, if you edit an SLO using the UI or a direct API call outside of terraform and set it to an array of strings, the terraform provider will fail on subsequent runs. 

An example of the error would be:

```
Error: json: cannot unmarshal array into Go struct field SloResponse.groupBy of type string

  with elasticstack_kibana_slo.slo_task_claim_success_rate["all"],
  on slo-task-claim-success-rate.tf line 1, in resource "elasticstack_kibana_slo" "slo_task_claim_success_rate":
   1: resource "elasticstack_kibana_slo" "slo_task_claim_success_rate" ***


Error: json: cannot unmarshal array into Go struct field SloResponse.groupBy of type string
```

I intended to submit a PR to update the type, changing the above code to something like this:

```go
		"group_by": {
			Description: "Optional group by field to use to generate an SLO per distinct value.",
			Type:        schema.TypeList,
			Optional:    true,
			ForceNew:    false,
			Elem: &schema.Schema{
				Type: schema.TypeString,
			},
		},
```

But on reading the docs for schema validation, it doesn't seem to be possible to specify a field that can be EITHER one type OR another, like this one would need to be. I may be missing something, though! 

**To Reproduce**
1. Create an SLO using this provider's SLO resource.
2. Edit the SLO in the UI, manually (you don't need to edit the group by field, necessarily) 
3. Run terraform again, you will see an error similar to the above example

**Expected behavior**
Terraform should handle cases where an SLO has a group_by field set to a single string OR a group_by field set to an array of strings. 

**Debug output**

I don't have access to this, but @jloleysens may be able to provide this. 

**Versions (please complete the following information):**

@jloleysens could you provide these details, also?

 - OS: [e.g. Linux]
 - Terraform Version [e.g. 1.0.0]
 - Provider version [e.g. v0.1.0]
 - Elasticsearch Version [e.g. 7.16.0]

**Additional context**

None at this time.

jasonrhodes:
[...]

PR #878, opened 2024-10-25
Support mutliple group_by fields in SLOs 

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/872

Makes a breaking change to the provider to support multiple group by fields in SLO definitions. Existing SLO's defining `group_by = "field"` will need to be updated to `group_by = ["field"]` when upgrading provider versions.


gigerdo (approved):
LGTM :+1:

Issue #637, opened 2024-05-17
Handle Deprecations: Kibana Saved Objects APIs


TinaHeiligers:
Most of the Kibana Saved Objects APIs were [deprecated](https://github.com/elastic/kibana/pull/150124) in 8.7 (released Mar 30, 2023). 

The APIs that are to remain are:

`/api/saved_objects/_import`
`/api/saved_objects/_export`
`/api/saved_objects/_resolve_import_errors`

The provider will need to [change](const (
	basePathKibanaSavedObject = "/api/saved_objects" // Base URL to access on Kibana save objects
)) the implementation before these APIs are removed.

TinaHeiligers:
cc @jloleysens

tobio:
IIUC the provider is only using the `_import` endpoint now (see [here](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/kibana/import_saved_objects/create.go#L37)).  

Is there something else here that I'm not aware of?

TinaHeiligers:
> IUC the provider is only using the _import endpoint now (see [here](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/kibana/import_saved_objects/create.go?rgh-link-date=2024-05-17T02%3A13%3A36Z#L37)).

Fantastic!

> Is there something else here that I'm not aware of?

No, the `_import` API isn't deprecated and will be supported for the foreseeable future.

@tobio Thanks for the quick response!

TinaHeiligers:
Reopened to discuss the follow-up action to delete the deprecated saved objects APIs.

If we plan to support `terraform-provider-elasticstack`, keeping deprecated APIs around runs the risk of consumers building integrations with APIs that aren't supported anymore, if they even still exist!

We strongly suggest handling the necessary changes a.s.a.p.

Current:
```
type OptionalFindParameters struct {
	ObjectsPerPage        int
	Page                  int
	Search                string
	DefaultSearchOperator string
	SearchFields          []string
	Fields                []string
	SortField             string
	HasReference          string
}

// KibanaSavedObjectGet permit to get saved object from Kibana
type KibanaSavedObjectGet func(objectType string, id string, kibanaSpace string) (map[string]interface{}, error)

// KibanaSavedObjectFind permit to find saved objects from Kibana
type KibanaSavedObjectFind func(objectType string, kibanaSpace string, optionalParameters *OptionalFindParameters) (map[string]interface{}, error)

[...]

PR #948, opened 2024-12-12
Add kibana_saved_object resource

ydubreuil:
This PR implements proper support for managing a Kibana saved object.

Unlike existing resource `kibana_import_saved_objects`, this new resource can only manage one single saved object but it keeps track of the changes, if the resource is changed in the Terraform code or in Kibana, the TF plan will show the difference. It also supports removing the saved object.

Providing support for all CRUD operations is required to expose this resource in a Crossplane provider.

This PR has been tested both with Terraform and with a Crossplane provider that relies on this code.


internal/kibana/saved_object/read.go
```diff
@@ -0,0 +1,45 @@
...
+		return
+	}
+
+	kibanaClient, err := r.client.GetKibanaClient()
+	if err != nil {
+		resp.Diagnostics.AddError("unable to get kibana client", err.Error())
+		return
+	}
+
+	result, err := kibanaClient.KibanaSavedObject.Get(model.Type.ValueString(), model.ID.ValueString(), model.SpaceID.ValueString())
```
tobio, line 25: Reading saved objects has been [deprecated](https://github.com/elastic/terraform-provider-elasticstack/issues/637) since 8.7. It'll be completely removed shortly, we could keep this for versions which support the `Get`, but IMO the UX where drift detection is silently dropped when upgrading the underlying Kibana version is pretty bad.

I think the `Read` should be basically a no-op here, simply maintaining the current state.

internal/kibana/saved_object/schema.go
```diff
@@ -0,0 +1,80 @@
...
+			"space_id": schema.StringAttribute{
+				Description: "An identifier for the space. If space_id is not provided, the default space is used.",
+				Optional:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"object": schema.StringAttribute{
+				Description: "Kibana object to import in JSON format",
+				Required:    true,
```
tobio, line 49:
```suggestion
				Required:    true,
				CustomType: jsontypes.NormalizedType{},
```

The [index](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/index/index/schema.go#L48) resource has an example of using this. It means simply restructuring the JSON won't be seen as a diff.

internal/kibana/saved_object/create.go
```diff
@@ -0,0 +1,51 @@
...
+		return
+	}
+
+	kibanaClient, err := r.client.GetKibanaClient()
+	if err != nil {
+		resp.Diagnostics.AddError("unable to get kibana client", err.Error())
+		return
+	}
+
+	result, err := kibanaClient.KibanaSavedObject.Import([]byte(model.Imported.ValueString()), false, model.SpaceID.ValueString())
```
tobio, line 30:
```suggestion
	result, err := kibanaClient.KibanaSavedObject.Import([]byte(model.Imported.ValueString()), true, model.SpaceID.ValueString())
```

Given the use case here, I wonder if we should just always overwrite existing objects. It would simplify the workflow for bringing existing objects into TF, especially in a resource where a proper `import` handler is kind of tricky given the API constraints.

internal/kibana/saved_object/read.go
ydubreuil: I was not aware of the deprecation of the API. I guess this PR doesn't make such sense then :(

Crossplane relies on the Read to be functional to work. That's the whole point of this PR. Without this, the reconciliation loop can't work.

There's a new API in tech preview to manage dashboards, I think it makes sense to refactor this code to use this new API.

internal/kibana/saved_object/schema.go
ydubreuil: Nice, thanks for pointing that out

internal/kibana/saved_object/create.go
ydubreuil: I wondered about that and initially had overwrite set to `true`. I agree it would simplify importing resources into TF to have it set to `true`.

Issue #933, opened 2024-11-27
[Bug] HTTP 400 Bad Request When Creating elasticstack_kibana_security_role


browol:
**Describe the bug** 

I encountered an error while attempting to create the `elasticstack_kibana_security_role` resource. Although the Kibana role appears to be created successfully, the Terraform resource (`elasticstack_kibana_security_role.role`) fails to be created.  

Upon reviewing the Kibana logs, it seems the issue is caused by a 400 Bad Request response. Specifically:  
1. A `PUT` request to `/api/security/role/developer_role?createOnly=true` is successfully processed.  
2. Immediately afterward, a `GET` request to `/api/security/role/developer_role?createOnly=true` is sent, which results in a 400 Bad Request error from the Kibana server.  

This behavior suggests that the provider's code is making an unnecessary or improperly formatted `GET` request after the successful `PUT`. 

I believe the problem is related to the implementation at the following code location:  

<https://github.com/elastic/terraform-provider-elasticstack/blob/v0.11.11/internal/kibana/role.go#L273>

Could you please investigate and provide a solution?  

**To Reproduce**

1. TF configuration used '...'

```bash
resource "elasticstack_kibana_security_role" "role" {
  name = "developer_role"

  elasticsearch {
    cluster = []
    indices {
      names      = ["application-dev-*"]
      privileges = ["read"]
    }
    run_as = []
  }

  kibana {
    feature {
      name       = "dashboard"
      privileges = ["minimal_read"]
    }

    feature {
      name       = "discover"
      privileges = ["minimal_read"]
    }

    spaces = ["default"]
  }
}
```

2. TF operations to execute to get the error '...'

```bash
tofu apply
```

3. See the error in the output '...'

```bash
2024-11-27T18:30:42.394+0700 [ERROR] provider.terraform-provider-elasticstack_v0.11.11: Response contains error diagnostic: @caller=github.com/hashicorp/terraform-plugin-go@v0.24.0/tfprotov6/internal/diag/diagnostics.go:58 @module=sdk.proto diagnostic_severity=ERROR diagnostic_summary="400 Bad Request" tf_resource_type=elasticstack_kibana_security_role diagnostic_detail="" tf_provider_addr=registry.terraform.io/elastic/elasticstack tf_rpc=ApplyResourceChange tf_proto_version=6.6 tf_req_id=da98b79c-16ca-15e9-6e5f-c3b209d16c3c timestamp="2024-11-27T18:30:42.394+0700"
[...]

Issue #926, opened 2024-11-22
[Bug] elasticstack_elasticsearch_data_stream_lifecycle data_retention serialization issue on apply


howtwizer:
**Describe the bug**
Truing to apply 
```
resource "elasticstack_elasticsearch_data_stream_lifecycle" "proxy_log_data_stream_lifecycle_multiple" {
  name           = "logs-proxy-*"
  data_retention = "30d"
}

```

Failing with error
```
│ Error: json: cannot unmarshal object into Go value of type []models.DataStreamLifecycle
│ 
│   with elasticstack_elasticsearch_data_stream_lifecycle.proxy_log_data_stream_lifecycle_multiple,
│   on data-stream-lifecycles.tf line 1, in resource "elasticstack_elasticsearch_data_stream_lifecycle" "proxy_log_data_stream_lifecycle_multiple":
│    1: resource "elasticstack_elasticsearch_data_stream_lifecycle" "proxy_log_data_stream_lifecycle_multiple" {
│ 
```

**To Reproduce**
1. TF configuration used '
``` terraform {
  required_version = ">= 0.12"
  required_providers {
    elasticstack = {
      source  = "elastic/elasticstack"
      version = "0.11.11"
    }
  }```'
2. TF operations to execute to get the error 'terraform apply' 
3. See the error in the output '
```
2024-11-22T13:02:01.692+0100 [ERROR] AttachSchemaTransformer: No provider config schema available for provider["terraform.io/builtin/terraform"]
2024-11-22T13:02:01.698+0100 [INFO]  provider: configuring client automatic mTLS
2024-11-22T13:02:01.727+0100 [INFO]  provider.terraform-provider-elasticstack_v0.11.11: configuring server automatic mTLS: timestamp="2024-11-22T13:02:01.726+0100"
elasticstack_elasticsearch_data_stream_lifecycle.proxy_log_data_stream_lifecycle_multiple: Creating...
2024-11-22T13:02:01.747+0100 [INFO]  Starting apply for elasticstack_elasticsearch_data_stream_lifecycle.proxy_log_data_stream_lifecycle_multiple
[...]

PR #964, opened 2024-12-18
chore: add support 8.16.2 to go-kibana-rest and acc tests

biscout42:
### TODOs

- [x] TestAccResourceDataStreamLifecycle - could be related to https://github.com/elastic/terraform-provider-elasticstack/issues/926 ? 
- [x] TestAccResourceIntegrationPolicySecretsFromSDK
- [x] TestAccDataSourceKibanaSecurityRole - probably related to https://github.com/elastic/terraform-provider-elasticstack/issues/933

### DIFF
--- a/internal/fleet/integration_policy/resource_test.go
+++ b/internal/fleet/integration_policy/resource_test.go
@@ -75,6 +75,9 @@ func TestAccResourceIntegrationPolicy(t *testing.T) {
 func TestAccResourceIntegrationPolicySecretsFromSDK(t *testing.T) {
 	policyName := sdkacctest.RandStringFromCharSet(22, sdkacctest.CharSetAlphaNum)
 
+	sdkConstrains, err := version.NewConstraint(">=8.10.0,<8.16.0")
+	require.NoError(t, err)
+
 	resource.Test(t, resource.TestCase{
 		PreCheck:     func() { acctest.PreCheck(t) },
 		CheckDestroy: checkResourceIntegrationPolicyDestroy,
@@ -86,7 +89,7 @@ func TestAccResourceIntegrationPolicySecretsFromSDK(t *testing.T) {
 						VersionConstraint: "0.11.7",
 					},
 				},
-				SkipFunc:           versionutils.CheckIfVersionIsUnsupported(minVersionIntegrationPolicy),
+				SkipFunc:           versionutils.CheckIfVersionMeetsConstraints(sdkConstrains),
 				Config:             testAccResourceIntegrationPolicySecretsCreate(policyName, "created"),
 				ExpectNonEmptyPlan: true, // secret churn
 				Check: resource.ComposeTestCheckFunc(
@@ -103,7 +106,7 @@ func TestAccResourceIntegrationPolicySecretsFromSDK(t *testing.T) {
 			},
 			{
 				ProtoV6ProviderFactories: acctest.Providers,
-				SkipFunc:                 versionutils.CheckIfVersionIsUnsupported(minVersionIntegrationPolicy),
+				SkipFunc:                 versionutils.CheckIfVersionMeetsConstraints(sdkConstrains),
 				Config:                   testAccResourceIntegrationPolicySecretsCreate(policyName, "created"),
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("elasticstack_fleet_integration_policy.test_policy", "name", policyName),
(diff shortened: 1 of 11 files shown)


internal/fleet/integration_policy/resource_test.go
```diff
@@ -75,6 +75,9 @@ func TestAccResourceIntegrationPolicy(t *testing.T) {
 func TestAccResourceIntegrationPolicySecretsFromSDK(t *testing.T) {
 	policyName := sdkacctest.RandStringFromCharSet(22, sdkacctest.CharSetAlphaNum)
 
+	sdkConstrains, err := version.NewConstraint(">=8.10.0,<8.16.0")
```
dimuon, line 78: Why do we disable it for >= 8.16.0?

biscout42: it doesn't work with the requested provider version and 8.16. Please, note, `0.11.7`. Asked here for more details: https://github.com/elastic/terraform-provider-elasticstack/commit/ba9abc91c9879de5c6a384d1437c07cf53813ea6#r150494267

daemitus: This was in the test results, prior to the selective disabling.
```
TestAccResourceIntegrationPolicySecretsFromSDK
    resource_test.go:78: Step 1/2 error: Check failed: Check 5/9 error: elasticstack_fleet_integration_policy.test_policy: Attribute 'vars_json' didn't match "{\"access_key_id\":{\"id\":\"\\S+\",\"isSecretRef\":true},\"default_region\":\"us-east-1\",\"endpoint\":\"endpoint\",\"secret_access_key\":{\"id\":\"\\S+\",\"isSecretRef\":true},\"session_token\":{\"id\":\"\\S+\",\"isSecretRef\":true}}", got "{}"
```

I think rather than constraining it to pre-8.16, just remove step 1 check 5 in the 0.11.7 check. The test passes after that, and the important part is how the data is represented in step 2.

biscout42: @daemitus , I assume step 2 is the same code as step 1 from `TestAccResourceIntegrationPolicySecrets`. My intention was to avoid code duplication and limit this case for previous versions as a migration sample (historical reference).

Would you still see the benefit of step 2 while we have  `TestAccResourceIntegrationPolicySecrets`? Please, advice, if I miss something.

daemitus: So it looks like 8.16 stopped wrapping the vars in a `{"value": "..."}` block when doing read requests.
This is actually required in the 0.11.7 provider logic, hence why we're getting an unexpected empty map (`{}`) as vars_json, rather than secretRefs. So this does indeed seem like a <8.16 legacy test. Additionally, anyone using integration policy resources on <=0.11.7 and 8.16.x will see a different kind of secret churn than they did before.

I now concur with the sdkConstrains check.

See below.

```
8.15.1: 
  "vars": {
    "shared_credential_file": { "type": "text" },
    "access_key_id": { "value": { "id": "xaJG25MBfiehiN8gM7uT", "isSecretRef": true }, "type": "password"},
    ...
  },

8.16.2
  "vars": {
    "access_key_id": { "id": "jm4l25MBNi2Twv6L31b2", "isSecretRef": true },
    ...
  },
```

biscout42: Superb, thanks. How did you managed to get debug output? somehow I wasn't able to get request / responses from integration policies by using `TF_LOG=DEBUG` 🤔 .

daemitus:
I tried a bunch of things, until I started sending it to disk with
`TF_ACC_LOG_PATH=/tmp/tf.log TF_ACC_LOG=DEBUG TF_LOG=DEBUG`

The TF_LOG might be extraneous, but it was working.

internal/fleet/integration_policy/resource_test.go
dimuon: @daemitus , I'm curious, kbapi is generated from Kibana `main` docs which seems to differ from the ones from 8.16. Can this lead to the test failure?

daemitus: See https://github.com/elastic/terraform-provider-elasticstack/blob/main/generated/kbapi/transform_schema.go#L959

While the api does list types for each, last time I checked the anyOf/oneOf union was broken upstream in the generated API. So instead kbapi just casts it to `any`. The current method extracts anything inside a `value` field and then continues into the regular process at https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/fleet/integration_policy/secrets.go#L83

dimuon: Thanks for the details.

dimuon (approved):
LGTM 👍

PR #974, opened 2025-01-07
fix: update team name - control-plane-hosted-applications

aribray:
The `control-plane-stateful-applications` team was renamed to `control-plane-hosted-applications`. This change needs to be reflected in this `catalog-info.yaml` as well.


nick-benoit (approved):
👍

Issue #975, opened 2025-01-09
[Bug] Since 0.11.12, elasticstack_kibana_data_view crashing with "panic: runtime error: invalid memory address or nil pointer dereference"


gavD:
**Describe the bug**
Our Terraform works fine with 0.11.11 and prior versions.

However, if we use 0.11.12, we get errors whenever using `elasticstack_kibana_data_view`

We suspect it's something to do with this change https://github.com/elastic/terraform-provider-elasticstack/pull/881/ which reworks dataview

**To Reproduce**
1. TF configuration used: hard to say what's relevant and what's not, but here's our TF:
```terraform
resource "elasticstack_kibana_import_saved_objects" "settings" {
  overwrite     = true
  file_contents = local.advanced_settings_content
}

resource "elasticstack_kibana_data_view" "index_pattern" {
  for_each = var.index_patterns

  override = true

  data_view = {
    allow_no_index  = true
    id              = each.key
    name            = each.value.title
    time_field_name = each.value.time_field
    title           = each.value.title
  }

  depends_on = [elasticstack_kibana_import_saved_objects.settings]
}
```
3. TF operations to execute to get the error: `terraform apply`
4. See the error in the output 
**Expected behavior**
It should not crash

**Debug output**

```
Planning failed. Terraform encountered an error while generating this plan.
243 |  
244 | ╷
245 | │ Error: Request cancelled
246 | │
247 | │ The plugin6.(*GRPCProvider).ReadResource request was cancelled.
248 | ╵
249 | ╷
250 | │ Error: Plugin did not respond
251 | │
252 | │ The plugin encountered an error, and failed to respond to the
253 | │ plugin6.(*GRPCProvider).ReadResource call. The plugin logs may contain more
254 | │ details.
255 | ╵
256 | ╷
257 | │ Error: Request cancelled
258 | │
259 | │ The plugin6.(*GRPCProvider).UpgradeResourceState request was cancelled.
260 | ╵
261 | ╷
262 | │ Error: Plugin did not respond
263 | │
264 | │ The plugin encountered an error, and failed to respond to the
265 | │ plugin6.(*GRPCProvider).ReadResource call. The plugin logs may contain more
266 | │ details.
267 | ╵
268 | ╷
269 | │ Error: Plugin did not respond
270 | │
271 | │ The plugin encountered an error, and failed to respond to the
272 | │ plugin6.(*GRPCProvider).ReadResource call. The plugin logs may contain more
273 | │ details.
[...]

PR #979, opened 2025-01-11
Restore support for non-composite data view ids

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/975

Basically restores the old implementation for [this function](https://github.com/elastic/terraform-provider-elasticstack/pull/881/files#diff-6ee1b4568dae9bbeea095558f5ebfd094ba702a77d3eda1e3a5f90f385def37dL278-L288). A previous provider version didn't include the composite ID and so we need to continue support. We could in theory write a state migration but that's a lot more work.


dimuon (approved):

dimuon:
just changelog is needed.

Issue #956, opened 2024-12-16
[Bug] security `es_client_authentication` should refer to the JWT token


Leaf-Lin:
**Describe the bug**
Documentation bug only. 

According to the [code](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/clients/config/elasticsearch.go#L253), this line of the elasticsearch security option [`es_client_authentication`](https://registry.terraform.io/providers/elastic/elasticstack/latest/docs#es_client_authentication-1) refers to JWT Token as refer set by the header [`ES-Client-Authentication `](https://www.elastic.co/guide/en/elasticsearch/reference/current/jwt-auth-realm.html#jwt-realm-configuration). But the docs has:
> [es_client_authentication](https://registry.terraform.io/providers/elastic/elasticstack/latest/docs#es_client_authentication-1) (String, Sensitive) ES Client Authentication field to be used with the bearer token

PR #982, opened 2025-01-11
Update es_client_authentication description

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/956

@Leaf-Lin is this the change you're after (you can just look at the first commit)? Otherwise, want to let me know what you'd like this to say?


Leaf-Lin (approved):
Thank you for the quick fix! I didn't realise it has been mentioned in so many different places. This will clear up the usage of the variable.

dimuon (approved):

tobio:
> Thank you for the quick fix! I didn't realise it has been mentioned in so many different places. This will clear up the usage of the variable. 

All good, the docs are generated so it's only one code change.

PR #1001, opened 2025-02-03
Additional dataview field formats

daemitus:
This mostly comes from a coworker who noticed I had touched the kbapi generator.

- Since the dataview field_format params are explicily defined, rather than another big jsonencode() block, we get to add all these fields manually.  
- Updated the transform script to add all the missing dataview field format params
- Updated the transform script to allow for arbitrary github refs rather than just head/main.
- Using said github_ref, pulled 2024-11-19 (close to when I originally PR'd the whole thing), which pulled in one new field that needed handling (in agent policies, required_versions) by adding an omit-empty.
- Added -f to the kbapi curl command, was getting silent 429 errors
- A bunch of tests/docs to support all the params.


tobio (approved):
LGTM. Thanks for thing folks!

PR #1037, opened 2025-02-26
[ResponseOps] Maintenance Window Resource

adcoelho:
Closes https://github.com/elastic/kibana/issues/197021

## Summary

The response-ops team has been working on a public maintenance window API.

Most of the work is done and the feature should be available for v9.1.0. ([Feature branch PR](https://github.com/elastic/kibana/pull/216756))

This PR enables users to define maintenance windows tf resources.

## Testing

The server version needs to be `>=9.1.0` so on all the tests I have been using the maintenance window API [feature branch](https://github.com/elastic/kibana/pull/216756).

I built it locally and ran `testacc` or Terraform commands.

**What would the proper procedure be to deploy this change?**

## Example resource
```
resource "elasticstack_kibana_maintenance_window" "my_maintenance_window" {
  title   = "Terraform Maintenance Window Updated"
  enabled = true
  custom_schedule {
    start    = "1992-01-01T05:00:00.200Z"
    duration = "12d"

    recurring {
      every       = "20d"
      end         = "2029-05-17T05:05:00.000Z"
      on_week_day = ["MO", "TU"]
    }
  }

  scope {
    alerting {
      kql = "_id: '1234'"
    }
  }
}
```


internal/kibana/maintenance_window.go
```diff
@@ -0,0 +1,455 @@
...
+			Summary:  "Maintenance window API not supported",
+			Detail:   fmt.Sprintf(`The maintenance Window public API feature requires a minimum Elasticsearch version of "%s"`, maintenanceWindowPublicAPIMinSupportedVersion),
+		})
+		return diags
+	}
+	return nil
+}
+
+func ResourceMaintenanceWindow() *schema.Resource {
+	apikeySchema := map[string]*schema.Schema{
```
jcger, line 34: seems like this wasn't generated but implemented by you. If so, couldn't this be generated based on the OAS files we create on the Kibana side?

adcoelho: The generation is only for the API client, so the resource specification is usually done manually. Maybe there is a way but I don't think we use it in this project.

tobio (commented):
Are we able to use the consolidated [kibana](https://github.com/elastic/terraform-provider-elasticstack/tree/main/generated/kbapi) spec for this new resource rather than on the older alerting client?

internal/kibana/maintenance_window.go
```diff
@@ -0,0 +1,455 @@
...
+			Severity: diag.Error,
+			Summary:  "Maintenance window API not supported",
+			Detail:   fmt.Sprintf(`The maintenance Window public API feature requires a minimum Elasticsearch version of "%s"`, maintenanceWindowPublicAPIMinSupportedVersion),
+		})
+		return diags
+	}
+	return nil
+}
+
+func ResourceMaintenanceWindow() *schema.Resource {
```
tobio, line 33: Can we rebuild this as a plugin-framework based resource ([data view](https://github.com/elastic/terraform-provider-elasticstack/tree/main/internal/kibana/data_view) as an example). It automatically solves a bunch of TF related limitations, and should be the future for any new work in the provider.

internal/kibana/maintenance_window.go
tobio: Hashicorp have an openapi [generator](https://developer.hashicorp.com/terraform/plugin/code-generation/openapi-generator) which builds a schema based on a set of openapi routes. It's not used in this project yet, but we're using it to help with [Serverless project resources](https://github.com/elastic/terraform-provider-ec/tree/master/ec/internal/gen/serverless). It's... ok, but has some limitations to deal with.

adcoelho:
I added 2 commits that address different issues.

1. [bf84803](https://github.com/elastic/terraform-provider-elasticstack/pull/1037/commits/bf84803175e85d5beddd02cf6d484c1024e0dbb3) removes the `occurrences` field. When this field was `undefined` it was being wrongly set to `0`. A future update to use the Terraform plugin framework should fix this.
2. [2a65569](https://github.com/elastic/terraform-provider-elasticstack/pull/1037/commits/2a655691a55d8f51079b9440d54bc6a7111f9992) adds a "serverless check". The public Maintenance Window API is currently available on serverless. The resource can now be used with `serverless || version >= 9.1.0`.

I will create two follow-up issues to update the alerting and the maintenance window resources to use the plugin framework.

@tobio could you please take another look at this PR?

adcoelho:
Closing because I am opening a new PR with a different implementation https://github.com/elastic/terraform-provider-elasticstack/pull/1224

Issue #667, opened 2024-06-17
[Feature] Support data tagging/add_field in Agent Policy (8.15)


mholttech:
release 8.15 has a new feature planned to add global fields at the Agent Policy. I want to make sure that the functionality is added to the terraform provider upon the release of 8.15.

Please see https://github.com/elastic/kibana/issues/179915

mholttech:
CC: @bvader

Zyther:
re-exploring implementing this against latest..

PR #1044, opened 2025-03-05
Issue667 - Add `global_data_tags` to fleet agent policies

Zyther:
Whew. 

Picking up from where @mholttech left off, this implements global data tags on the agent policy resource. 

This change: 
* implements [Issue 667](https://github.com/elastic/terraform-provider-elasticstack/issues/667)
* Modifies CRUD of agent policies to include a global_data_tags block, if it exists
* asks for a nested attribute object at the hcl/tf resource level
* updates kibana.gen.go to 3757e641278a5186919e35a0f980ac3cda7e8ccd
* links the global data tags schemas together (post, put, get) so they can be used to marshal/unmarshal universally

Thanks! Let me know what else is needed.


Zyther:
oi, need to fix this. apologies

Zyther:
Question.....

what commit of `elastic/kibana` was used to generate the latest `kibana.gen.go` ?

mholttech:
@tobio This is the follow up from the work I was originally doing on this. Please take a look when you can

Zyther (commented):
uhh.. did you guys remove the tcp integration from 9.0??

Zyther:
This last commit, generating the oapi spec off of github_ref=3757e641278a5186919e35a0f980ac3cda7e8ccd fixed all issues except that 9.0 tcp issue

internal/fleet/agent_policy/models.go
```diff
@@ -54,10 +59,23 @@ func (model *agentPolicyModel) populateFromAPI(data *kbapi.AgentPolicy) {
 	model.MonitoringOutputId = types.StringPointerValue(data.MonitoringOutputId)
 	model.Name = types.StringValue(data.Name)
 	model.Namespace = types.StringValue(data.Namespace)
+	if serverVersion.GreaterThanOrEqual(MinVersionGlobalDataTags) && utils.Deref(data.GlobalDataTags) != nil {
```
tobio:
```suggestion
	if utils.Deref(data.GlobalDataTags) != nil {
```

Do we need the server version check here? If the response from Kibana includes global_data_tags it seems reasonable to assume they're supported.

internal/fleet/agent_policy/schema.go
```diff
@@ -12,74 +12,86 @@ import (
...
-		"skip_destroy": schema.BoolAttribute{
-			Description: "Set to true if you do not wish the agent policy to be deleted at destroy time, and instead just remove the agent policy from the Terraform state.",
-			Optional:    true,
-		},
-		"sys_monitoring": schema.BoolAttribute{
-			Description: "Enable collection of system logs and metrics.",
-			Optional:    true,
-			PlanModifiers: []planmodifier.Bool{
-				boolplanmodifier.RequiresReplace(),
+			"global_data_tags": schema.StringAttribute{
```
tobio: I'm not a big fan of this being an arbitrary json string. It's hard to document and build proper validation around. 

At the very least we should set `CustomType:  jsontypes.NormalizedType{}` and add a validator ensuring the json string can be serialised into `[]kbapi.AgentPolicyGlobalDataTagsItem`.

I do think it would be nicer to use a `MapNestedAttribute` here with an element like `object({ string_value = optional(string), number_value = optional(number) })`. IMO it would provide a much nicer provider first user experience rather than having to dive through both provider and Fleet documentation. Language server users would also get proper autocompletion/inline linting as well.

tobio (commented):
Thanks for taking the time to add this in! Are you able to take a look at using a strongly typed schema for these tags rather than an arbitrary json blob?

Zyther:
ok! mapnested implemented. Still fails on 9.0, but passes everywhere else

internal/fleet/agent_policy/models.go
```diff
@@ -22,11 +33,12 @@ type agentPolicyModel struct {
 	MonitorMetrics     types.Bool   `tfsdk:"monitor_metrics"`
 	SysMonitoring      types.Bool   `tfsdk:"sys_monitoring"`
 	SkipDestroy        types.Bool   `tfsdk:"skip_destroy"`
+	GlobalDataTags     types.Map    `tfsdk:"global_data_tags"` //> globalDataTagsModel
 }
 
-func (model *agentPolicyModel) populateFromAPI(data *kbapi.AgentPolicy) {
+func (model *agentPolicyModel) populateFromAPI(ctx context.Context, data *kbapi.AgentPolicy, serverVersion *version.Version) diag.Diagnostics {
```
tobio: `serverVersion` is unused now.

internal/fleet/agent_policy/models.go
```diff
@@ -77,10 +118,53 @@ func (model agentPolicyModel) toAPICreateModel() kbapi.PostFleetAgentPoliciesJSO
 		Namespace:          model.Namespace.ValueString(),
 	}
 
-	return body
+	if len(model.GlobalDataTags.Elements()) > 0 {
+		var diags diag.Diagnostics
+		if serverVersion.LessThan(MinVersionGlobalDataTags) {
+			diags.AddError("global_data_tags ES version error", "Global data tags are only supported in Elastic Stack 8.15.0 and above")
```
tobio:
```suggestion
			diags.AddError("global_data_tags ES version error", fmt.Sprintf("Global data tags are only supported in Elastic Stack %s and above", MinVersionGlobalDataTags))
```

internal/fleet/agent_policy/models.go
```diff
@@ -100,5 +184,47 @@ func (model agentPolicyModel) toAPIUpdateModel() kbapi.PutFleetAgentPoliciesAgen
...
```
tobio: Same here, can we use the version variable rather than a hardcoded string.

internal/fleet/agent_policy/models.go
```diff
@@ -77,10 +118,53 @@ func (model agentPolicyModel) toAPICreateModel() kbapi.PostFleetAgentPoliciesJSO
...
+			return kbapi.PostFleetAgentPoliciesJSONRequestBody{}, diags
+		}
+
+		items := utils.MapTypeToMap(ctx, model.GlobalDataTags, path.Root("global_data_tags"), &diags,
+			func(item globalDataTagsItemModel, meta utils.MapMeta) kbapi.AgentPolicyGlobalDataTagsItem {
+				// do some checks
+				if item.StringValue.ValueStringPointer() == nil && item.NumberValue.ValueFloat32Pointer() == nil || item.StringValue.ValueStringPointer() != nil && item.NumberValue.ValueFloat32Pointer() != nil {
+					diags.AddError("global_data_tags validation_error", "Global data tags must have exactly one of string_value or number_value")
+					return kbapi.AgentPolicyGlobalDataTagsItem{}
+				}
```
tobio:
```suggestion
```

We can define this validation within the schema

internal/fleet/agent_policy/models.go
```diff
@@ -100,5 +184,47 @@ func (model agentPolicyModel) toAPIUpdateModel() kbapi.PutFleetAgentPoliciesAgen
...
+			return kbapi.PutFleetAgentPoliciesAgentpolicyidJSONRequestBody{}, diags
+		}
+
+		items := utils.MapTypeToMap(ctx, model.GlobalDataTags, path.Root("global_data_tags"), &diags,
+			func(item globalDataTagsItemModel, meta utils.MapMeta) kbapi.AgentPolicyGlobalDataTagsItem {
+				// do some checks
+				if item.StringValue.ValueStringPointer() == nil && item.NumberValue.ValueFloat32Pointer() == nil || item.StringValue.ValueStringPointer() != nil && item.NumberValue.ValueFloat32Pointer() != nil {
+					diags.AddError("global_data_tags validation_error", "Global data tags must have exactly one of string_value or number_value")
+					return kbapi.AgentPolicyGlobalDataTagsItem{}
+				}
```
tobio:
```suggestion
```

Similarly

internal/fleet/agent_policy/schema.go
```diff
@@ -3,83 +3,112 @@ package agent_policy
...
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"string_value": schema.StringAttribute{
+							Description: "String value for the field. If this is set, number_value must not be defined.",
+							Optional:    true,
+						},
+						"number_value": schema.Float32Attribute{
+							Description: "Number value for the field. If this is set, string_value must not be defined.",
+							Optional:    true,
+						},
```
tobio:
```suggestion
						"string_value": schema.StringAttribute{
							Description: "String value for the field. If this is set, number_value must not be defined.",
							Optional:    true,
							Validators: []validator.String{
								stringvalidator.ConflictsWith(path.MatchRelative().AtName("number_value")),
							},
						},
						"number_value": schema.Float32Attribute{
							Description: "Number value for the field. If this is set, string_value must not be defined.",
							Optional:    true,
							Validators: []validator.Float32{
								float32validator.ConflictsWith(path.MatchRelative().AtName("number_value")),
							},
						},
```

Are you able to add an acceptance test verifying that this validation is enforced as well?

tobio (commented):
Looking great! A couple of small changes before we can get this merged.

internal/fleet/agent_policy/schema.go
Zyther: All set, with a new test to test a bad global_data_tag setup

mholttech:
@tobio can you review the updates provided? Would love to get this merged and a new release generated by the end of the week if possible.

internal/fleet/agent_policy/schema.go
```diff
@@ -3,83 +3,122 @@ package agent_policy
...
+							Optional:    true,
+							Validators: []validator.Float32{
+								float32validator.ConflictsWith(path.MatchRelative().AtParent().AtName("string_value")),
+							},
+						},
+					},
+				},
+				Optional: true,
+				PlanModifiers: []planmodifier.Map{
+					mapplanmodifier.RequiresReplace(),
```
tobio: We're specifying global data tags in the update request, why are we setting this as `RequiresReplace`?

Zyther: Let me test this without it -- when it was a jsonencode block, there were state issues if this modifier was not present.

Zyther:
this still persists without it
    resource_test.go:72: Step 6/6 error: Error running apply: exit status 1
        
Zyther:
based on this, i will readd

tobio:
> this still persists without it resource_test.go:72: Step 6/6 error: Error running apply: exit status 1
> 
> ```
>     Error: Provider produced inconsistent result after apply
> ```

This sounds like a bug that needs to be fixed. `RequiresReplace` will create usability issues and should only be in place if it's truly required.

Zyther:
if i understand this correctly, I think global_data_tags needs to exist empty on the payload going out on update -- if there are no tags. If this doesnt work, i may need to modify the apiUpdate to have an empty array if there are none

Zyther:
how do we feel about this?

tobio:
I'm taking a look here, I think this behaviour (unsetting global data tags when they're removed from the TF definition) is different to our other resources and it would be good to keep things consistent.

mholttech:
Can you clarify what you mean here Toby? I would expect that if the global data tags are being managed in TF then every data tag in the target policy is added and removed based on the TF. If you are doing a resource import of an existing policy then then during a plan or apply operation TF will let you know if the policy has tags that are not defined in TF through the normal resource diff.

tobio:
Yep, that's a reasonable expectation. IIRC that's not the behaviour for the other resources in the Elastic stack provider which follow the Elasticsearch API behaviour, i.e a field retains its current value unless explicitly set to null.

mholttech:
Hi @tobio,

I haven't had the chance to review the way that other resources in the provider function, however after talking through this with the team I feel strongly that the way this is implemented is correct and is the normal behavior expected when using Terraform.

From a Terraform design perspective, the current implementation follows the standard declarative approach where your configuration represents the complete desired state. The typical behavior across mature Terraform providers is that:

1. When you define a resource in Terraform, all manageable properties should be represented
2. If a property is removed from your Terraform configuration, it should be removed from the actual resource
3. The resource state should exactly match what's defined in your configuration

This approach prevents configuration drift and maintains Terraform as the single source of truth. If we were to preserve tags not defined in Terraform, or require forcing them to null to remove, it would create situations where resource properties exist outside of Terraform's management scope, making it difficult to track what is actually controlled by code versus manual changes.

While I understand your concern about consistency with other Elastic Stack resources that may behave differently, I believe following Terraform's conventional approach to resource management provides a more predictable and maintainable user experience in the long run.

mholttech:
I just finished reviewing the tests again and we did find one missing assertion that @Zyther just added. Overall the tests are showing the exact behavior that I expect when utilizing a terraform provider.

internal/fleet/agent_policy/models.go
```diff
@@ -100,5 +179,47 @@ func (model agentPolicyModel) toAPIUpdateModel() kbapi.PutFleetAgentPoliciesAgen
 		Namespace:          model.Namespace.ValueString(),
 	}
 
-	return body
+	if len(model.GlobalDataTags.Elements()) > 0 {
```
tobio: It looks like we copy this between create/update. Can we pull it out and DRY this code up a little?

internal/fleet/agent_policy/resource_test.go
```diff
@@ -118,6 +122,48 @@ func TestAccResourceAgentPolicy(t *testing.T) {
...
+				Config:   testAccResourceAgentPolicyUpdateWithNoGlobalDataTags(policyNameGlobalDataTags, false),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("elasticstack_fleet_agent_policy.test_policy", "name", fmt.Sprintf("Updated Policy %s", policyNameGlobalDataTags)),
+					resource.TestCheckResourceAttr("elasticstack_fleet_agent_policy.test_policy", "namespace", "default"),
+					resource.TestCheckResourceAttr("elasticstack_fleet_agent_policy.test_policy", "description", "This policy was updated without global data tags"),
+					resource.TestCheckResourceAttr("elasticstack_fleet_agent_policy.test_policy", "monitor_logs", "false"),
+					resource.TestCheckResourceAttr("elasticstack_fleet_agent_policy.test_policy", "monitor_metrics", "true"),
+					resource.TestCheckResourceAttr("elasticstack_fleet_agent_policy.test_policy", "skip_destroy", "false"),
+					resource.TestCheckNoResourceAttr("elasticstack_fleet_agent_policy.test_policy", "global_data_tags.tag1"),
+					resource.TestCheckNoResourceAttr("elasticstack_fleet_agent_policy.test_policy", "global_data_tags.tag2"),
```
tobio:
```suggestion
					resource.TestCheckResourceAttr("elasticstack_fleet_agent_policy.test_policy", "global_data_tags.#", "0"),
```

It's perhaps more clear/accurate to assert that there are no global data tags defined.

tobio (commented):
Couple of small things we can improve.

tobio:
> While I understand your concern about consistency with other Elastic Stack resources that may behave differently, I believe following Terraform's conventional approach to resource management provides a more predictable and maintainable user experience in the long run.

Ultimately everything you've said is correct, and I certainly don't disagree. Critically it doesn't look like this is the first case we exhibit the 'correct' behaviour within the provider, which makes it a pretty simple decision in this case :) Let's keep this as is, there's a couple of small issues to fix up but otherwise thanks for all the work making these changes!

tobio:
> @tobio Looks like all the work here is done and all the tests passed, when can we get this merged and released?

[This](https://github.com/elastic/terraform-provider-elasticstack/pull/1044#discussion_r2028000092) comment is still unresolved.

tobio (approved):
LGTM, thanks for persisting with this!

Issue #1050, opened 2025-03-11
[Bug] 400 while creating Kibana Security Role


ggilligan12:
### Bug report

I am experiencing HTTP 400 Bad Request errors on Kibana Security Role creation at apply time.

Provider version: `0.11.12`
Terraform version: `1.6.6`
Elasticsearch/Kibana version: `8.16.3`

My plan:

```
$ /usr/bin/terraform plan -target module.roles.elasticstack_kibana_security_role.index_readonly_roles[\"tf-provider-bug-reproduction\"]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # module.roles.elasticstack_kibana_security_role.index_readonly_roles["tf-provider-bug-reproduction"] will be created
  + resource "elasticstack_kibana_security_role" "index_readonly_roles" {
      + id       = (known after apply)
      + metadata = (known after apply)
      + name     = "tf-provider-bug-reproduction"

      + elasticsearch {
          + cluster = []
          + run_as  = []

          + indices {
              + names      = [
                  + "nonexistent-index-pattern-*",
                ]
              + privileges = [
                  + "read",
                ]
                # (1 unchanged attribute hidden)
            }
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.
```

At apply time I see the following in the debug output:
```
2025-03-11T09:26:20.434Z [ERROR] provider.terraform-provider-elasticstack_v0.11.12: Response contains error diagnostic: tf_provider_addr=registry.terraform.io/elastic/elasticstack tf_rpc=ApplyResourceChange @module=sdk.proto diagnostic_severity=ERROR diagnostic_summary="400 Bad Request" tf_proto_version=6.7 tf_resource_type=elasticstack_kibana_security_role @caller=github.com/hashicorp/terraform-plugin-go@v0.25.0/tfprotov6/internal/diag/diagnostics.go:58 diagnostic_detail="" tf_req_id=03fe5746-9de4-7e6f-0a79-611f35c69afc timestamp=2025-03-11T09:26:20.434Z
2025-03-11T09:26:20.434Z [WARN]  unexpected data:
  registry.terraform.io/elastic/elasticstack:stderr=
  | 2025/03/11 09:26:20.434057 DEBUG RESTY
  | ==============================================================================
  | ~~~ REQUEST ~~~
[...]

PR #1051, opened 2025-03-11
Bugfix for query param reuse (#1050)

ggilligan12:
Proposed fix for issue raised in #1050. Would appreciate a maintainer taking a look.

Yes we could just unset the query param but that feels like the wrong approach, we shouldn't be statically modifying this context object if its going to be used for both requests I feel... happy to hear dissent though!

Please do comment on code quality/style also. Happy to conform to whatever standards you guys prefer.

Nb. I had difficulties getting your `Makefile` to run on my machine as I work in an environment with a restrictive network policy. I'd really appreciate someone taking the time to run this locally to validate it works. Really more of a candidate fix as a starting point for someone else, I've no idea if this is going to do the trick.


ggilligan12:
Had this completely backwards. Latest version of the provider fixes this issue. The change I identified as the problem actually introduced the fix. Incorrect understanding of how a pass by reference works in Go on my part.

PR #1057, opened 2025-03-13
Add Request Headers to bypass security gates 

Dakumo:
We have everything which is not necassarily needed to be public behind a security gate which can be bypassed by adding a header to the request.

So I extended the provider to handle headers...

maybe the config should look different... but my go coding skills are very basic

Idea from me: 
```
 headers = {
    "X-Custom-Header" = "CustomValue"
  }
```

Currently you can add headers like: 
```
headers = ["X-Custom-Header:CustomValue"]
```


internal/schema/connection.go
```diff
@@ -311,6 +317,15 @@ func GetEsConnectionSchema(keyName string, isProviderConfiguration bool) *schema
 						Type: schema.TypeString,
 					},
 				},
+				"headers": {
+					Description: "A list of headers to be sent with each request to Elasticsearch.",
+					Type:        schema.TypeList,
+					Optional:    true,
+					Sensitive:   true,
+					Elem: &schema.Schema{
+						Type: schema.TypeString,
```
tobio, line 326: We should make these a map for `{"Header" = "value"}`

internal/clients/config/elasticsearch.go
```diff
@@ -145,6 +145,24 @@ func newElasticsearchConfigFromFramework(ctx context.Context, cfg ProviderConfig
 		config.config.Addresses = endpoints
 	}
 
+	var headers []string
```
tobio: You'll need to add the equivalent functionality into `newElasticsearchConfigFromSDK` to cover legacy resources.

tobio (commented):
Thanks for adding this in. Can we update the schema attribute to be a map rather than require the `key:value` formatting.

internal/clients/config/elasticsearch.go
Dakumo: added it there. :) thanks for that advice

internal/schema/connection.go
Dakumo: can you please help me with that? 

i tried it but I failed... you are also welcome to implement that... since I'm just a beginner in go / terraform provider coding.

tobio: I've made this change (and regenerated the docs) in [this](https://github.com/elastic/terraform-provider-elasticstack/compare/headers-map?expand=1) branch. You should be able to pull that down and cherry pick the commits (I can't push to the main branch in your fork).

Dakumo:
@tobio sorry for the long delay... merged your stuff in my branch

thanks for your support here! :)

internal/schema/connection.go
Dakumo: done! :)

tobio (commented):
LGTM. Are you able to add a changelog entry for this PR too? 

I strongly suspect the CLA checker if failing due to these two unverified commits. You should be able to either rewrite the author of those commits to use an email associated with your Github account, or add the email used in those commits to your Github account. 

image

Dakumo:
hi there, 

sorry for the delay... i just fix the issue about the verified commits... 

and added information to the changelog.md

tobio (approved):
LGTM, thanks for taking the time to get this added in!

Dakumo:
can you please create a new release with that change?

PR #1065, opened 2025-03-19
Update custom metrics `field` to be optional

wandergeek:
This PR updates the SLO spec to make `field` optional when dealing with a custom metrics SLO. I have had to make a very unfortunate [hack to the client ](https://github.com/elastic/terraform-provider-elasticstack/commit/2bf27a7a613a21a3b5b4f12395c5e736d2b25542) from what appears to be a bug in the generator. It would be good to merge this in now and we can try and fix the upstream issue soon. We are likely due for some more changes to this resource so I think we will need to fix this sooner rather than later.


tobio:
The Kibana spec still lists `field` as [required](https://github.com/elastic/kibana/blob/main/x-pack/solutions/observability/plugins/slo/docs/openapi/slo/bundled.yaml#L824)?

tobio:
So do the published [docs](https://www.elastic.co/docs/api/doc/kibana/v8/operation/operation-createsloop#operation-createsloop-body-application-json-elastic-api-version-2023-10-31-slos_indicator_properties_custom_metric-indicator-params-good-metrics-field)

wandergeek:
@tobio soon it will not :)
https://github.com/elastic/kibana/pull/215082
https://elastic.slack.com/archives/C044PV8EJ4X/p1742173010868859

wandergeek:
Closing in favor of a larger PR that updates to the latest client

Issue #1070, opened 2025-03-21
[Feature] Support preventinitialbackfill in kibana_slo


neiljbrookes:
**Is your feature request related to a problem? Please describe.**

N/A

**Describe the resource you would like to have implemented.**

Support for the `preventinitialbackfill` [setting](https://www.elastic.co/docs/api/doc/serverless/operation/operation-createsloop#operation-createsloop-body-application-json-settings-preventinitialbackfill) in the `kibana_slo` resource

**Describe the solution you'd like**

To be able to set `settings.preventinitialbackfill : true` when creating an SLO.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #1071, opened 2025-03-21
Add optional preventInitialBackfill for SLO API

jasonrhodes:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1070

Docs for this option are here: https://www.elastic.co/docs/api/doc/serverless/operation/operation-createsloop#operation-createsloop-body-application-json-settings-preventinitialbackfill

I've not done any Terraform additions like this one, so there is a good chance the syntax here is wrong — I see that this new setting likely needs to be added in a few additional places to make use of it after it's been added to the schema, so I'd appreciate guidance on where that needs to happen, exactly. 

At any rate, it would be great to add this field, and as it's a new, optional field, it shouldn't have any breaking change consequences whatsoever.

UPDATE: I re-generated the SLO model files with openapi-generator, hoping it would only produce a small scope of changes in those files. It seems to have produced a LOT of changes, so I'll wait for guidance on how to handle that.

### DIFF
--- a/internal/kibana/slo.go
+++ b/internal/kibana/slo.go
@@ -890,7 +890,7 @@ func resourceSloCreate(ctx context.Context, d *schema.ResourceData, meta interfa
 			return diag.Errorf("The 'prevent_initial_backfill' setting requires Elastic Stack version %s or higher.", SLOSupportsPreventInitialBackfillMinVersion)
 		}
 	}
-  
+
 	// Version check for data_view_id support
 	if !serverVersion.GreaterThanOrEqual(SLOSupportsDataViewIDMinVersion) {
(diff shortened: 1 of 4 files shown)


neiljbrookes:
Will address https://github.com/elastic/terraform-provider-elasticstack/issues/1070

internal/kibana/slo.go
```diff
@@ -457,6 +457,11 @@ func getSchema() map[string]*schema.Schema {
 						Optional: true,
 						Computed: true,
 					},
+					"preventInitialBackfill": {
```
kdelemme: other fields follow snake_case

wandergeek (approved):

PR #1114, opened 2025-04-25
Improve error messaging if no Kibana endpoint is configured.

gigerdo:
Fixes two issues that lead to confusing errors being shown:

- The `agent_policy` resource did not throw an error if getting the stack version fails
- When getting the stack version fails, it only returned the http connection error, not actually stating what it was trying to do.

In combination this will now lead to an error like
> `Error: failed to get version from Kibana API: Get "http://localhost:5601/api/status": dial tcp [::1]:5601: connect: connection refused,  Please ensure a working 'kibana' endpoint is configured.`


claudia-correia (approved):
LGTM 👍🏻

tobio:
@gigerdo are you able to add a change log entry for this too?

PR #1154, opened 2025-06-04
Migrate the System User resource to the plugin framework. 

tobio:
Mostly done by Windsurf so I guess there could be dragons... I think I've found the insane bits though.


internal/elasticsearch/security/system_user/acc_test.go
```diff
@@ -44,6 +44,37 @@ func TestAccResourceSecuritySystemUserNotFound(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				// Create the system user with the last provider version where the system user resource was built on the SDK
+				ExternalProviders: map[string]resource.ExternalProvider{
+					"elasticstack": {
+						Source:            "elastic/elasticstack",
+						VersionConstraint: "0.11.15",
+					},
+				},
+				ProtoV6ProviderFactories: acctest.Providers,
```
dimuon: Do we need it if we rely on `ExternalProviders`?

internal/elasticsearch/security/system_user/update.go
```diff
@@ -0,0 +1,85 @@
...
+	if diags.HasError() {
+		return diags
+	}
+	if user == nil || !user.IsSystemUser() {
+		diags.AddError("", fmt.Sprintf(`System user "%s" not found`, usernameId))
+		return diags
+	}
+
+	var userPassword models.UserPassword
+	// TB TODO Fix up this logic. It should only set the password when it's set in config
```
dimuon: Shall we fix this in this PR or in a follow-up one?

dimuon (dismissed):
Looks good besides 2 nit-picks and the missing changelog 👍

Issue #1152, opened 2025-06-02
[Feature] Add ability to configure Global Parameters


jkomara:
**Is your feature request related to a problem? Please describe.**
I am trying to use the provider to configure Synthetics Monitors that use secrets. The existing pattern to do this is to reference the secrets as environment variables. Those environment variables are set as Global Parameters. I would like to be able to configure the Global Parameters in Terraform so that I can use them with the Synthetics Monitors.

**Describe the resource you would like to have implemented.**
https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-parameters

**Describe the solution you'd like**
I would like to be able to set Global Parameters via Terraform.

**Describe alternatives you've considered**
Currently people are using node wrapper scripts to hit the Kibana API so that they can configure the parameters. This is possible, but would require us to set up a completely different environment just to manage these parameters.

**Additional context**
None

PR #1155, opened 2025-06-04
Add support for synthetics global parameters

agunnerson-elastic:
Add support for synthetics global parameters

This adds support for creating synthetics global parameters, which are
used to store secrets used for synthetics monitors.

This implementation is based on what we already do for synthetics
private locations. This is mostly just a simple wrapper around API calls, except we currently requery the resource after creating it or updating it. This is because Kibana (at least in 9.0.0) has some unexpected responses for `/api/synthetics/params`:

1. The POST request to create a parameter does not include the `value` field in the response.
2. The PUT request to update a parameter returns the new values for all fields, except for `value`, which contains the old secret.
3. None of the responses include the `share_across_spaces` value, which can only be inferred by checking of `namespaces` is `["*"]`.

We also use the undocumented `DELETE /api/synthetics/params` API that the Kibana UI uses due to many quirks with the deletion APIs in various versions of Kibana. I tested every single version of Kibana since the parameter APIs were introduced (8.12.0) and this one is the only one that works on all versions.

* `DELETE /api/synthetics/params/<id>` (from documentation)
    - Works on >=8.17.0.
    - HTTP 404 on 8.12.x through 8.16.x.
* `DELETE /api/synthetics/params/_bulk_delete` (from documentation)
    - HTTP 400 on 9.0.x with error message about URL parameters and body provided at the same time (`_bulk_delete` interpreted as parameter ID?).
    - HTTP 403 on 8.18.x with error message about missing `uptime-read` and `uptime-write` permissions despite having the `superuser` role.
    - HTTP 403 with no details on 8.17.x.
    - HTTP 404 on 8.12.x through 8.16.x.
* `POST /api/synthetics/params/_bulk_delete` (from comment in documentation example)
    - Works on >=8.17.0.
    - HTTP 404 on 8.12.x through 8.16.x.
* `DELETE /api/synthetics/params` (what the Kibana UI does)
    - Works on >=8.12.0.

Finally, this updates the Github Actions workflow to run the tests using HTTP basic auth instead of API keys. On all versions of Kibana, the params endpoint fails to return the current value when using API keys, even if permissions are not restricted for the key. Explicitly granting the key all permissions to Kibana does not work either.

Fixes: #1152


tobio:
Have we opened issues in the Kibana repo to cover these problems?

internal/kibana/synthetics/parameter/create.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		response.Diagnostics.AddError(fmt.Sprintf("Failed to create parameter `%s`", input.Key), err.Error())
+		return
+	}
+
+	resourceId := result.Id
+
+	// We can't trust the response from the POST request. At least with Kibana
+	// 9.0.0, it responds without the `value` field set.
+	result, err = kibanaClient.KibanaSynthetics.Parameter.Get(ctx, resourceId)
```
tobio: Can we just re-use the read code here?

internal/kibana/synthetics/parameter/read.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	state = toModelV0(*result)
+
+	// Set refreshed state
+	diags = response.State.Set(ctx, &state)
+	response.Diagnostics.Append(diags...)
+	if response.Diagnostics.HasError() {
+		return
+	}
```
tobio: Can we pull this code out so it can be re-used in create/update rather than duplicate the read logic?

libs/go-kibana-rest/kbapi/api._.go
```diff
@@ -128,6 +129,12 @@ func New(c *resty.Client) *API {
 				Delete: newKibanaSyntheticsPrivateLocationDeleteFunc(c),
 				Get:    newKibanaSyntheticsPrivateLocationGetFunc(c),
 			},
+			Parameter: &KibanaSyntheticsParameterAPI{
+				Add:    newKibanaSyntheticsParameterAddFunc(c),
+				Delete: newKibanaSyntheticsParameterDeleteFunc(c),
+				Get:    newKibanaSyntheticsParameterGetFunc(c),
+				Update: newKibanaSyntheticsParameterUpdateFunc(c),
+			},
```
tobio: We're trying to move everything to a client generated from the Kibana openapi spec. It's not as smooth as I'd like at the moment, but I've done a bit of the work to get started in [this](https://github.com/elastic/terraform-provider-elasticstack/commit/34cfe56c2716e18373a6c38eee42511cb3d22970) commit. Would you be able to migrate this PR over to that client instead of updating this one? 

Given the delete endpoint you're using here is undocumented we may need to leave that code in here, but at least for get/post/put we should be able to use the generated client.

tobio (commented):
This is great! Thanks for adding this in, a couple of small comments but it would be great if we can base this off the open api client if possible.

libs/go-kibana-rest/kbapi/api._.go
agunnerson-elastic: Thanks for starting the work on that! I'll give it a shot.

internal/kibana/synthetics/parameter/read.go
agunnerson-elastic: Good idea. Done in 13df2f0352e2d0888dd31c64df19831c12c368a3.

libs/go-kibana-rest/kbapi/api._.go
agunnerson-elastic: I just pushed c7d42a3c7a45e526f99118dd5ae1acd13dc90ccf. The synthetics parameters stuff is working, but it looks like the fleet tests are failing. The only place I'm seeing any changes related to fleet is in `generated/kbapi/kibana.gen.go`, but I'm not sure why. It seems like the OAPI changes are additive. Any idea what might be going on?

tobio: Sorry for taking ages here. The failures in this PR are all fixed by https://github.com/elastic/terraform-provider-elasticstack/pull/1189. We should just be able to merge main when it's fixed.

agunnerson-elastic: Thank you!

PR #1182, opened 2025-07-06
Remove explicit misspell

tobio:
This is included in golangci-lint so we're just duplicating work here


dimuon (commented):
I think `misspell` linter is disabled by default in `golangci-lint` and isn't specified in `.golangci.yaml` so it's not invoked.

tobio:
> I think `misspell` linter is disabled by default in `golangci-lint` and isn't specified in `.golangci.yaml` so it's not invoked.

You're right, I misread the linter output. Explicitly enabled it in config.

dimuon (approved):

Issue #814, opened 2024-10-01
[Feature] Timeslice metric indicator for  `elasticstack_kibana_slo` (supporting percentile aggregations)


irregulator:
Hi! Thanks for developing this Terraform provider!

**Describe the resource you would like to have implemented.**

We would like to manage `elasticstack_kibana_slo` objects with Terraform and employ Timeslice metric indicator. We are particularly interested in **Percentile** aggregation. This feature is already present in Kibana as described in:
- https://www.elastic.co/guide/en/observability/current/slo-create.html#timeslice-metric-sli
- https://www.elastic.co/docs/api/doc/kibana/v8/operation/operation-createsloop#operation-createsloop-body-application-json-elastic-api-version-2023-10-31-slos_indicator_properties_timeslice_metric-indicator

**Describe the solution you'd like**

A new indicator could look like this:

```hcl
timeslice_custom_indicator {
  index           = "heartbeat-*"
  filter          = "monitor.name: \"loadbalancer\""
  timestamp_field = "@timestamp"
  metrics = [
    {
      name = "p95"
      field = "http.rtt.total.us"
      aggregation = "percentile"
      percentile = 95
    } 
  ]  
  equation = "p95"
  comparator = "LT"
  threshold = 2000000
}
```

**Describe alternatives you've considered**

None

**Additional context**

Here's the equivalent API request when creating an example latency SLO with Timeslice metrics and Percentile aggregation through Kibana UI:

```javascript
{
  "name": "Latency SLO",
  "description": "Latency SLO",
  "indicator": {
    "type": "sli.metric.timeslice",
    "params": {
      "index": "heartbeat-*",
      "filter": "monitor.name: \"loadbalancer\"",
      "metric": {
        "metrics": [
          {
            "name": "A",
            "aggregation": "percentile",
            "field": "http.rtt.total.us",
            "filter": "",
            "percentile": 95
          }
        ],
        "equation": "A",
        "comparator": "LT",
        "threshold": 2000000
      },
      "timestampField": "@timestamp"
    }
  },
  "budgetingMethod": "timeslices",
  "timeWindow": {
    "duration": "30d",
    "type": "rolling"
  },
  "objective": {
    "target": 0.99,
    "timesliceTarget": 0.95,
    "timesliceWindow": "1m"
  },
  "tags": [],
  "groupBy": "url.domain"
}
```

maquema:
it will be great to be implemented! we really need it !

[...]

PR #1195, opened 2025-07-09
Add support for timeslice metrics indicator

lloydmeta:
* Add support for the resource based on OpenAPI [docs](https://www.elastic.co/docs/api/doc/kibana.yaml)
* Updates SLO OpenAPI spec in this project to add discriminator.
* Add tests
* Generate docs

Closes https://github.com/elastic/terraform-provider-elasticstack/issues/814


internal/kibana/slo.go
```diff
@@ -873,6 +1007,42 @@ func resourceSloRead(ctx context.Context, d *schema.ResourceData, meta interface
...
+				metric["name"] = m.TimesliceMetricPercentileMetric.Name
+				metric["aggregation"] = m.TimesliceMetricPercentileMetric.Aggregation
+				metric["field"] = m.TimesliceMetricPercentileMetric.Field
+				metric["percentile"] = m.TimesliceMetricPercentileMetric.Percentile
+			}
+			if m.TimesliceMetricDocCountMetric != nil {
+				metric["name"] = m.TimesliceMetricDocCountMetric.Name
+				metric["aggregation"] = m.TimesliceMetricDocCountMetric.Aggregation
+			}
+			metrics = append(metrics, metric)
```
lloydmeta, line 1031: There's an issue here that I'm not sure can be easily resolved, which leads to `percentile` always being 0 when demarshalled....:

https://github.com/elastic/terraform-provider-elasticstack/blob/77c15d2b032ffa41bcb4d791b263cf6ee422b0cc/generated/slo/model_indicator_properties_timeslice_metric_params_metric_metrics_inner.go#L26-L68

This generated client simply attempts to unmarshal as the simplest type, `TimesliceMetricBasicMetricWithField`, **first** and if successful, leaves the other ones empty.  Unfortunately, `TimesliceMetricPercentileMetric` is a superset of `TimesliceMetricBasicMetricWithField`

* https://github.com/elastic/terraform-provider-elasticstack/blob/77c15d2b032ffa41bcb4d791b263cf6ee422b0cc/generated/slo/model_timeslice_metric_basic_metric_with_field.go#L20-L30
* https://github.com/elastic/terraform-provider-elasticstack/blob/77c15d2b032ffa41bcb4d791b263cf6ee422b0cc/generated/slo/model_timeslice_metric_percentile_metric.go#L21-L32

And is attempted _last_ .

As a result, https://github.com/lloydmeta/terraform-provider-elasticstack/blob/b4efb3ae031bdc4b72864c6086e22003371ab4e0/internal/kibana/slo_test.go#L372 [fails](https://github.com/elastic/terraform-provider-elasticstack/actions/runs/16168761300/job/45636885076#step:11:615) because it is the default of 0...

lloydmeta: Note that the multi-metric case shows this clearly:

https://github.com/elastic/terraform-provider-elasticstack/actions/runs/16168761300/job/45636885076#step:11:647

The `percentile` source from `TimesliceMetricPercentileMetric` is demarshalled as 0 hence TF thinks there's a change, but, fields that source from  `TimesliceMetricDocCountMetric` happen to work because it is _not_ super set of `TimesliceMetricBasicMetricWithField`.....

lloydmeta: Going to try to use the [discriminator](https://swagger.io/docs/specification/v3_0/data-models/inheritance-and-polymorphism/#discriminator) in the `slo-spec.yaml` file.

generated/slo-spec.yml
```diff
@@ -969,6 +969,18 @@ components:
...
+                      mapping:
+                        percentile: '#/components/schemas/timeslice_metric_percentile_metric'
+                        doc_count: '#/components/schemas/timeslice_metric_doc_count_metric'
+                        sum: '#/components/schemas/timeslice_metric_basic_metric_with_field'
+                        avg: '#/components/schemas/timeslice_metric_basic_metric_with_field'
+                        min: '#/components/schemas/timeslice_metric_basic_metric_with_field'
+                        max: '#/components/schemas/timeslice_metric_basic_metric_with_field'
+                        std_deviation: '#/components/schemas/timeslice_metric_basic_metric_with_field'
+                        last_value: '#/components/schemas/timeslice_metric_basic_metric_with_field'
+                        cardinality: '#/components/schemas/timeslice_metric_basic_metric_with_field'
```
lloydmeta, line 983: This resulted in generating an updated `IndicatorPropertiesTimesliceMetricParamsMetricMetricsInner` but also a bunch of other stuff

```sh
On branch feature/timeslice-metric-indicator-slo-support
Your branch is up to date with 'origin/feature/timeslice-metric-indicator-slo-support'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   generated/alerting/.openapi-generator/FILES
	modified:   generated/alerting/README.md
	modified:   generated/alerting/model_actions_inner_frequency.go
	modified:   generated/slo/model_slo_response_group_by.go

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	generated/alerting/model_notify_when_action.go
```

I ignored most of them.

examples/resources/elasticstack_kibana_slo/resource.tf
```diff
@@ -176,3 +176,38 @@ resource "elasticstack_kibana_slo" "custom_metric" {
   }
 
 }
+
+//Available from 8.12.0
```
lloydmeta, line 180: https://github.com/elastic/kibana/pull/168539

internal/kibana/slo.go
```diff
@@ -634,6 +714,60 @@ func getSloFromResourceData(d *schema.ResourceData) (models.Slo, diag.Diagnostic
...
+						Aggregation: agg,
+						Field:       metric["field"].(string),
+					},
+				}
+			case "percentile":
+				metrics[i] = slo.IndicatorPropertiesTimesliceMetricParamsMetricMetricsInner{
+					TimesliceMetricPercentileMetric: &slo.TimesliceMetricPercentileMetric{
+						Name:        metric["name"].(string),
+						Aggregation: agg,
+						Field:       metric["field"].(string),
```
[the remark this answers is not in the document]
lloydmeta: Going to push back on this one: I believe this validation is more suited to be handled in the Kibana server/backend. Doing it here is _nice_, but couples to implementation details that could change and potentially makes this provider more brittle.

internal/kibana/slo.go
```diff
@@ -634,6 +714,60 @@ func getSloFromResourceData(d *schema.ResourceData) (models.Slo, diag.Diagnostic
...
+						Percentile:  metric["percentile"].(float64),
```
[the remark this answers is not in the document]
lloydmeta: Going to push back on this one: I believe this validation is more suited to be handled in the Kibana server/backend. Doing it here is _nice_, but couples to implementation details that could change and potentially makes this provider more brittle.

internal/kibana/slo_test.go
```diff
@@ -237,6 +257,281 @@ func TestAccResourceSloGroupBy(t *testing.T) {
...
+							duration = "7d"
+							type     = "rolling"
+						}
+						depends_on = [elasticstack_elasticsearch_index.my_index]
+					}
+				`, sloName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("elasticstack_kibana_slo.test_slo", "timeslice_metric_indicator.0.index", "my-index"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_slo.test_slo", "timeslice_metric_indicator.0.timestamp_field", "@timestamp"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_slo.test_slo", "timeslice_metric_indicator.0.filter", ""),
```
lloydmeta, line 377: This one was interesting: AFAICT, this is being set to a `nil`-able value:

* https://github.com/elastic/terraform-provider-elasticstack/blob/bc991a826f41ad68f478416c27ba618e27122981/internal/kibana/slo.go#L1042
* https://github.com/elastic/terraform-provider-elasticstack/blob/bc991a826f41ad68f478416c27ba618e27122981/internal/kibana/slo.go#L760

This is in keeping with existing patterns in `slo.go`:
* https://github.com/elastic/terraform-provider-elasticstack/blob/bc991a826f41ad68f478416c27ba618e27122981/internal/kibana/slo.go#L627
* https://github.com/elastic/terraform-provider-elasticstack/blob/bc991a826f41ad68f478416c27ba618e27122981/internal/kibana/slo.go#L915-L922

But I _think_ that pattern doesn't quite work with `TestCheckNoResourceAttr` (I even tried only setting "filter" if it is not nil); perhaps somewhere it ends up getting a "empty" default value, whether in TF or in TF's test harness. Thus I'm using `TestCheckResourceAttr` and compare with an empty string.

generated/slo-spec.yml
tobio: So you hand edited the spec here? Are you able to make the same change to the spec in the Kibana repo so we can get this fixed upstream as well?

internal/kibana/slo_test.go
```diff
@@ -29,6 +31,9 @@ func TestAccResourceSlo(t *testing.T) {
 	slo8_10Constraints, err := version.NewConstraint(">=8.10.0,!=8.11.0,!=8.11.1,!=8.11.2,!=8.11.3,!=8.11.4")
 	require.NoError(t, err)
 
+	slo8_12Constraints, err := version.NewConstraint(">=" + sloTimesliceMetricsMinVersion.String())
```
tobio: It looks like we can just get rid of this contraint and use `sloTimesliceMetricsMinVersion` directly?

internal/kibana/slo_test.go
```diff
@@ -155,6 +160,21 @@ func TestAccResourceSlo(t *testing.T) {
 					resource.TestCheckResourceAttr("elasticstack_kibana_slo.test_slo", "tags.1", "another_tag"),
 				),
 			},
+			{
+				SkipFunc: versionutils.CheckIfVersionMeetsConstraints(slo8_12Constraints),
```
tobio:
```suggestion
				SkipFunc: versionutils.CheckIfVersionIsUnsupported(sloTimesliceMetricsMinVersion),
```

generated/slo-spec.yml
lloydmeta: Yeah I hand edited it :D I tried to find the spec in the Kibana repo but failed, sadly. Any pointers?

lloydmeta: Oh I think I may have found it: https://github.com/elastic/kibana/blob/main/x-pack/solutions/observability/plugins/slo/docs/openapi/slo/components/schemas/indicator_properties_timeslice_metric.yaml

dimuon: IMO we are safe to ingore `generated/alerting/*` in this PR. Though I'm curious why `generated/slo/model_slo_response_group_by.go` is modified. I guess it's because the last commit that touched `slo-spec.yml` ignored chages to `generated/slo/model_slo_response_group_by.go` Anyway, IMO we are safe to ignore the change here too as long as the client works as expected.

dimuon: Btw, thanks for adding this discriminator field. The spec should have it from the start.

lloydmeta: https://github.com/elastic/kibana/pull/227400

internal/kibana/slo_test.go
dimuon: Hm... I think it should not be set to an empty string though (ofc API can filter out empty filters so we can be safe in this case). Anyway we use `TestCheckNoResourceAttr` in tests when field is nullable, e.g. https://github.com/elastic/terraform-provider-elasticstack/blob/bc991a826f41ad68f478416c27ba618e27122981/internal/elasticsearch/security/role_test.go#L78. However, the code looks good so I can just speculate whether using `struct` instead of `map[string]interface{}` in https://github.com/elastic/terraform-provider-elasticstack/blob/bc991a826f41ad68f478416c27ba618e27122981/internal/kibana/slo.go#L1042 would fix the issue.

lloydmeta: Can you elaborate what you're suggesting by "using `struct` instead of map[string]interface{}"? 

https://github.com/elastic/terraform-provider-elasticstack/blob/bc991a826f41ad68f478416c27ba618e27122981/internal/elasticsearch/security/role.go#L523

AFAICT this is also using a similar thing

https://github.com/elastic/terraform-provider-elasticstack/blob/bc991a826f41ad68f478416c27ba618e27122981/internal/elasticsearch/security/role.go#L568-L572

I wonder if could be different because _that_ field is a `*bool`?  This is speculation on speculation at this point.

Curious if @tobio has any ideas, though, again, this same pattern is used throughout this file (and not tested !)

dimuon: I was thinking about something like `indicator = append(indicator, slo.IndicatorPropertiesTimesliceMetricParams{...})` but it doesn't work. 
Also, I didn't manage to spot any diff in a way we use `allow_restricted_indices` besides the type.

dimuon: Also, I added debug output to the end of `resourceSloRead` func: 
```
	v, ok := d.GetOk("timeslice_metric_indicator.0.filter")
	log.Info(ctx, fmt.Sprintf("timeslice_metric_indicator.0.filter = %v,%v", v, ok))
```
It prints this for `TestAccResourceSlo_timeslice_metric_indicator_percentile`:
```
timeslice_metric_indicator.0.filter = ,false
```
So I assume this is a problem of `TestCheckNoResourceAttr` (but it's still unclear why it works for `allow_restricted_indices`).

Issue #315, opened 2023-05-03
[Feature] Support Kibana APM agent-configuration


audig:
**Describe the resource you would like to have implemented.**
I would like to declare as terraform resource the default configuration for APM agent.
I would really appreciate to declare all my custom configuration ofr kibana & elasticsearch in this provider

**Describe the solution you'd like**
I would like to declare the resources like this 
```
resource "elasticstack_kibana_apm_agent_configuration" "mydefaultagentconfiguration" {
  #Value can be null for configure all services
  service_name           = "myapmservice"
  #Value can be null for configure all enviroments
  environment_name = "staging"
  transaction_max_spans = "100"
  transaction_sample_rate = "0.5"
}
```
**Describe alternatives you've considered**
I try to deal with apm agent configuration by ussing a rest api terraform provider (rest api mastercard). But the existing kibana api for agent-config-api doest not fit all requirements of the rest api provider

**Additional context**
There is an api in kibana to create, update and delete apm agent configuration : https://www.elastic.co/guide/en/kibana/current/agent-config-api.html

PR #1196, opened 2025-07-09
Add APM agent configuration

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/315

A couple of things:
* This is largely AI generated 🥳 
* Broken API specs are super annoying
* This resource is explicitly setting the X-ElasticAPIVersion header. I can pull this out, but that means more code in `transform_schema` which I don't really want to add to unless it's strictly required. When these endpoints are eventually moved to the automatically generated schema (as opposed to the hand crafted one in use today), the existing code handling this header will kick in and we can tidy it up.


docs/resources/apm_agent_configuration.md
```diff
@@ -0,0 +1,54 @@
...
+subcategory: "Kibana"
+layout: ""
+page_title: "Elasticstack: elasticstack_apm_agent_configuration Resource"
+description: |-
+  Creates or updates an APM agent configuration
+---
+
+# Resource: elasticstack_apm_agent_configuration
+
+Creates or updates a an APM agent configuration. See https://www.elastic.co/docs/solutions/observability/apm/apm-agent-central-configuration
```
dimuon:
```suggestion
Creates or updates an APM agent configuration. See https://www.elastic.co/docs/solutions/observability/apm/apm-agent-central-configuration
```

templates/resources/apm_agent_configuration.md.tmpl
```diff
@@ -0,0 +1,23 @@
...
+subcategory: "Kibana"
+layout: ""
+page_title: "Elasticstack: elasticstack_apm_agent_configuration Resource"
+description: |-
+  Creates or updates an APM agent configuration
+---
+
+# Resource: elasticstack_apm_agent_configuration
+
+Creates or updates a an APM agent configuration. See https://www.elastic.co/docs/solutions/observability/apm/apm-agent-central-configuration
```
dimuon:
```suggestion
Creates or updates an APM agent configuration. See https://www.elastic.co/docs/solutions/observability/apm/apm-agent-central-configuration
```

internal/apm/agent_configuration/resource_test.go
```diff
@@ -0,0 +1,77 @@
...
+func TestAccResourceAgentConfiguration(t *testing.T) {
+	serviceName := tf_acctest.RandStringFromCharSet(10, tf_acctest.CharSetAlphaNum)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acctest.PreCheck(t) },
+		ProtoV6ProviderFactories: acctest.Providers,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccResourceAgentConfigurationCreate(serviceName),
+				Check: resource.ComposeTestCheckFunc(
```
dimuon, line 21: IMO it makes sense to check `id` too.

internal/apm/agent_configuration/create.go
```diff
@@ -0,0 +1,65 @@
...
+	if diags := utils.CheckHttpErrorFromFW(apiResp, "Failed to create APM agent configuration"); diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+
+	id := plan.SetIDFromService()
+	tflog.Trace(ctx, fmt.Sprintf("Created APM agent configuration with ID: %s", id))
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+}
```
dimuon, line 71: I think some resources in this provider (maybe ones based on SDKv2) call `read` in scope of `create`/`update` calls but it seems we don't use this pattern here. IIRC, the EC provider also calls `read` inside `create`/`update` calls. Does it make sense to do this here too?

internal/apm/agent_configuration/delete.go
```diff
@@ -0,0 +1,61 @@
...
+	defer apiResp.Body.Close()
+
+	if diags := utils.CheckHttpErrorFromFW(apiResp, "Failed to delete APM agent configuration"); diags.HasError() {
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+
+	tflog.Trace(ctx, fmt.Sprintf("Deleted APM agent configuration with ID: %s", state.ID.ValueString()))
+
+	resp.State.RemoveResource(ctx)
```
dimuon: IIRC, it's not needed in `Delete` method.

internal/apm/agent_configuration/read.go
```diff
@@ -0,0 +1,94 @@
...
+		ctx,
+		&kbapi.GetAgentConfigurationsParams{
+			ElasticApiVersion: elasticAPIVersion,
+		},
+	)
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to get APM agent configurations", err.Error())
+		return
+	}
+
```
dimuon, line 55:
```suggestion
	defer apiResp.Body.Close()
```

internal/apm/agent_configuration/read.go
```diff
@@ -0,0 +1,94 @@
...
+		ctx,
+		&kbapi.GetAgentConfigurationsParams{
+			ElasticApiVersion: elasticAPIVersion,
+		},
+	)
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to get APM agent configurations", err.Error())
+		return
+	}
+	defer apiResp.Body.Close()
```
dimuon: Looks like `GetAgentConfigurationsWithResponse` returns own type (not `http.Response`) so GitHub Actions comment is correct and my previous suggestion is wrong. TL;DR - we need to remove line 37. Sorry for the confusion.

dimuon:
```suggestion
```

Issue #1179, opened 2025-07-04
[Feature] Add ability to add supports_agentless in resource elasticstack_fleet_agent_policy


Arnuzasz:
**Is your feature request related to a problem? Please describe.**
It is not possible to create e.g. AWS integration with agentless policy

**Describe the resource you would like to have implemented.**
kbn:/api/fleet/agent_policies

**Describe the solution you'd like**
In resource **elasticstack_fleet_agent_policy** add the ability to set **supports_agentless**

**Describe alternatives you've considered**
Neither writing a script nor creating it through the UI is a good alternative.

**Additional context**

PR #1197, opened 2025-07-09
Support the supports_agentless field in Fleet agent policies. 

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1179


dimuon (dismissed):

internal/clients/api_client.go
```diff
@@ -352,6 +354,24 @@ func (a *ApiClient) serverInfo(ctx context.Context) (*models.ClusterInfo, diag.D
 	return &info, diags
 }
 
+func (a *ApiClient) EnforceMinVersion(ctx context.Context, minVersion *version.Version) (bool, diag.Diagnostics) {
+	flavor, diags := a.ServerFlavor(ctx)
+	if diags.HasError() {
+		return false, diags
+	}
+
+	if flavor == ServerlessFlavor {
```
dimuon, line 363: IMO the func's behavior doesn't correspond to its name (or vice versa), e.g. if server flavor is "serverless" the func doesn't enforce min version.

Issue #631, opened 2024-05-15
[Feature] ignore_missing_component_templates 


cstegm:
https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/elasticsearch_index_template#nestedblock--template

**Is your feature request related to a problem? Please describe.**
I want to use the ignore_missing_component_templates feature when setting up an index template
as composed_of is already supported - ignore_missing_component_templates should be supported as well.

tobio:
Fixed in #1206

PR #1206, opened 2025-07-16
Adding ignore_missing_component_templates

togish:
https://github.com/elastic/terraform-provider-elasticstack/issues/631

Adding the ignore_missing_component_templates based on the composed_of key.

### DIFF
--- a/internal/elasticsearch/index/template.go
+++ b/internal/elasticsearch/index/template.go
@@ -38,6 +38,15 @@ func ResourceTemplate() *schema.Resource {
 				Type: schema.TypeString,
 			},
 		},
+		"ignore_missing_component_templates": {
+			Description: "A list of component template names that are ignored if missing.",
+			Type:        schema.TypeList,
+			Optional:    true,
+			Computed:    true,
+			Elem: &schema.Schema{
+				Type: schema.TypeString,
+			},
+		},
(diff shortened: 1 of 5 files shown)


internal/elasticsearch/index/template.go
```diff
@@ -221,6 +230,14 @@ func resourceIndexTemplatePut(ctx context.Context, d *schema.ResourceData, meta
 	}
 	indexTemplate.ComposedOf = compsOf
 
+	compsOfIgnore := make([]string, 0)
+	if v, ok := d.GetOk("ignore_missing_component_templates"); ok {
```
tobio, line 244: From the acceptance tests, this is only supported from 8.7 onwards. Are you able to add a version check on this attribute, something similar to what's in the [role resource](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/security/role.go#L384-L386)

togish: That is also what the API docs says more or less. I have changed the code in such way that the original tests are changed back to their original state then creating a new set that are ignored for the versions where the parameter are not supported.

tobio (dismissed):
Thanks @togish! This is looking 👌

tobio (approved):

PR #1209, opened 2025-07-18
Install go, tools, and dependencies for Copilot

tobio:


.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,114 @@
...
+      ELASTICSEARCH_PASSWORD: password
+      KIBANA_ENDPOINT: "http://localhost:5601"
+      KIBANA_USERNAME: "elastic"
+      KIBANA_PASSWORD: password
+      KIBANA_SYSTEM_USERNAME: kibana_system
+      KIBANA_SYSTEM_PASSWORD: password
+      TF_ACC: "1"
+    services:
+      elasticsearch:
+        image: docker.elastic.co/elasticsearch/elasticsearch:9.0.3
```
wandergeek, line 36: Do we want to parameterize this somehow?

tobio: AFAICT the only way to do that at this level is to configure a [repo level variable](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#creating-configuration-variables-for-a-repository), workflow level environment variables aren't accessible at this point (take a look at [this](https://github.com/elastic/terraform-provider-elasticstack/actions/runs/16408281626/job/46358057933) result). 

IMO having the stack version defined in some esoteric repo setting is way worse than duplicating it a couple of times within this single file 🤷.

PR #1224, opened 2025-08-01
[ResponseOps] Maintenance window resource

adcoelho:
Closes https://github.com/elastic/kibana/issues/197021

## Summary

This PR introduces a new Maintenance Window resource.
- Created with the plugin framework
- Using a client generated from Kibana's OpenAPI specification

### DIFF
--- a/generated/kbapi/transform_schema.go
+++ b/generated/kbapi/transform_schema.go
@@ -108,6 +108,7 @@ func (s Schema) MustGetPath(path string) *Path {
 type Path struct {
 	Parameters []Map `yaml:"parameters,omitempty"`
 	Get        Map   `yaml:"get,omitempty"`
+	Patch      Map   `yaml:"patch,omitempty"`
 	Post       Map   `yaml:"post,omitempty"`
 	Put        Map   `yaml:"put,omitempty"`
 	Delete     Map   `yaml:"delete,omitempty"`
@@ -123,6 +124,9 @@ func (p Path) Endpoints(yield func(key string, endpoint Map) bool) {
 	if p.Put != nil {
 		yield("put", p.Put)
 	}
+	if p.Patch != nil {
+		yield("patch", p.Patch)
+	}
 	if p.Delete != nil {
 		yield("delete", p.Delete)
 	}
@@ -136,6 +140,8 @@ func (p Path) GetEndpoint(method string) Map {
 		return p.Post
 	case "put":
 		return p.Put
+	case "patch":
+		return p.Patch
 	case "delete":
 		return p.Delete
 	default:
@@ -159,6 +165,8 @@ func (p *Path) SetEndpoint(method string, endpoint Map) {
 	case "post":
 		p.Post = endpoint
 	case "put":
+		p.Put = endpoint
+	case "patch":
 		p.Put = endpoint
 	case "delete":
 		p.Delete = endpoint
@@ -571,6 +579,8 @@ func transformFilterPaths(schema *Schema) {
 		"/api/synthetics/params":                         {"post"},
 		"/api/synthetics/params/{id}":                    {"get", "put", "delete"},
 		"/api/apm/settings/agent-configuration":          {"get", "put", "delete"},
+		"/api/maintenance_window":                        {"post"},
+		"/api/maintenance_window/{id}":                   {"delete", "get", "patch"},
 	}
 
 	for path, pathInfo := range schema.Paths {
@@ -717,6 +727,8 @@ func transformKibanaPaths(schema *Schema) {
 		"/api/data_views",
 		"/api/data_views/data_view",
 		"/api/data_views/data_view/{viewId}",
+		"/api/maintenance_window",
+		"/api/maintenance_window/{id}",
 	}
 
 	// Add a spaceId parameter if not already present
--- a/internal/kibana/maintenance_window/acc_test.go
+++ b/internal/kibana/maintenance_window/acc_test.go
@@ -1,0 +1,153 @@
+package maintenance_window_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/acctest"
+	"github.com/elastic/terraform-provider-elasticstack/internal/versionutils"
+	"github.com/hashicorp/go-version"
+	sdkacctest "github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+var minDataViewAPISupport = version.Must(version.NewVersion("8.1.0"))
+var minFullDataviewSupport = version.Must(version.NewVersion("8.8.0"))
+
+func TestAccResourceDataView(t *testing.T) {
+	indexName := "my-index-" + sdkacctest.RandStringFromCharSet(4, sdkacctest.CharSetAlphaNum)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { acctest.PreCheck(t) },
+		ProtoV6ProviderFactories: acctest.Providers,
+		Steps: []resource.TestStep{
+			{
+				SkipFunc: versionutils.CheckIfVersionIsUnsupported(minDataViewAPISupport),
+				Config:   testAccResourceDataViewPre8_8DV(indexName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet("elasticstack_kibana_data_view.dv", "id"),
+				),
+			},
+			{
+				SkipFunc: versionutils.CheckIfVersionIsUnsupported(minFullDataviewSupport),
+				Config:   testAccResourceDataViewBasicDV(indexName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet("elasticstack_kibana_data_view.dv", "id"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_data_view.dv", "override", "true"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_data_view.dv", "data_view.name", indexName),
+					resource.TestCheckResourceAttr("elasticstack_kibana_data_view.dv", "data_view.source_filters.#", "2"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_data_view.dv", "data_view.field_formats.event_time.id", "date_nanos"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_data_view.dv", "data_view.field_formats.machine.ram.params.pattern", "0,0.[000] b"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_data_view.dv", "data_view.runtime_field_map.runtime_shape_name.script_source", "emit(doc['shape_name'].value)"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_data_view.dv", "data_view.field_attrs.ingest_failure.custom_label", "error.ingest_failure"),
+				),
+			},
+			{
+				SkipFunc: versionutils.CheckIfVersionIsUnsupported(minFullDataviewSupport),
+				Config:   testAccResourceDataViewBasicDVUpdated(indexName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet("elasticstack_kibana_data_view.dv", "id"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_data_view.dv", "override", "false"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_data_view.dv", "data_view.name", indexName),
+					resource.TestCheckNoResourceAttr("elasticstack_kibana_data_view.dv", "data_view.source_filters"),
+					resource.TestCheckNoResourceAttr("elasticstack_kibana_data_view.dv", "data_view.field_formats"),
+					resource.TestCheckNoResourceAttr("elasticstack_kibana_data_view.dv", "data_view.runtime_field_map"),
+				),
+			},
+			{
+				SkipFunc:          versionutils.CheckIfVersionIsUnsupported(minFullDataviewSupport),
+				Config:            testAccResourceDataViewBasicDVUpdated(indexName),
+				ImportState:       true,
+				ImportStateVerify: true,
+				ResourceName:      "elasticstack_kibana_data_view.dv",
+			},
+		},
+	})
+}
+
+func testAccResourceDataViewPre8_8DV(indexName string) string {
+	return fmt.Sprintf(`
+provider "elasticstack" {
+	elasticsearch {}
+	kibana {}
+}
+
+resource "elasticstack_elasticsearch_index" "my_index" {
+	name                = "%s"
+	deletion_protection = false
+}
+
+resource "elasticstack_kibana_data_view" "dv" {
+	data_view = {
+	  title = "%s*"
+	}
+}`, indexName, indexName)
+}
+
+func testAccResourceDataViewBasicDV(indexName string) string {
+	return fmt.Sprintf(`
+provider "elasticstack" {
+	elasticsearch {}
+	kibana {}
+}
+
+resource "elasticstack_elasticsearch_index" "my_index" {
+	name                = "%s"
+	deletion_protection = false
+}
+
+resource "elasticstack_kibana_data_view" "dv" {
+	override = true
+	data_view = {
+		title           = "%s*"
+		name            = "%s"
+		time_field_name = "@timestamp"
+		source_filters  = ["event_time", "machine.ram"]
+		allow_no_index  = true
+		namespaces      = ["default", "foo", "bar"]
+		field_formats = {
+			event_time = {
+				id = "date_nanos"
+			}
+			"machine.ram" = {
+				id = "number"
+				params = {
+					pattern = "0,0.[000] b"
+				}
+			}
+		}
+		runtime_field_map = {
+			runtime_shape_name = {
+				type          = "keyword"
+				script_source = "emit(doc['shape_name'].value)"
+			}
+		}
+		field_attrs = {
+		  ingest_failure = { custom_label = "error.ingest_failure", count = 6 },
+		}
+	}
+}`, indexName, indexName, indexName)
+}
+
+func testAccResourceDataViewBasicDVUpdated(indexName string) string {
+	return fmt.Sprintf(`
+provider "elasticstack" {
+	elasticsearch {}
+	kibana {}
+}
+
+resource "elasticstack_elasticsearch_index" "my_index" {
+	name                = "%s"
+	deletion_protection = false
+}
+
+resource "elasticstack_kibana_data_view" "dv" {
+	override = false
+	data_view = {
+		title           = "%s*"
+		name            = "%s"
+		time_field_name = "@timestamp"
+		allow_no_index  = true
+	}
+}`, indexName, indexName, indexName)
+}
--- a/internal/kibana/maintenance_window/create.go
+++ b/internal/kibana/maintenance_window/create.go
@@ -1,0 +1,50 @@
+package maintenance_window
+
+import (
+	"context"
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients/kibana_oapi"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+func (r *MaintenanceWindowResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	// Retrieve values from plan
+	var planMaintenanceWindow MaintenanceWindowModel
+
+	diags := req.Plan.Get(ctx, &planMaintenanceWindow)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Generate API request body from plan
+	body, diags := planMaintenanceWindow.toAPICreateRequest(ctx)
+
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	client, err := r.client.GetKibanaOapiClient()
+	if err != nil {
+		resp.Diagnostics.AddError(err.Error(), "")
+		return
+	}
+
+	spaceID := planMaintenanceWindow.SpaceID.ValueString()
+	maintenanceWindowAPIResponse, diags := kibana_oapi.CreateMaintenanceWindow(ctx, client, spaceID, body)
+
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	diags = planMaintenanceWindow.fromAPICreateResponse(ctx, maintenanceWindowAPIResponse)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	diags = resp.State.Set(ctx, planMaintenanceWindow)
+	resp.Diagnostics.Append(diags...)
+}
--- a/internal/kibana/maintenance_window/delete.go
+++ b/internal/kibana/maintenance_window/delete.go
@@ -1,0 +1,28 @@
+package maintenance_window
+
+import (
+	"context"
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients/kibana_oapi"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+func (r *MaintenanceWindowResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	var stateModel MaintenanceWindowModel
+
+	diags := req.State.Get(ctx, &stateModel)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	client, err := r.client.GetKibanaOapiClient()
+	if err != nil {
+		resp.Diagnostics.AddError(err.Error(), "")
+		return
+	}
+
+	viewID, spaceID := stateModel.getMaintenanceWindowIDAndSpaceID()
+	diags = kibana_oapi.DeleteMaintenanceWindow(ctx, client, spaceID, viewID)
+	resp.Diagnostics.Append(diags...)
+}
--- a/internal/kibana/maintenance_window/models.go
+++ b/internal/kibana/maintenance_window/models.go
@@ -1,0 +1,446 @@
+package maintenance_window
+
+import (
+	"context"
+
+	"github.com/elastic/terraform-provider-elasticstack/generated/kbapi"
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+type MaintenanceWindowModel struct {
+	ID             types.String              `tfsdk:"id"`
+	SpaceID        types.String              `tfsdk:"space_id"`
+	Title          types.String              `tfsdk:"title"`
+	Enabled        types.Bool                `tfsdk:"enabled"`
+	CustomSchedule MaintenanceWindowSchedule `tfsdk:"custom_schedule"`
+	Scope          *MaintenanceWindowScope   `tfsdk:"scope"`
+}
+
+type MaintenanceWindowScope struct {
+	Alerting MaintenanceWindowAlertingScope `tfsdk:"alerting"`
+}
+
+type MaintenanceWindowAlertingScope struct {
+	Kql types.String `tfsdk:"kql"`
+}
+
+type MaintenanceWindowSchedule struct {
+	Start     types.String                        `tfsdk:"start"`
+	Duration  types.String                        `tfsdk:"duration"`
+	Timezone  types.String                        `tfsdk:"timezone"`
+	Recurring *MaintenanceWindowScheduleRecurring `tfsdk:"recurring"`
+}
+
+type MaintenanceWindowScheduleRecurring struct {
+	End         types.String  `tfsdk:"end"`
+	Every       types.String  `tfsdk:"every"`
+	Occurrences types.Float32 `tfsdk:"occurrences"`
+	OnWeekDay   types.List    `tfsdk:"on_week_day"`
+	OnMonthDay  types.List    `tfsdk:"on_month_day"`
+	OnMonth     types.List    `tfsdk:"on_month"`
+}
+
+/* CREATE */
+
+func (model MaintenanceWindowModel) toAPICreateRequest(ctx context.Context) (kbapi.PostMaintenanceWindowJSONRequestBody, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	body := kbapi.PostMaintenanceWindowJSONRequestBody{
+		Enabled: model.Enabled.ValueBoolPointer(),
+		Title:   *model.Title.ValueStringPointer(),
+	}
+
+	body.Schedule.Custom.Duration = model.CustomSchedule.Duration.ValueString()
+	body.Schedule.Custom.Start = model.CustomSchedule.Start.ValueString()
+
+	if !model.CustomSchedule.Timezone.IsNull() && !model.CustomSchedule.Timezone.IsUnknown() {
+		body.Schedule.Custom.Timezone = model.CustomSchedule.Timezone.ValueStringPointer()
+	}
+
+	if model.CustomSchedule.Recurring != nil {
+
+		body.Schedule.Custom.Recurring = &struct {
+			End         *string    `json:"end,omitempty"`
+			Every       *string    `json:"every,omitempty"`
+			Occurrences *float32   `json:"occurrences,omitempty"`
+			OnMonth     *[]float32 `json:"onMonth,omitempty"`
+			OnMonthDay  *[]float32 `json:"onMonthDay,omitempty"`
+			OnWeekDay   *[]string  `json:"onWeekDay,omitempty"`
+		}{
+			End:   model.CustomSchedule.Recurring.End.ValueStringPointer(),
+			Every: model.CustomSchedule.Recurring.Every.ValueStringPointer(),
+		}
+
+		if !model.CustomSchedule.Recurring.Occurrences.IsNull() && !model.CustomSchedule.Recurring.Occurrences.IsUnknown() && model.CustomSchedule.Recurring.Occurrences.ValueFloat32() > 0 {
+			body.Schedule.Custom.Recurring.Occurrences = model.CustomSchedule.Recurring.Occurrences.ValueFloat32Pointer()
+		}
+
+		if !model.CustomSchedule.Recurring.OnWeekDay.IsNull() && !model.CustomSchedule.Recurring.OnWeekDay.IsUnknown() {
+			var onWeekDay []string
+			diags = model.CustomSchedule.Recurring.OnWeekDay.ElementsAs(ctx, &onWeekDay, true)
+			body.Schedule.Custom.Recurring.OnWeekDay = &onWeekDay
+		}
+
+		if !model.CustomSchedule.Recurring.OnMonth.IsNull() && !model.CustomSchedule.Recurring.OnMonth.IsUnknown() {
+			var onMonth []float32
+			diags = model.CustomSchedule.Recurring.OnMonth.ElementsAs(ctx, &onMonth, true)
+			body.Schedule.Custom.Recurring.OnMonth = &onMonth
+		}
+
+		if !model.CustomSchedule.Recurring.OnMonthDay.IsNull() && !model.CustomSchedule.Recurring.OnMonthDay.IsUnknown() {
+			var onMonthDay []float32
+			diags = model.CustomSchedule.Recurring.OnMonthDay.ElementsAs(ctx, &onMonthDay, true)
+			body.Schedule.Custom.Recurring.OnMonthDay = &onMonthDay
+		}
+	}
+
+	if model.Scope != nil {
+		// Yes, I hate it too
+		body.Scope = &struct {
+			Alerting struct {
+				Query struct {
+					Kql string `json:"kql"`
+				} `json:"query"`
+			} `json:"alerting"`
+		}{
+			Alerting: struct {
+				Query struct {
+					Kql string `json:"kql"`
+				} `json:"query"`
+			}{
+				Query: struct {
+					Kql string `json:"kql"`
+				}{
+					Kql: model.Scope.Alerting.Kql.ValueString(),
+				},
+			},
+		}
+	}
+
+	return body, diags
+}
+
+func (model *MaintenanceWindowModel) fromAPICreateResponse(ctx context.Context, data *kbapi.PostMaintenanceWindowResponse) diag.Diagnostics {
+	if data == nil {
+		return nil
+	}
+
+	response := data.JSON200
+
+	var diags diag.Diagnostics
+
+	resourceID := clients.CompositeId{
+		ClusterId:  model.SpaceID.ValueString(),
+		ResourceId: response.Id,
+	}
+
+	model.ID = types.StringValue(resourceID.String())
+	model.Title = types.StringValue(response.Title)
+	model.Enabled = types.BoolValue(response.Enabled)
+
+	model.CustomSchedule = MaintenanceWindowSchedule{
+		Start:    types.StringValue(response.Schedule.Custom.Start),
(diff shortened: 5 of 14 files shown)


docs/resources/kibana_maintenance_window.md
```diff
@@ -0,0 +1,72 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "elasticstack_kibana_maintenance_window Resource - terraform-provider-elasticstack"
+subcategory: ""
+description: |-
+  Manages Kibana data views
+---
+
+# elasticstack_kibana_maintenance_window (Resource)
```
gigerdo: To have a nicer docs file here, a template like this one can be added:
https://github.com/elastic/terraform-provider-elasticstack/blob/main/templates/resources/kibana_data_view.md.tmpl

This way you can add a description of this resource and optionally add examples if those are helpful.

generated/kbapi/transform_schema.go
```diff
@@ -160,6 +166,8 @@ func (p *Path) SetEndpoint(method string, endpoint Map) {
 		p.Post = endpoint
 	case "put":
 		p.Put = endpoint
+	case "patch":
+		p.Put = endpoint
```
gigerdo:
```suggestion
		p.Patch = endpoint
```

internal/kibana/maintenance_window/schema.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func (r *MaintenanceWindowResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Manages Kibana data views",
```
gigerdo: This description still references the data view schema.

internal/kibana/maintenance_window/schema.go
```diff
@@ -0,0 +1,146 @@
...
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "Generated ID for the data view.",
```
gigerdo: Also still references the wrong resource

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,375 @@
...
+				} `json:"query"`
+			}{
+				Query: struct {
+					Kql string `json:"kql"`
+				}{
+					Kql: model.Scope.Alerting.Kql.ValueString(),
+				},
+			},
+		}
+	}
```
gigerdo: (Just a comment/discussion, no feedback to address)

The reason behind this is that the openapi spec defines this model inline without a ref/name. So in Go these are also generated as anonymous types.

I can't think of any way to have types here, except changing the spec coming from Kibana.

@tobio @dimuon I was wondering if you had any ideas on how to improve the typing here.

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,375 @@
...
+			OnMonthDay: types.ListNull(types.Int32Type),
+		}
+
+		if response.Schedule.Custom.Recurring.Occurrences != nil {
+			occurrences := int32(*response.Schedule.Custom.Recurring.Occurrences)
+			model.CustomSchedule.Recurring.Occurrences = types.Int32PointerValue(&occurrences)
+		}
+
+		if response.Schedule.Custom.Recurring.OnWeekDay != nil {
+			onWeekDay, _ := types.ListValueFrom(ctx, types.StringType, response.Schedule.Custom.Recurring.OnWeekDay)
```
gigerdo: Are the potential errors here intentionally ignored?

internal/kibana/maintenance_window/response_types.go
```diff
@@ -0,0 +1,49 @@
...
+	Duration  string                 `json:"duration"`
+	Recurring *ResponseJsonRecurring `json:"recurring,omitempty"`
+	Start     string                 `json:"start"`
+	Timezone  *string                `json:"timezone,omitempty"`
+}
+
+type ResponseJsonRecurring struct {
+	End         *string    `json:"end,omitempty"`
+	Every       *string    `json:"every,omitempty"`
+	Occurrences *float32   `json:"occurrences,omitempty"`
```
gigerdo, line 33: This is always mapped to `int32`. Is this due to how the spec is defined? I think technically there is also an `integer` type in the spec, but that would also need some upstream change in the Kibana spec.

internal/kibana/maintenance_window/create.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"context"
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients/kibana_oapi"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+func (r *MaintenanceWindowResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	// Retrieve values from plan
```
gigerdo, line 14: It could be useful to have a version check here. (Theoretically it should be enough to check this when creating)

Something like this:
https://github.com/elastic/terraform-provider-elasticstack/blob/ec089b4c8dea548f0607b8ba148d6359153e0857/internal/kibana/role.go#L305-L307

(Btw. does this resource also work in serverless? If yes, that might be something to test when adding such a check, to ensure it works)

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,375 @@
...
+		}
+
+		if !model.CustomSchedule.Recurring.Occurrences.IsNull() && !model.CustomSchedule.Recurring.Occurrences.IsUnknown() && model.CustomSchedule.Recurring.Occurrences.ValueInt32() > 0 {
+			occurrences := float32(model.CustomSchedule.Recurring.Occurrences.ValueInt32())
+			body.Schedule.Custom.Recurring.Occurrences = &occurrences
+		}
+
+		if !model.CustomSchedule.Recurring.OnWeekDay.IsNull() && !model.CustomSchedule.Recurring.OnWeekDay.IsUnknown() {
+			var onWeekDay []string
+			diags = model.CustomSchedule.Recurring.OnWeekDay.ElementsAs(ctx, &onWeekDay, true)
```
gigerdo: I think this will override the diags so if you have multiple errors you would only see the last one. To accumulate this snippet works:

```
diags.Append(model.CustomSchedule.Recurring.OnWeekDay.ElementsAs(ctx, &onWeekDay, true)...)
```

gigerdo (commented):
Thank you for writing this resource. Great work! 🚀 

I left some smaller nits and comments, but looks great in general.

internal/kibana/maintenance_window/models.go
adcoelho: (We discussed this offline already, but I'll leave it here for everyone else.)

My team has an upcoming OnWeek, and one of the things I am planning to look into is this. Not all resources have this problem, so there is something in the way the schema variables are defined in Kibana that leads to better-organized, automatically generated types.

In this case, I suspect this happens because in Kibana we do not have a separate variable for the `Scope` type used in the POST body type definition.

One of the things I suspect(not directly connected to the example you commented on) is that although the type is the same for the PATCH and POST MW responses, we use different variables. That may lead to different response objects here.

PS.: I left the comment here hoping for this discussion, I will remove it now ;)

internal/kibana/maintenance_window/response_types.go
adcoelho: I was wondering the same. In Kibana, this is defined as `number`, which always leads to a `float32` generated type.

I couldn't figure out how to change the type upstream, but I thought the conversion from one to the other wouldn't be a problem.

wdyt?

internal/kibana/maintenance_window/create.go
adcoelho: But `serverVersion` is part of the other client. Should I use both here, one for version checks and the other for the actual API calls?

internal/kibana/maintenance_window/models.go
adcoelho: changed here and in the places below, thank you ;)

internal/kibana/maintenance_window/models.go
adcoelho: fixed in [eb56228](https://github.com/elastic/terraform-provider-elasticstack/pull/1224/commits/eb56228453cd0c1e3a11cdb927d5fa5b25ca54e4), would you please double check?

docs/resources/kibana_maintenance_window.md
adcoelho: done in [caebd0b](https://github.com/elastic/terraform-provider-elasticstack/pull/1224/commits/caebd0b9e8ff88751ec84ee74e35cbd65e1968ae)

internal/kibana/maintenance_window/response_types.go
gigerdo: Yeah seems fine to leave it as it is.

internal/kibana/maintenance_window/create.go
gigerdo: Yes I think that is fine.

adcoelho: > But serverVersion is part of the other client. Should I use both here, one for version checks and the other for the actual API calls?

I was wrong in this assumption; the version checks are part of the APIClient to which we already have access at this point. I added a version and a serverless check for all 4 operations in this commit:  [d731442](https://github.com/elastic/terraform-provider-elasticstack/pull/1224/commits/d731442ea661eb2fc19488b26b90142f27939665)

examples/resources/elasticstack_kibana_maintenance_window/resource.tf
```diff
@@ -0,0 +1,27 @@
+provider "elasticstack" {
+  elasticsearch {}
+  kibana {}
+}
+
+resource "elasticstack_kibana_maintenance_window" "my_maintenance_window" {
+  title   = "UPDATE TEST"
+  enabled = true
+
+  custom_schedule {
```
gigerdo: This example will not work, the syntax here is to use an =: `custom_schedule = {` (same for the other configs)

examples/resources/elasticstack_kibana_maintenance_window/import.sh
```diff
@@ -0,0 +1 @@
+terraform import elasticstack_kibana_maintenance_window.my_maintenance_window <space id>/<maintenance window id>
```
gigerdo, line 1: This command will not work for the resource as it doesn't implement the import function. 

Terraform import is not supported by default, it has to be implemented in the resource:

https://github.com/elastic/terraform-provider-elasticstack/blob/4e803b236197eb784d94c4af1e464d012f595f25/internal/kibana/data_view/resource.go#L37-L53

internal/kibana/maintenance_window/read.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	client, err := r.client.GetKibanaOapiClient()
+	if err != nil {
+		resp.Diagnostics.AddError(err.Error(), "")
+		return
+	}
+
+	viewID, spaceID := stateModel.getMaintenanceWindowIDAndSpaceID()
+	dataView, diags := kibana_oapi.GetMaintenanceWindow(ctx, client, spaceID, viewID)
```
gigerdo: nit: rename `dataView`

internal/kibana/maintenance_window/schema.go
```diff
@@ -0,0 +1,146 @@
...
+						Validators: []validator.String{
+							StringIsAlertingDuration{},
+						},
+					},
+					"timezone": schema.StringAttribute{
+						Description: "The timezone of the schedule. The default timezone is UTC.",
+						Optional:    true,
+						Computed:    true,
+					},
+					"recurring": schema.SingleNestedAttribute{
```
gigerdo, line 76: I was trying the resource a bit and I think there is an issue with the definition of the this field:
It is possible to define it empty `recurring = {}` but that will cause a bug:

```
│
│ When applying changes to elasticstack_kibana_maintenance_window.test_maintenance_window2, provider "provider[\"registry.terraform.io/elastic/elasticstack\"]" produced an unexpected
│ new value: .custom_schedule.recurring: was cty.ObjectVal(map[string]cty.Value{"end":cty.NullVal(cty.String), "every":cty.NullVal(cty.String), "occurrences":cty.NullVal(cty.Number),
│ "on_month":cty.NullVal(cty.List(cty.Number)), "on_month_day":cty.NullVal(cty.List(cty.Number)), "on_week_day":cty.NullVal(cty.List(cty.String))}), but now null.
│
```

This is happening because the input defined `recurring`, but then the new state after the update does not define it (=null). Since this field is required, the fix is to always set `recurring` in the state model when updating it (currently it is only set if the response contains a `recurring`):

https://github.com/elastic/terraform-provider-elasticstack/blob/eb56228453cd0c1e3a11cdb927d5fa5b25ca54e4/internal/kibana/maintenance_window/models.go#L334-L335

internal/kibana/maintenance_window/models.go
dimuon: I guess Kibana specs is the first candidate for improvement. However, perhaps @jamietanna could share his thoughts on forcing oapi-codegen to generate types in this situation. (sorry for the direct ping Jamie).

tobio: There's been some discussion around improving this as part of https://github.com/elastic/kibana/issues/228077, I assume that won't happen in the timeframe we'd like to improve the automation here. I suspect the simplest short term path here might be to extract these anonymous types in a script prior to passing the spec to oapi-codegen.

jamietanna: > I suspect the simplest short term path here might be to extract these anonymous types in a script prior to passing the spec to oapi-codegen.

Correct - right now, `oapi-codegen` will not generate these anonymous types.

One option we've got is that we may be able to use [OpenAPI Overlay](https://github.com/oapi-codegen/oapi-codegen?tab=readme-ov-file#modifying-the-input-openapi-specification-with-openapi-overlay) which may be a little more "scalable" as a solution to make the extraction a little more straightforward to maintain

(Aside: I'm leaving Elastic on 2025-09-05, so you'll be best raising things like this with me via the public `oapi-codegen` issue tracker in the future - or DM me on Slack in the next few weeks to discuss alternatives)

gigerdo: Do you mean in a transform script as we already have here: https://github.com/elastic/terraform-provider-elasticstack/blob/6b34f127309413a25a7f3073f74da99a37656c67/generated/kbapi/transform_schema.go#L826 ?

My main concern is that then we have additional code to maintain there. (Looking at the existing code, I can't easily figure out what it is achieving. But probably it is also fixing some issues with the upstream spec). It seems more easily maintainable to have the ugly instantiation logic as in this PR.

tobio: > It seems more easily maintainable to have the ugly instantiation logic as in this PR.

There's ugly instantiation logic (like in this PR), but very similar spec code results in an unusable generated client (https://github.com/elastic/kibana/issues/197153). There's oapi-codegen issues (e.g https://github.com/oapi-codegen/oapi-codegen/issues/2018) around this, but we need a path forward now. 

None of this should be a blocker for this PR, but there seems to be three options here (in order of preference):
* oapi-codegen generates a valid client for anonymous polymorphic types
* Kibana generates a spec using only named types within a polymorphic container
* The provider rewrites the Kibana spec to only use names types within a polymorphic container

Why does this relate to this anonymous type? Well if we're going to use only named types in some specific situations then it's not much more effort to use named types in all situations since it avoid this ugliness (it's also really hard to figure out compile time errors if the types don't match exactly so it's not just ugliness). IMO, ideally the Kibana spec would avoid these anonymous types altogether, even if that happens in the future though the provider likely needs to solve this problem sooner. Maybe it's worth looking into solving it within the Kibana repo (I haven't looked at their spec generator at all), but I suspect it's easier to do so within `transform_schema` in the short term.

tobio: > One option we've got is that we may be able to use [OpenAPI Overlay](https://github.com/oapi-codegen/oapi-codegen?tab=readme-ov-file#modifying-the-input-openapi-specification-with-openapi-overlay) which may be a little more "scalable" as a solution to make the extraction a little more straightforward to maintain

@jamietanna FWIW the JSON path library used in oapi-codegen doesn't support at least some valid selectors (e.g `$..properties[?(@.nullable==true)]`). I'd like to look at widening the support if I had some time, but at the moment it's a non-starter for us.

internal/kibana/maintenance_window/create.go
```diff
@@ -0,0 +1,65 @@
...
+
+	serverFlavor, sdkDiags := r.client.ServerFlavor(ctx)
+	if sdkDiags.HasError() {
+		return
+	}
+
+	diags = validateMaintenanceWindowServer(serverVersion, serverFlavor)
+	if diags.HasError() {
+		return
+	}
```
tobio:
```suggestion
	isSupported, sdkDiags := r.client.EnforceMinVersion(ctx, version.Must(version.NewVersion("9.1.0")))
	resp.Diagnostics.Append(utils.FrameworkDiagsFromSDK(diags)...)
	if resp.Diagnostics.HasError() {
		return
	}
	
	if !isSupported {
		resp.Diagnostics.AddError("Unsupported server version", "Maintenance windows are not supported until Elastic Stack v9.0. Upgrade the target server to use this resource")
		return
	}
```

EnforceMinVersion will bundle the version/serverless check.

internal/kibana/maintenance_window/delete.go
```diff
@@ -0,0 +1,43 @@
...
+	if sdkDiags.HasError() {
+		return
+	}
+
+	serverFlavor, sdkDiags := r.client.ServerFlavor(ctx)
+	if sdkDiags.HasError() {
+		return
+	}
+
+	diags = validateMaintenanceWindowServer(serverVersion, serverFlavor)
```
tobio: We've typically only checked the server version during creations (or updates if it relates to a specific attribute). It's possible that a user has changed the underlying stack, and downgraded to an unsupported version, but it's pretty unlikely and the delete would result in a 404 which would then correctly result in the resource being removed from state. 

I don't mind leaving this in, but I don't think it's required.

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,388 @@
...
+}
+
+/* CREATE */
+
+func (model MaintenanceWindowModel) toAPICreateRequest(ctx context.Context) (kbapi.PostMaintenanceWindowJSONRequestBody, diag.Diagnostics) {
+	var diags = diag.Diagnostics{}
+
+	body := kbapi.PostMaintenanceWindowJSONRequestBody{
+		Enabled: model.Enabled.ValueBoolPointer(),
+		Title:   *model.Title.ValueStringPointer(),
```
tobio:
```suggestion
		Title:   model.Title.ValueString(),
```

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,388 @@
...
+			Occurrences *float32   `json:"occurrences,omitempty"`
+			OnMonth     *[]float32 `json:"onMonth,omitempty"`
+			OnMonthDay  *[]float32 `json:"onMonthDay,omitempty"`
+			OnWeekDay   *[]string  `json:"onWeekDay,omitempty"`
+		}{
+			End:   model.CustomSchedule.Recurring.End.ValueStringPointer(),
+			Every: model.CustomSchedule.Recurring.Every.ValueStringPointer(),
+		}
+
+		if !model.CustomSchedule.Recurring.Occurrences.IsNull() && !model.CustomSchedule.Recurring.Occurrences.IsUnknown() && model.CustomSchedule.Recurring.Occurrences.ValueInt32() > 0 {
```
tobio:
```suggestion
		if utils.IsKnown(model.CustomSchedule.Recurring.Occurrences) {
```

IMO the `ValueInt32() > 0` check shouldn't be here. The attribute is being validated within the schema, ignoring an invalid value here feels like a potential bug (somehow a negative `occurrences` becomes meaningful but we forget to remove this check) rather than a feature.

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,388 @@
...
+			occurrences := float32(model.CustomSchedule.Recurring.Occurrences.ValueInt32())
```
tobio: Can we open an issue for Kibana to update this field to include `format: int32` so the type is correctly defined?

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,388 @@
...
+			body.Schedule.Custom.Recurring.Occurrences = &occurrences
+		}
+
+		if !model.CustomSchedule.Recurring.OnWeekDay.IsNull() && !model.CustomSchedule.Recurring.OnWeekDay.IsUnknown() {
```
tobio:
```suggestion
		if utils.IsKnown(model.CustomSchedule.Recurring.OnWeekDay) {
```

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,388 @@
...
+			var onWeekDay []string
+			diags.Append(model.CustomSchedule.Recurring.OnWeekDay.ElementsAs(ctx, &onWeekDay, true)...)
+			body.Schedule.Custom.Recurring.OnWeekDay = &onWeekDay
+		}
+
+		if !model.CustomSchedule.Recurring.OnMonth.IsNull() && !model.CustomSchedule.Recurring.OnMonth.IsUnknown() {
```
tobio:
```suggestion
		if utils.IsKnown(model.CustomSchedule.Recurring.OnMonth) {
```

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,388 @@
...
+			var onMonth []float32
+			diags.Append(model.CustomSchedule.Recurring.OnMonth.ElementsAs(ctx, &onMonth, true)...)
+			body.Schedule.Custom.Recurring.OnMonth = &onMonth
+		}
+
+		if !model.CustomSchedule.Recurring.OnMonthDay.IsNull() && !model.CustomSchedule.Recurring.OnMonthDay.IsUnknown() {
```
tobio:
```suggestion
		if utils.IsKnown(model.CustomSchedule.Recurring.OnMonthDay) {
```

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,388 @@
...
+		}
+
+		if !model.CustomSchedule.Recurring.OnWeekDay.IsNull() && !model.CustomSchedule.Recurring.OnWeekDay.IsUnknown() {
+			var onWeekDay []string
+			diags.Append(model.CustomSchedule.Recurring.OnWeekDay.ElementsAs(ctx, &onWeekDay, true)...)
+			body.Schedule.Custom.Recurring.OnWeekDay = &onWeekDay
+		}
+
+		if !model.CustomSchedule.Recurring.OnMonth.IsNull() && !model.CustomSchedule.Recurring.OnMonth.IsUnknown() {
+			var onMonth []float32
```
tobio: Similarly these fields should be `format: int32` in the Kibana spec.

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,388 @@
...
+			diags.Append(model.CustomSchedule.Recurring.OnMonth.ElementsAs(ctx, &onMonth, true)...)
+			body.Schedule.Custom.Recurring.OnMonth = &onMonth
+		}
+
+		if !model.CustomSchedule.Recurring.OnMonthDay.IsNull() && !model.CustomSchedule.Recurring.OnMonthDay.IsUnknown() {
+			var onMonthDay []float32
```
tobio: Similarly these fields should be `format: int32` in the Kibana spec.

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,388 @@
...
+
+	body := kbapi.PatchMaintenanceWindowIdJSONRequestBody{}
+
+	if !model.Enabled.IsNull() {
+		body.Enabled = model.Enabled.ValueBoolPointer()
+	}
+
+	if !model.Title.IsNull() {
+		body.Title = model.Title.ValueStringPointer()
+	}
```
tobio:
```suggestion
	body := kbapi.PatchMaintenanceWindowIdJSONRequestBody{
		Enabled: model.Enabled.ValueBoolPointer(),
		Title: model.Title.ValueStringPointer(),
	}
```

Any reason we can't just assign these directly?

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,388 @@
...
+			Timezone *string `json:"timezone,omitempty"`
+		}{
+			Duration: model.CustomSchedule.Duration.ValueString(),
+			Start:    model.CustomSchedule.Start.ValueString(),
+		},
+	}
+
+	body.Schedule = &schedule
+
+	if !model.CustomSchedule.Timezone.IsNull() && !model.CustomSchedule.Timezone.IsUnknown() {
```
tobio:
```suggestion
	if utils.IsKnown(model.CustomSchedule.Timezone) {
```

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,388 @@
...
+
+	if !model.Enabled.IsNull() {
+		body.Enabled = model.Enabled.ValueBoolPointer()
+	}
+
+	if !model.Title.IsNull() {
+		body.Title = model.Title.ValueStringPointer()
+	}
+
+	schedule := struct {
```
tobio: Maybe I'm missing something, the schedule looks identical between the create/update paths. Are we able to re-use the code here?

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,388 @@
...
+		}
+
+		if !model.CustomSchedule.Recurring.OnMonthDay.IsNull() && !model.CustomSchedule.Recurring.OnMonthDay.IsUnknown() {
+			var onMonthDay []float32
+			diags.Append(model.CustomSchedule.Recurring.OnMonthDay.ElementsAs(ctx, &onMonthDay, true)...)
+			body.Schedule.Custom.Recurring.OnMonthDay = &onMonthDay
+		}
+	}
+
+	if model.Scope != nil {
```
tobio: Same with scope re create/update

internal/kibana/maintenance_window/schema.go
```diff
@@ -0,0 +1,146 @@
...
+				Description: "Whether the current maintenance window is enabled.",
+				Optional:    true,
+				Computed:    true,
+				Default:     booldefault.StaticBool(false),
+			},
+			"custom_schedule": schema.SingleNestedAttribute{
+				Description: "A set schedule over which the maintenance window applies.",
+				Required:    true,
+				Attributes: map[string]schema.Attribute{
+					"start": schema.StringAttribute{
```
tobio, line 57: It would be nice to have some validation on this attribute. 

The 'right' way would be to define an ISO8601 string type (Hashicorp have an [RFC3339](https://github.com/hashicorp/terraform-plugin-framework-timetypes/blob/main/timetypes/rfc3339_type.go) type you could use as an example). Having a validation function like `StringIsAlertingDuration` would be fine as well.

One benefit of the custom type that might be applicable is the ability to mark two string dates as semantically equal (e.g `2025-03-12T12:00:00` and ``2025-03-12T12:00:00.000Z`) such that they don't get marked as requiring an update in a subsequent plan.

internal/kibana/maintenance_window/schema.go
```diff
@@ -0,0 +1,146 @@
...
+					"timezone": schema.StringAttribute{
+						Description: "The timezone of the schedule. The default timezone is UTC.",
+						Optional:    true,
+						Computed:    true,
+					},
+					"recurring": schema.SingleNestedAttribute{
+						Description: "A set schedule over which the maintenance window applies.",
+						Required:    true,
+						Attributes: map[string]schema.Attribute{
+							"end": schema.StringAttribute{
```
tobio, line 80: Same as `start`, it'd be nice to validate these attributes.

internal/kibana/maintenance_window/schema.go
```diff
@@ -0,0 +1,146 @@
...
+								Description: "The start date and time of the schedule, provided in ISO 8601 format and set to the UTC timezone. For example: `2025-03-12T12:00:00.000Z`.",
```
tobio:
```suggestion
								Description: "The end date and time of the schedule, provided in ISO 8601 format and set to the UTC timezone. For example: `2025-03-12T12:00:00.000Z`.",
```

?

internal/kibana/maintenance_window/create.go
```diff
@@ -0,0 +1,65 @@
...
+
+	diags = planMaintenanceWindow.fromAPICreateResponse(ctx, maintenanceWindowAPIResponse)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	diags = resp.State.Set(ctx, planMaintenanceWindow)
+	resp.Diagnostics.Append(diags...)
```
tobio, line 82: In create/update paths we typically have followed up the write operation with a read, and then set the state from the read. 

There's a lot of cases where we've gotten slightly different responses between the two calls and so have avoided a dirty plan immediately after an apply.

internal/kibana/maintenance_window/validators.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func StringMatchesAlertingDurationRegex(s string) (matched bool, err error) {
+	pattern := "^[1-9][0-9]*(?:d|h|m|s)$"
+
+	return regexp.MatchString(pattern, s)
+}
+
+type StringIsAlertingDuration struct{}
```
tobio: We have `stringIsAlertingDuration` in `internal/kibana/alerting.go`. Can we move both these duplicate functions to a new `utils/validators` package?

internal/kibana/maintenance_window/models.go
gigerdo: Having the Kibana spec be generated with named refs is the way to go long-term. I think the overlay files could also be a good temporary solution, it seems more readable compared to the programmatic transformers.

tobio: > I think the overlay files could also be a good temporary solution

I mean in theory, sure. They don't appear to support the selectors we need to make use of though.

internal/kibana/maintenance_window/delete.go
adcoelho: Makes sense, removed.

internal/kibana/maintenance_window/models.go
adcoelho: The types generated for the client differ `PatchMaintenanceWindowIdJSONRequestBody` vs `PostMaintenanceWindowIdJSONRequestBody`.

I definitely agree, and I would also rather have a single function, but I could not get around the type errors with my GoLang skills(or lack thereof).

internal/kibana/maintenance_window/models.go
adcoelho: We use the `@kbn/config-schema` package for the schema/types in this API. This field in particular is defined as `schema.number` [here](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/alerting/common/routes/schedule/schema/v1.ts#L106).

That leads to `occurences` being:
```
occurrences:
  description: The total number of recurrences of the schedule.
  minimum: 1
  type: number
```

I don't think there is currently a way for the `format` field to be set.

Assuming this is what you meant in your comment(and I didn't get this completely wrong), what I can do is create a follow-up ticket to extend the `TypeMeta` [interface](https://github.com/elastic/kibana/blob/main/src/platform/packages/shared/kbn-config-schema/src/types/type.ts#L27) in `config-schema`.

We could have an optional `format` field here that would then end up in the generated docs.

wdyt @jloleysens ?

tobio: Yep, that's exactly what I meant. I was slightly wrong about the OAS change required here though, we'd need `type: integer` instead of `type: number`. 

From a consumers perspective I think there's sufficient value in supporting both `number` and `integer` types, the overhead on the Kibana side is relatively low. Supporting the `format` attribute (would allow `int32|int64` and `float32|float64` modifiers) would be nice, but perhaps less important.

internal/kibana/maintenance_window/models.go
tobio: Something like [this](https://github.com/elastic/terraform-provider-elasticstack/commit/d06c09a2b5b5db33bcc68a42d4017fd9df4e6a08) should work.

internal/kibana/maintenance_window/acc_test.go
```diff
@@ -0,0 +1,116 @@
+package maintenance_window_test
+
+import (
+	"testing"
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/acctest"
+	"github.com/elastic/terraform-provider-elasticstack/internal/versionutils"
+	"github.com/hashicorp/go-version"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
```
tobio:
```suggestion
	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
```

Changed with https://github.com/elastic/terraform-provider-elasticstack/pull/1256

internal/kibana/maintenance_window/schema.go
adcoelho: Fixed in [199d5fa](https://github.com/elastic/terraform-provider-elasticstack/pull/1224/commits/199d5faf09bff5a26264b3732512c725e876a8fd)

internal/kibana/maintenance_window/schema.go
adcoelho: Fixed by adding validation in [199d5fa](https://github.com/elastic/terraform-provider-elasticstack/pull/1224/commits/199d5faf09bff5a26264b3732512c725e876a8fd)

internal/kibana/maintenance_window/models.go
adcoelho:
```
func (model *MaintenanceWindowScheduleRecurring) toAPIRequest...
func (model *MaintenanceWindowScope) toAPIRequest...
```
Smart, thanks!

Fixed in [d623300](https://github.com/elastic/terraform-provider-elasticstack/pull/1224/commits/d6233000c86f57586515d8d0753ecf913bf8e9d4).

internal/kibana/maintenance_window/models.go
adcoelho: Fixed in [d623300](https://github.com/elastic/terraform-provider-elasticstack/pull/1224/commits/d6233000c86f57586515d8d0753ecf913bf8e9d4).

internal/kibana/maintenance_window/create.go
adcoelho: So this eliminates the need for `func (model *MaintenanceWindowModel) fromAPICreateResponse` and `func (model *MaintenanceWindowModel) fromAPIUpdateResponse`?

adcoelho: Done in [bd62e01](https://github.com/elastic/terraform-provider-elasticstack/pull/1224/commits/bd62e01348d09dd0778b3e8d5c08296d1aa22f0b).

internal/kibana/maintenance_window/schema.go
adcoelho: Nice catch, thank you!

Fixed in [ec95b76](https://github.com/elastic/terraform-provider-elasticstack/pull/1224/commits/ec95b76efd582f5b59c6aff44a32ba05eca96a1f).

examples/resources/elasticstack_kibana_maintenance_window/import.sh
adcoelho: Fixed in [0d590bc](https://github.com/elastic/terraform-provider-elasticstack/pull/1224/commits/0d590bce4dccc978702a39d0f64527523ac26036).

internal/kibana/maintenance_window/models.go
adcoelho: https://github.com/elastic/kibana/issues/233739

internal/kibana/maintenance_window/models.go
adcoelho: https://github.com/elastic/kibana/issues/233739

internal/kibana/maintenance_window/models.go
```diff
@@ -0,0 +1,321 @@
...
+		return nil
+	}
+
+	var diags = diag.Diagnostics{}
+	var response = &ResponseJson{}
+
+	if err := json.Unmarshal(data.Body, response); err != nil {
+		diags.AddError(err.Error(), "cannot unmarshal GetMaintenanceWindowIdResponse")
+		return diags
+	}
```
tobio, line 82: Can we get rid of the intermediate types now we're only ever populating this from the `GetMaintenanceWindowIdResponse`? I'm happy if you still see value in this step.

internal/kibana/maintenance_window/resource.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+var (
+	_ resource.Resource              = &MaintenanceWindowResource{}
+	_ resource.ResourceWithConfigure = &MaintenanceWindowResource{}
```
tobio:
```suggestion
	_ resource.ResourceWithConfigure = &MaintenanceWindowResource{}
	_ resource.ResourceWithImportState = &MaintenanceWindowResource{}
```

internal/kibana/utils/validators.go
tobio, line 1: Idiomatic Go would have this structured something like:

* In an `internal/kibana/validators` package. `utils` is generally a smell, we should be aiming to remove things from the existing `internal/utils` package to more focused groups. 
* Having a file for each validator type, e.g `is_maintenance_window_interval_frequency.go`, `is_maintenance_window_on_week_day.go`

tobio (commented):
LGTM, some small nits.

internal/kibana/utils/validators.go
adcoelho: I updated it. I didn't like the `package validator_utils` I  had, and just `utils` was conflicting with another folder.

This does look better, thanks!

tobio (approved):
LGTM, thanks!

internal/kibana/maintenance_window/schema.go
```diff
@@ -0,0 +1,153 @@
...
+				Attributes: map[string]schema.Attribute{
+					"start": schema.StringAttribute{
+						Description: "The start date and time of the schedule, provided in ISO 8601 format and set to the UTC timezone. For example: `2025-03-12T12:00:00.000Z`.",
+						Required:    true,
+						Validators: []validator.String{
+							validators.StringIsISO8601{},
+						},
+					},
+					"duration": schema.StringAttribute{
+						Description: "The duration of the schedule. It allows values in `<integer><unit>` format. `<unit>` is one of `d`, `h`, `m`, or `s` for hours, minutes, seconds. For example: `1d`, `5h`, `30m`, `5000s`.",
```
jcger:
```suggestion
						Description: "The duration of the schedule. It allows values in `<integer><unit>` format. `<unit>` is one of `d`, `h`, `m`, or `s` for days, hours, minutes, seconds. For example: `1d`, `5h`, `30m`, `5000s`.",
```

internal/kibana/maintenance_window/schema.go
```diff
@@ -0,0 +1,153 @@
...
+						Attributes: map[string]schema.Attribute{
+							"end": schema.StringAttribute{
+								Description: "The end date and time of the schedule, provided in ISO 8601 format and set to the UTC timezone. For example: `2025-03-12T12:00:00.000Z`.",
+								Optional:    true,
+								Validators: []validator.String{
+									validators.StringIsISO8601{},
+								},
+							},
+							"every": schema.StringAttribute{
+								Description: "The duration of the schedule. It allows values in `<integer><unit>` format. `<unit>` is one of `d`, `h`, `m`, or `s` for hours, minutes, seconds. For example: `1d`, `5h`, `30m`, `5000s`.",
```
jcger:
```suggestion
								Description: "The duration of the schedule. It allows values in `<integer><unit>` format. `<unit>` is one of `d`, `h`, `m`, or `s` for days, hours, minutes, seconds. For example: `1d`, `5h`, `30m`, `5000s`.",
```

jcger (approved):

PR #1186, opened 2025-07-06
Add description to kibana role resource

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1172

PR #1235, opened 2025-08-07
fix: ensure description field for elasticstack_kibana_security_role behaves as optional

snobear:
The description field added in https://github.com/elastic/terraform-provider-elasticstack/pull/1186 was intended to be optional, but I found that when it is not set it causes:

```shell
│ Error: Provider produced invalid plan
│ 
│ Provider "registry.terraform.io/elastic/elasticstack" planned an invalid value for
│ elasticstack_kibana_security_role.kibana_guest_user_role.description: planned value cty.StringVal("") for a
│ non-computed attribute.
│ 
```

It was being set unconditionally in `resourceRoleRead`, which caused Terraform to reject the plan when the description was empty because it was trying to set a non-computed attribute to an empty string. This ensures that the description field is only set when it contains actual content, so it can be optional.


tobio (commented):
Thanks for getting involved @snobear.

Two things:
* The current acceptance tests are passing, but don't set the description attribute. It seems like the actual bug here is slightly more subtle (or maybe specific to some Terraform version). It would be great to understand what that is though. Are you able to add an acceptance test which fails without this change, but is fixed by this PR?
* Are you able to add an entry to `CHANGELOG.md` for this PR?

Issue #693, opened 2024-08-01
[Feature] Add Cross Cluster API keys management


wandergeek:
**Describe the resource you would like to have implemented.**
Cross cluster API keys: https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-cross-cluster-api-key.html#security-api-create-cross-cluster-api-key-request. These are not currently supported in the provider. 

**Describe the solution you'd like**
We will need a new resource, similar to `elasticstack_elasticsearch_security_api_key` , but for cross-cluster keys.

The resource should allow a user to create a cross cluster specific key using the intended API, with a spec that should be familiar to users that have created a "regular" API key using the provider

Kagamashi:
Any updates on when can we expect this feature to be live? 
Is there any deadline for milestone 0.12.0?

PR #1252, opened 2025-08-21
Support cross cluster API keys with the existing API key resource. 

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/693

I started off creating an independent resource for cross cluster keys, but there's a lot of duplication and it's not clear what benefit there is in splitting them apart.


nick-benoit:
It seems that updating the key access:
```
  # elasticstack_elasticsearch_security_api_key.cross_cluster_key will be updated in-place
  ~ resource "elasticstack_elasticsearch_security_api_key" "cross_cluster_key" {
      ~ access               = {
          ~ replication = [
              ~ {
                  ~ names = [
                      ~ "archive-*" -> "archive-test-*",
                    ]
                },
            ]
        }
        id                   = "SgYidikGTeuWQ-qWz5pU_w/z2PdC5kBKm2x_cQXbsR7"
        name                 = "My Cross-Cluster API Key"
        # (8 unchanged attributes hidden)
    }

```

The role descriptions are also updated which results in this error:

```
elasticstack_elasticsearch_security_api_key.cross_cluster_key: Modifying... [id=SgYidikGTeuWQ-qWz5pU_w/z2PdC5kBKm2x_cQXbsR7]
╷
│ 
│ When applying changes to elasticstack_elasticsearch_security_api_key.cross_cluster_key, provider "provider[\"registry.terraform.io/elastic/elasticstack\"]" produced an unexpected
│ new value: .role_descriptors: was
│ cty.StringVal("{\"cross_cluster\":{\"cluster\":[\"cross_cluster_replication\"],\"indices\":[{\"names\":[\"archive-*\"],\"privileges\":[\"cross_cluster_replication\",\"cross_cluster_replication_internal\"],\"allow_restricted_indices\":false}]}}"),
│ but now
│ cty.StringVal("{\"cross_cluster\":{\"cluster\":[\"cross_cluster_replication\"],\"indices\":[{\"names\":[\"archive-test-*\"],\"privileges\":[\"cross_cluster_replication\",\"cross_cluster_replication_internal\"],\"allow_restricted_indices\":false}]}}").
│ 
```

I wonder if we should get rid of `UseStateForUnknown` for the `role_descriptors` field? It seems like this field is not updatable in the cross cluster api ([docs](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-security-update-cross-cluster-api-key)). Perhaps this is a good argument for a different resource like you mentioned in the PR description?


>IMPORTANT: If you don't specify role_descriptors in the request, a call to this API might still change the API key's access scope. This change can occur if the owner user's permissions have changed since the API key was created or last modified.

[link](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-security-update-api-key)

It looks like this field is a bit sticky for non-cross cluster api key updates as well.

internal/elasticsearch/security/api_key/acc_test.go
```diff
@@ -507,3 +518,36 @@ resource "elasticstack_elasticsearch_security_api_key" "test" {
...
+	return fmt.Sprintf(`
+provider "elasticstack" {
+  elasticsearch {}
+}
+
+resource "elasticstack_elasticsearch_security_api_key" "test" {
+  name = "%s"
+  type = "cross_cluster"
+
+  access = {
```
nick-benoit, line 532: Does it make sense to have this acceptance test update something in `access` to get some IT style acceptance test coverage of `SetUnknownIfAccessHasChanges`?

nick-benoit (approved):

internal/elasticsearch/security/api_key/acc_test.go
tobio: Yea, it does exactly that.

Issue #1230, opened 2025-08-04
[Bug] incorrectly named resource for elasticsearch_data_stream_lifecycle on registry.terraform.io


roelofvsr:
**Describe the bug**
resource is named on the page:
elasticsearch_data_stream

resource should be:
elasticsearch_data_stream_lifecycle

concerns this page:
https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/elasticsearch_data_stream_lifecycle

**Screenshots**

<img width="938" height="662" alt="Image" src="https://github.com/user-attachments/assets/64eadadb-ea9a-4315-8f55-1d4ea98794e9" />

tobio:
Thanks for reporting this one @roelofvsr, it could have easily lay around like this for ages.

PR #1253, opened 2025-08-21
Fix copy pasta in datastream lifecycle docs

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1230


nick-benoit (approved):
🚀

Issue #1088, opened 2025-04-01
[Bug] `connector_id` is not used when creating an alerting connector


wandergeek:
**Describe the bug**
When creating a `elasticstack_kibana_action_connector` with a configured `connector_id`, the connector is created with a standard machine generated UUID rather than using the provided one.  

**To Reproduce**
1. TF configuration used '...'
2. TF operations to execute to get the error '...' [e.g `terraform plan`,`terraform apply`, `terraform destroy`]
3. See the error in the output '...'

1. TF configuration used 
```
resource "elasticstack_kibana_action_connector" "my-new-connector3" {
  name              = "human-uuid"
  connector_id      = "lugoz-safes-rusin-bubov-fytex-cydeb"
  connector_type_id = ".slack_api"
  secrets = jsonencode({
    token = "dummy value"
  })
}
```

2. `terraform apply`

3. In devtools for target cluster: `GET kbn:/api/actions/connectors`
```
  {
    "id": "aa1b7fc0-0f4e-11f0-91a0-5df3db21f544",
    "name": "human-uuid",
    "config": {},
    "connector_type_id": ".slack_api",
    "is_preconfigured": false,
    "is_deprecated": false,
    "referenced_by_count": 0,
    "is_missing_secrets": false,
    "is_system_action": false
  },
```

Note that the ID is not as configured. 

**Expected behavior**
The `connector_id` should be configured as the ID of the resource on the target cluster.

**Debug output**
An interesting debugging message appears in a `TRACE`:
```
2025-04-01T23:11:39.118Z [WARN]  Provider "provider[\"registry.terraform.io/elastic/elasticstack\"].cluster-mgmt-a" produced an unexpected new value for module.connectors__cluster-mgmt-a.elasticstack_kibana_action_connector.my-new-connector3, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .connector_id: was cty.StringVal("lugoz-safes-rusin-bubov-fytex-cydeb"), but now cty.StringVal("aa1b7fc0-0f4e-11f0-91a0-5df3db21f544")
2025-04-01T23:11:39.118Z [TRACE] terraform.contextPlugins: Schema for provider "registry.terraform.io/elastic/elasticstack" is in the global cache
[...]

PR #1260, opened 2025-08-25
Migrate Kibana connectors to use the bundled openapi generated client

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1088

This PR:
* Removes the dedicated connectors API client. There's nothing technically wrong with this client, but it's generated from an old spec. Rather than updating that client it makes sense to migrate this resource to the bundled client.
* Adds support for the connectors endpoints to the bundled client
* Updates the bundled client generator to generically add `x-omitempty` into the spec for any nullable parameter. We've been explicitly listing anything that 'breaks' up until now, which isn't really scalable or automatable. There's no real case where we don't want this applied, so lets just add it everywhere. 
* Migrates the connectors resource to the bundled Kibana client :) 
* Updates the resource to support passing through an explicit `connector_id` during creation. I'm not a huge fan of having to define a request editor to mutate the request path if there's no ID specified, but the Kibana spec doesn't include an operation for creating a connector without an ID instead it marks the `id` parameter as optional which is entirely valid. Sadly, requests with a trailing slash aren't properly handled by Kibana 8.7 and lower and so we need to trim that for those client version.


tobio:
> Updates the bundled client generator to generically add x-omitempty into the spec for any nullable parameter.

I'm happy to split this out into a prefactor PR if that makes reviewing this easier. There's not a huge amount of code churn related to this change which is why it's included, but it definitely should be it's own PR.

internal/kibana/connector.go
```diff
@@ -18,11 +18,12 @@ var MinVersionSupportingPreconfiguredIDs = version.Must(version.NewVersion("8.8.
...
-			Type:        schema.TypeString,
-			Computed:    true,
-			Optional:    true,
-			ForceNew:    true,
+			Description:  "A UUID v1 or v4 to use instead of a randomly generated ID.",
+			Type:         schema.TypeString,
+			Computed:     true,
+			Optional:     true,
+			ForceNew:     true,
+			ValidateFunc: validation.IsUUID,
```
nick-benoit, line 26: Are we sure `IsUUID` is the right validation here? This seems to have a very particular view of what counts as a UUID. 

```
func ParseUUID(uuid string) ([]byte, error) {
	if len(uuid) != 2 * uuidLen + 4 {
		return nil, fmt.Errorf("uuid string is wrong length")
	}

	if uuid[8] != '-' ||
		uuid[13] != '-' ||
		uuid[18] != '-' ||
		uuid[23] != '-' {
		return nil, fmt.Errorf("uuid is improperly formatted")
	}

	hexStr := uuid[0:8] + uuid[9:13] + uuid[14:18] + uuid[19:23] + uuid[24:36]

	ret, err := hex.DecodeString(hexStr)
	if err != nil {
		return nil, err
	}
	if len(ret) != uuidLen {
		return nil, fmt.Errorf("decoded hex is the wrong length")
	}

	return ret, nil
}
```

Notably it doesn't seem to allow the example given in the issue that inspired this change `"lugoz-safes-rusin-bubov-fytex-cydeb"`. 

I can't seem to find any details in the connector api [docs](https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-actions-connector-id) about requirements (but maybe i'm just not looking in the right place). 

[This](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/cases/docs/openapi/components/parameters/connector_id.yaml) open api doc I found in the Kibana repo seems to have a less opinionated view though

tobio: In the description of this field it mentions that the `connector_id` must be either UUIDv1 or UUIDv4, both of which are expected to match this validation. Manually testing with Kibana seems to back that assertion up, but I can't find anything in the Kibana docs to link to here. 

> Notably it doesn't seem to allow the example given in the issue that inspired this change "lugoz-safes-rusin-bubov-fytex-cydeb".

FWIW this value is rejected by Kibana (at least v9), I expect that ID has been made up for the issue rather than something that's expected to work.

nick-benoit: > In the description of this field it mentions that the connector_id must be either UUIDv1 or UUIDv4

Yeah this is a good point. Presumably this comes from the Kibana open api spec somewhere? 

This sounds good 👍 I just wanted to make sure we weren't adding a more strict validation than we intended.

nick-benoit (approved):
🚀

internal/kibana/connector.go
gigerdo: I think it would also be totally fine to just rely on the validation of the API. So it would construct a request with a wrong UUID, and then Kibana would respond with a validation error that we can present to the user.
(I thas the advantage that we avoid the risk of having a more strict validation than the underlying API and preventing something that should actually work)

PR #1262, opened 2025-08-25
Extend CONTRIBUTING.md

gigerdo:
Add some more detailed instructions on how to write code for the provider. This is more of a starting point which can be extended further. The goal is to have a basic guidance for anyone new to the provider development: Which plugin to use, which resource to use as a baseline example, etc.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,8 +37,12 @@ When creating new resources:
 * `make lint`: Lints and formats the code.
 * `make test`: Run unit tests.
 * `make docs-generate`: Generate documentation.
+* [Running & Debugging the Provider](#running--debugging-the-provider)
+* [Running Acceptance Tests](#running-acceptance-tests)
 
 ### Running & Debugging the Provider
+
+Also see [Terraform docs on debugging](https://developer.hashicorp.com/terraform/plugin/debugging#starting-a-provider-in-debug-mode).
 
 Run the provider in debug mode and reattach the provider in Terraform:
 * Launch `main.go` with the `-debug` flag from your IDE.
@@ -46,7 +50,7 @@ Run the provider in debug mode and reattach the provider in Terraform:
 * Export it in your shell where you run Terraform: `export TF_REATTACH_PROVIDERS='{…}'`.
 * Terraform will now talk to your debug instance, and you can set breakpoints.
 
-#### Running Acceptance Tests
+### Running Acceptance Tests
 
 Acceptance tests spin up Elasticsearch, Kibana, and Fleet with Docker and run tests in a Go container.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,10 +42,13 @@ When creating new resources:
 
 ### Running & Debugging the Provider
 
+You can run the currently checked-out code for local testing and use it with Terraform.
+
 Also see [Terraform docs on debugging](https://developer.hashicorp.com/terraform/plugin/debugging#starting-a-provider-in-debug-mode).
 
 Run the provider in debug mode and reattach the provider in Terraform:
 * Launch `main.go` with the `-debug` flag from your IDE.
+  * Or launch it with `go run main.go -debug` from the command line.
 * After launching, the provider will print an env var. Copy the printed `TF_REATTACH_PROVIDERS='{…}'` value.
 * Export it in your shell where you run Terraform: `export TF_REATTACH_PROVIDERS='{…}'`.
 * Terraform will now talk to your debug instance, and you can set breakpoints.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,14 +1,60 @@
-# Typical development workflow
+# Contributing
 
-Fork the repo, work on an issue
+This guide explains how to set up your environment, make changes, and submit a PR.
 
-## Updating the generated Kibana client.
+## Development Setup
 
-If your work involves the Kibana API, the endpoints may or may not be included in the generated client.
-Check [generated/kbapi](./generated/kbapi/) for more details.
+* Fork and clone the repo.
+* Setup your preferred IDE (Intelliji, VSCode, etc.)
 
-## Acceptance tests
+Requirements:
+* [Terraform](https://www.terraform.io/downloads.html) >= 1.0.0
+* [Go](https://golang.org/doc/install) >= 1.25
+* Docker (for acceptance tests)
 
+## Development Workflow
+
+* Create a new branch for your changes.
+* Make your changes. See [Useful Commands](#useful-commands) and [Debugging](#debugging).
+  * GitHub Copilot can be used to help with these changes (See [Using Copilot](#using-copilot))
+* Validate your changes
+  * Run unit and acceptance tests (See [Running Acceptance Tests](#running-acceptance-tests)).
+  * Run `make fmt`, `make lint`.
+  * All checks also run automatically on every PR.
+* Add a changelog entry in `CHANGELOG.md` under the `Unreleased` section. This will be included in the release notes of the next release.
+* Submit your PR for review.
+
+When creating new resources:
+* Use the [Plugin Framework](https://developer.hashicorp.com/terraform/plugin/framework/getting-started/code-walkthrough) for new resources.
+  * Use an existing resource (e.g. `internal/elasticsearch/security/system_user`) as a template.
+  * Some resources use the deprecated Terraform SDK, so only resources using the new Terraform Framework should be used as reference.
+* Use the generated API clients to interact with the Kibana APIs. (See [Working with Generated API Clients](#working-with-generated-api-clients)
+* Add documentation and examples for the resource. Update the generated docs with `make docs-generate`.
+* Write unit and acceptance tests.
+
+
+## Using Copilot
+
+GitHub Copilot can speed up development, but you’re responsible for correctness:
+* Create an issue describing the desired change and acceptance criteria.
+* Assign the issue to Copilot and iterate with prompts.
+* Review outputs carefully; add tests and adjust as needed.
+* Example issue: https://github.com/elastic/terraform-provider-elasticstack/issues/1219
+
+### Useful Commands
+
+* `make build`: Build the provider.
+* `make install`: Install the provider.
+* `make fmt`: Format the code.
+* `make lint`: Lint the code.
+* `make test`: Run unit tests.
+* `make docs-generate`: Generate documentation.
+
+#### Running Acceptance Tests
+
+Acceptance tests spin up Elasticsearch, Kibana, and Fleet with Docker and run tests in a Go container.
+
+Quick start (default stack version from Makefile):
 ```bash
 make docker-testacc
 ```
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,60 +1,145 @@
...
+* Add a changelog entry in `CHANGELOG.md` under the `Unreleased` section. This will be included in the release notes of the next release.
+* Submit your PR for review.
 
-A way to forward debug logs to a file:
-```bash
-env TF_ACC_LOG_PATH=/tmp/tf.log TF_ACC_LOG=DEBUG TF_LOG=DEBUG make docker-testacc
-```
+When creating new resources:
+* Use the [Plugin Framework](https://developer.hashicorp.com/terraform/plugin/framework/getting-started/code-walkthrough) for new resources.
+  * Use an existing resource (e.g. `internal/elasticsearch/security/system_user`) as a template.
```
gigerdo, line 28: @dimuon @tobio is `internal/elasticsearch/security/system_user` the best possible example to use as a reference for new resources? I took this from the copilot instructions.

tobio: I think it's a good example to point towards. It's relatively simple in terms of both the provider behaviour and the schema but presents a good overview for how things should work. 

I don't feel strongly here though if someone has a better example.

CONTRIBUTING.md
```diff
@@ -1,60 +1,145 @@
...
-```
+* Create a new branch for your changes.
+* Make your changes. See [Useful Commands](#useful-commands) and [Debugging](#debugging).
+  * GitHub Copilot can be used to help with these changes (See [Using Copilot](#using-copilot))
+* Validate your changes
+  * Run unit and acceptance tests (See [Running Acceptance Tests](#running-acceptance-tests)).
+  * Run `make fmt`, `make lint`.
+  * All checks also run automatically on every PR.
+* Add a changelog entry in `CHANGELOG.md` under the `Unreleased` section. This will be included in the release notes of the next release.
+* Submit your PR for review.
```
tobio: Currently changelog entries link back to the PR introducing the change which means we first need to open the PR and then add the changelog entry. That situation has never really made sense, but it does make sense to be consistent for now. 

IMO rather than improving the ordering here, it would likely make more sense to look at automatically adding changelog entries based on the PR title.

CONTRIBUTING.md
```diff
@@ -1,60 +1,145 @@
...
-A way to forward debug logs to a file:
-```bash
-env TF_ACC_LOG_PATH=/tmp/tf.log TF_ACC_LOG=DEBUG TF_LOG=DEBUG make docker-testacc
-```
+When creating new resources:
+* Use the [Plugin Framework](https://developer.hashicorp.com/terraform/plugin/framework/getting-started/code-walkthrough) for new resources.
+  * Use an existing resource (e.g. `internal/elasticsearch/security/system_user`) as a template.
+  * Some resources use the deprecated Terraform SDK, so only resources using the new Terraform Framework should be used as reference.
+* Use the generated API clients to interact with the Kibana APIs. (See [Working with Generated API Clients](#working-with-generated-api-clients)
+* Add documentation and examples for the resource. Update the generated docs with `make docs-generate`.
```
tobio: We also need to add the doc template ([example](https://github.com/elastic/terraform-provider-elasticstack/blob/main/templates/resources/elasticsearch_security_system_user.md.tmpl)) as part of this work. 

The docs generator doesn't (or didn't, I haven't checked recently) populate the `subcategory` field in the frontmatter which we're using to improve the docs organisation.

CONTRIBUTING.md
```diff
@@ -1,60 +1,145 @@
...
 
-Update documentation templates in `./templates` directory and re-generate docs via:
-```bash
-make docs-generate
-```
+GitHub Copilot can speed up development, but you’re responsible for correctness:
+* Create an issue describing the desired change and acceptance criteria.
+* Assign the issue to Copilot and iterate with prompts.
+* Review outputs carefully; add tests and adjust as needed.
+* Example issue: https://github.com/elastic/terraform-provider-elasticstack/issues/1219
```
tobio: So this is kind of correct, but with some limitations. 

* Contributions should already have an open issue, people shouldn't be creating an issue just to get coding agent involved. We also want to create some guidance about the content of that issue, we should aim to have some example issues which have provided good prompts to coding agent. 
* There's restrictions on who can assign an issue to Copilot. We need to make sure folks have a clear path there
* IMO this section is targeted purely at Elasticians, whilst the rest of this doc mostly makes sense for any community members. I think there's enough content here (i.e 'Adding support for Stack features via Copilot') to have a dedicated internal doc.

CONTRIBUTING.md
```diff
@@ -1,60 +1,145 @@
...
 
-## Update `./CHANGELOG.md`
+### Useful Commands
 
-List of previous commits is a good example of what should be included in the changelog.
+* `make build`: Build the provider.
+* `make install`: Install the provider.
+* `make fmt`: Format the code.
+* `make lint`: Lint the code.
```
tobio:
```suggestion
* `make lint`: Lint the code.
```

Lint also formats

CONTRIBUTING.md
```diff
@@ -1,60 +1,145 @@
...
+* Assign the issue to Copilot and iterate with prompts.
+* Review outputs carefully; add tests and adjust as needed.
+* Example issue: https://github.com/elastic/terraform-provider-elasticstack/issues/1219
 
-## Update `./CHANGELOG.md`
+### Useful Commands
 
-List of previous commits is a good example of what should be included in the changelog.
+* `make build`: Build the provider.
+* `make install`: Install the provider.
```
tobio: I don't know why we maintain this make target tbh, I've never used it and would suggest that using it as part of a development workflow is broken. IMO we shouldn't mention this as part of a contributing guide. 

Should we remove the make target? It depends, I guess it makes sense if any users are building the provider from source. Do we need to make that workflow seamless though...

CONTRIBUTING.md
```diff
@@ -1,60 +1,145 @@
...
+#### Running Acceptance Tests
 
-## Pull request
+Acceptance tests spin up Elasticsearch, Kibana, and Fleet with Docker and run tests in a Go container.
 
-Format the code before pushing:
+Quick start (default stack version from Makefile):
 ```bash
-make fmt
+make docker-testacc
```
tobio: This path is can be quite slow, it both spins up the Elastic Stack docker containers (Yay!), but also runs the actual provider tests inside a golang docker container.

Repeated testing (i.e what we'd expect in a dev workflow) is much faster if we re-use a single running Stack environment but also run the tests directly on the dev machine. 

 ```suggestion
make docker-fleet
make testacc
```

File: CONTRIBUTING.md [hunk 8]
```diff
@@ -1,60 +1,145 @@
...
+  ```sh
+  make transform generate
+  ```
+
+The codebase includes a number of deprecated clients which should not be used anymore:
+- `libs/go-kibana-rest`: Fork of an external library, which is not maintained anymore.
+- `generated/alerting`
+- `generated/connectors`
+- `generated/slo`
+- `internal/clients/*`: Manually written clients. These should only be used/extended if it is not possible to use the generated clients.
```
tobio:
```suggestion
```

This isn't really true. Much of this package configures one of the Stack clients, providing endpoints/credentials/etc from the provider configuration to the client. 

There's some code in here which is used as a bridge between the provider resources and the older Kibana/Elasticsearch clients. We don't need to continue with that as much, but there are still cases where it may make sense, for instance in Kibana connectors have a bunch of non-trivial logic managing API defaults which kind of makes sense in a client layer. 

I don't think there's a need for a hard rule on not making additions here.

CONTRIBUTING.md
gigerdo: Yeah then I'll keep it in.

CONTRIBUTING.md
gigerdo: Yes good point.

CONTRIBUTING.md
gigerdo: I'll reorder it for now. Though it would be nice to have this automated so the release job just collects release notes from all PRs included in the release.

CONTRIBUTING.md
gigerdo: I'll remove it from this doc for now.

Should we have a separate doc in the repo for internal docs? The release process is also something that only applies to Elasticians.

CONTRIBUTING.md
gigerdo: Yeah I also never use it. The default workflow should probably be to use the debugging env-var. I'll remove it from the guide.

CONTRIBUTING.md
gigerdo: Sounds good, I updated the docs to use this faster workflow.

CONTRIBUTING.md
gigerdo: True, I didn't realize this also contains all the wrapper code for the generated client. I'll remove that entry.

CONTRIBUTING.md
```diff
@@ -1,60 +1,129 @@
...
-```bash
-make docs-generate
-```
+### Useful Commands
 
-## Update `./CHANGELOG.md`
+* `make build`: Build the provider.
+* `make lint`: Lints and formats the code.
+* `make test`: Run unit tests.
+* `make docs-generate`: Generate documentation.
```
dimuon, line 39: It makes sense to run acceptance tests locally too. Unlike EC provider, the stack provider can run acceptance tests against locally running ES and Kibana, e.g. please refer to `make docker-testacc` (it starts the apps in docker)  or `make testacc`. The later one requires already running ES and Kibana (they can run in docker) and env vars pointing to them.

dimuon: Oh, just noticed you mentioned that a few lines down. :).

gigerdo: Yeah I put the more complex commands into a separate subsection. But maybe a reference here could be useful.

CONTRIBUTING.md
```diff
@@ -1,60 +1,129 @@
...
 
-List of previous commits is a good example of what should be included in the changelog.
+### Running & Debugging the Provider
 
+Run the provider in debug mode and reattach the provider in Terraform:
+* Launch `main.go` with the `-debug` flag from your IDE.
+* After launching, the provider will print an env var. Copy the printed `TF_REATTACH_PROVIDERS='{…}'` value.
+* Export it in your shell where you run Terraform: `export TF_REATTACH_PROVIDERS='{…}'`.
+* Terraform will now talk to your debug instance, and you can set breakpoints.
```
dimuon, line 54: Maybe it makes sense to refer also to some relevant TF docs - https://developer.hashicorp.com/terraform/plugin/debugging#starting-a-provider-in-debug-mode.

dimuon (approved):
👍

CONTRIBUTING.md
gigerdo: Good point :+1:

tobio (approved):
LGTM

CONTRIBUTING.md
```diff
@@ -1,60 +1,133 @@
...
-Update documentation templates in `./templates` directory and re-generate docs via:
-```bash
-make docs-generate
-```
-
-## Update `./CHANGELOG.md`
+### Running & Debugging the Provider
 
-List of previous commits is a good example of what should be included in the changelog.
+Also see [Terraform docs on debugging](https://developer.hashicorp.com/terraform/plugin/debugging#starting-a-provider-in-debug-mode).
```
tobio, line 47: Coming back to this. These instructions are good for debugging, but I think it'd be worth also documenting just running the locally checked out code as well. 

```suggestion
Also see [Terraform docs on debugging](https://developer.hashicorp.com/terraform/plugin/debugging#starting-a-provider-in-debug-mode).

#### Running the provider in a CLI

You can run the currently checked out code for local testing.

* `go run main.go -debug`* After launching, the provider will print an env var. Copy the printed `TF_REATTACH_PROVIDERS='{…}'` value.
* Export it in your shell where you run Terraform: `export TF_REATTACH_PROVIDERS='{…}'`.
* Terraform will now talk to your debug instance, and you can set breakpoints.

#### Debugging the provider in an IDE


```

tobio (dismissed):

CONTRIBUTING.md
gigerdo: Yeah that's useful. Added the command as an alternative step to the debugging steps.

Issue #1228, opened 2025-08-04
[ResponseOps][Terraform] Upgrade Connector Resource


adcoelho:
## Summary

Similar to what was done for the Maintenance Window resource, the implementation of the connector resource in the ElasticStack Terraform provider should be migrated from the `sdkv2` to the `plugin framework`.

Additionally, the client used should be the `kibana_oapi` generated from the OpenApi specification.

Issue #1214, opened 2025-07-22
Migrate stack provider to newer TF framework


Kushmaro:
Just like what we did with the Elastic Cloud TF provider, we should plan this migration.

The [TF framework](https://developer.hashicorp.com/terraform/plugin/framework) will allow us to tackle some of the current limitations like [these](https://github.com/elastic/terraform-provider-elasticstack/issues?q=is%3Aopen+is%3Aissue+label%3Aframework-limitation).

While it is possible to integrate both provider frameworks (the new - Plugin Framework - and the old - Tf SDK), we should eventually clean up the code and finish the migration to the Plugin Framework, making it easier and more straightforward to support this codebase.

PR #1269, opened 2025-09-01
Migrate elasticstack_kibana_action_connector resource to the plugin framework

tobio:
Does what it says on the tin.

Migrates the old suppress diff functionality into a new ConfigValue type which ignores any default values added into the connector response. 

Related to https://github.com/elastic/terraform-provider-elasticstack/issues/1088
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/1214
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1228


internal/kibana/connectors/delete.go
```diff
@@ -0,0 +1,43 @@
...
+	compositeID, diags := state.GetID()
+	response.Diagnostics.Append(diags...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	spaceId := state.SpaceID.ValueString()
+
+	response.Diagnostics.Append(kibana_oapi.DeleteConnector(ctx, oapiClient, compositeID.ResourceId, spaceId)...)
+	if response.Diagnostics.HasError() {
```
nick-benoit: I suppose this is just following the above pattern, but presumably this check is effectively a no-op?

Issue #1265, opened 2025-08-29
Fully cover the Kibana API in the bundled generated client.


tobio:
Currently `transform_schema` filters out an allowed list of API operations, initially we've done this as the broader API spec is invalid, or is impacted by issues with oapi-codegen. 

Adding new endpoints to the allowlist can often be time consuming uncovering spec issues on the fly. 

In order to simplify the developer experience when adding new functionality (either as entire new resources, or widening the coverage on existing resources) we need to remove this allow list and fully cover the Kibana API.

PR #1270, opened 2025-09-02
Don't allowlist API operations within the Kibana API spec

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1265

Are all these generated models correct? No idea, but it's what Kibana describes. Stack developers should be able to fix request/response models at the source (i.e the Kibana repo) when adding new resources.


nick-benoit (dismissed):
🚀

gigerdo (dismissed):

Issue #1085, opened 2025-04-01
[Bug] elastic_kibana_slo.custom_metric should not require "field" when the aggregation type is "doc_count"


anitahuang-accenture:
**Describe the bug**
Custom metric SLO doesn't require the "field" field when creating the SLO manually, so the resource should not require "field" when the aggregation type is "doc_count"

**To Reproduce**
1. Create a custom metric SLO with the doc_count type of aggregation and field = "*" or field = null, adding the "filter" field with the desired filter

**Expected behavior**
SLO created

**Debug output**
2025-04-01T17:21:46.260-0300 [DEBUG] provider.terraform-provider-elasticstack_v0.11.1: {"statusCode":400,"error":"Bad Request","message":"Failed to validate: \n  Excess keys are not allowed:\nbody.indicator.params.good.metrics.field\nbody.indicator.params.total.metrics.field","attributes":{"data":null}}

**Versions (please complete the following information):**
 - OS: [Linux]
 - Terraform Version [e.g. 1.6.6]
 - Provider version [e.g. v0.11.1]
 - Elasticsearch Version [e.g. 8.16.0]

tobio:
Duplicates https://github.com/elastic/terraform-provider-elasticstack/issues/1035

Issue #1035, opened 2025-02-25
[Feature] Support for doc_count aggregation without the need for field


anitahuang-accenture:
Currently, when using the doc_count aggregation in Elastic Stack to create Service Level Objectives (SLOs), there is no need to provide a field because the doc_count aggregation simply counts documents regardless of any specific field.

However, when trying to configure an SLO in Terraform using the terraform-provider-elasticstack, the configuration requires that the field be filled out, which causes a validation error when using the doc_count aggregation, as this field is not needed for this type of aggregation.

Expected Behavior:
I would like the terraform-provider-elasticstack provider to allow the use of the doc_count aggregation without requiring the field parameter. When the doc_count aggregation is selected, the field should be optional since Elastic Stack does not require this parameter for counting documents.

Proposed Solution:
Update the provider to allow the doc_count aggregation to be used without requiring the field parameter. When doc_count is selected as the aggregation type, the field should be optional and not trigger validation errors. This change would align Terraform's behavior with how doc_count works in Kibana Console and the Elastic Stack API.

tobio:
Duplicates https://github.com/elastic/terraform-provider-elasticstack/issues/1085

tobio:
Fixed in https://github.com/elastic/terraform-provider-elasticstack/pull/2394

PR #1277, opened 2025-09-04
Make field optional since if aggregation is doc_count it's should not…

wasfree:
… be set

Fix for https://github.com/elastic/terraform-provider-elasticstack/issues/1085 and https://github.com/elastic/terraform-provider-elasticstack/issues/1035

If aggregation in metric_custom_indicator is set to doc_count it's not allowed to specifie "field"


tobio:
There's a few more changes required here, one of which is potentially large:
* Update where we populate the API model to optionally read the `field` ([here](https://github.com/elastic/terraform-provider-elasticstack/blob/de1943db550371cf014945de19ed359a118751e3/internal/kibana/slo.go#L695) and [here](https://github.com/elastic/terraform-provider-elasticstack/blob/de1943db550371cf014945de19ed359a118751e3/internal/kibana/slo.go#L706)). We'd likely also want to add some validation in so the provider errors if field is set when for `doc_type` aggregations, or unset for other aggregation types. 
* Update the generated SLO client, the spec used to generate this client lists `field` as required, and so it's always being serialised in the API request body leading to 400's. That means updating [`slo-spec.yml`](https://github.com/elastic/terraform-provider-elasticstack/blob/de1943db550371cf014945de19ed359a118751e3/generated/slo-spec.yml) with the [current spec](https://github.com/elastic/kibana/blob/main/x-pack/solutions/observability/plugins/slo/docs/openapi/slo/bundled.yaml), regenerating the client with `make generate-slo-client` and fixing any errors caused by the potential model changes. 
* Adding an acceptance test to `slo_test.go` verifying the actual bug fix.

tobio:
@wasfree I've updated the SLO client as part of some other work, which should simplify the rest of this task if you've still got some time to look at it.

kpatticha:
@tobio I’m also running into this limitation, are there any alternative workarounds until it gets merged?

PR #1293, opened 2025-09-09
Add export saved objects data source

### LINKED ISSUE #688: [Feature] Export saved objects
Import saved objects already exists for the provider however would like to be able to export objects as well.

```
data "elasticstack_kibana_export_saved_objects" "dashboards" {
  space_id = "default" 
  exclude_export_details = true
  include_references_deep = true
  objects = jsonencode({
    {
      "type": "dashboard",
      "id": "be3733a0-9efe-11e7-acb3-3dab96693fab"
    },
    {
     "type": "dashboard",
      "id": "be3733a0-9efe-11e7-acb3-3dab96693fab"
    }     
  })
}
```

Attributes exported : 
- object.0.content
- object.1.content

This resource would be similar to the [http provider](https://registry.terraform.io/providers/hashicorp/http/latest/docs/data-sources/http)

This would store the object in the state file and allow a user to use the [local file provider](https://registry.terraform.io/providers/hashicorp/local/latest/docs/resources/file) to save to disk or reference it from another [terraform state](https://developer.hashicorp.com/terraform/language/state/remote-state-data)

Strictly just using the http data source is an option but it would be nice to have it in the same provider

**Additional context**
The use case I see this being used most often is when transferring saved objects between clusters that need to be fairly similar. For example, a development cluster where users have privileged access (creating dashboards and visualizations) can then export those and then get imported into a production cluster where those same users only have read access.

tehbooom:
Exports saved objects with the following

```tf
data "elasticstack_kibana_export_saved_objects" "example" {
  space_id                = "default"
  exclude_export_details  = true
  include_references_deep = true
  objects = jsonencode([
    {
      "type" : "dashboard",
      "id" : "7c5f07ee-7e41-4d50-ae1f-dfe54cc87209"
    }
  ])
}

output "saved_objects" {
  value = data.elasticstack_kibana_export_saved_objects.example.exported_objects
}
```

Tested on version 9.1.0

This PR will close #688


tobio (commented):
Thanks for taking the time to add this one in! We'd want this implemented with the `terraform-plugin-framework` rather than the old SDKv2, similarly to how `import_saved_objects` is implemented.

tehbooom:
@tobio running `make lint` and `make docs-generate` shows no change for me and is passing locally.

I'm not sure why CI is failing

internal/kibana/export_saved_objects/schema.go
```diff
@@ -0,0 +1,52 @@
...
+				Computed:    true,
+			},
+			"space_id": schema.StringAttribute{
+				Description: "An identifier for the space. If space_id is not provided, the default space is used.",
+				Optional:    true,
+			},
+			"objects": schema.StringAttribute{
+				Description: "JSON-encoded list of objects to export. Each object should have 'type' and 'id' fields.",
+				Required:    true,
+			},
```
tobio:
```suggestion
			"objects": schema.ListNestedAttribute{
				Description: "List of objects to export.",
				Required:    true,
				Validators: []validator.List{
					listvalidator.SizeAtLeast(1),
				},
				NestedObject: schema.NestedAttributeObject{
					Attributes: map[string]schema.Attribute{
						"type": schema.StringAttribute{
							Description: "The type of the saved object.",
							Required:    true,
						},
						"id": schema.StringAttribute{
							Description: "The ID of the saved object.",
							Required:    true,
						},
					},
				},
			},
```

We can make this much simpler for users by fully defining the input schema here.

internal/kibana/export_saved_objects/read.go
```diff
@@ -0,0 +1,117 @@
...
+			return
+		}
+		objectsList = append(objectsList, struct {
+			Id   string `json:"id"`
+			Type string `json:"type"`
+		}{
+			Id:   id,
+			Type: objType,
+		})
+	}
```
tobio: Assuming the schema change above

```suggestion
	objectsList := utils.ListTypeToSlice(ctx, config.Objects, path.Root("objects"), &resp.Diagnostics, func(item objectModel, meta utils.ListMeta) struct {
		Id   string `json:"id"`
		Type string `json:"type"`
	} {
		return struct {
			Id   string `json:"id"`
			Type string `json:"type"`
		}{
			Id:   item.ID.ValueString(),
			Type: item.Type.ValueString(),
		}
	})
```

internal/kibana/export_saved_objects/schema.go
```diff
@@ -0,0 +1,52 @@
...
+
+// dataSourceModel maps the data source schema data.
+type dataSourceModel struct {
+	ID                    types.String `tfsdk:"id"`
+	SpaceID               types.String `tfsdk:"space_id"`
+	Objects               types.String `tfsdk:"objects"`
+	ExcludeExportDetails  types.Bool   `tfsdk:"exclude_export_details"`
+	IncludeReferencesDeep types.Bool   `tfsdk:"include_references_deep"`
+	ExportedObjects       types.String `tfsdk:"exported_objects"`
+}
```
tobio, line 74: Assuming the schema change above

```suggestion
}

type objectModel struct {
	Type types.String `tfsdk:"type"`
	ID   types.String `tfsdk:"id"`
}
```

internal/kibana/export_saved_objects/acc_test.go
```diff
@@ -0,0 +1,54 @@
...
+data "elasticstack_kibana_export_saved_objects" "test" {
+  space_id = "default"
+  exclude_export_details = true
+  include_references_deep = true
+  objects = jsonencode([
+    {
+      "type": "action",
+      "id": elasticstack_kibana_action_connector.test.connector_id
+    }
+  ])
```
tobio:
```suggestion
  objects = [
    {
      type = "action",
      id = elasticstack_kibana_action_connector.test.connector_id
    }
  ]
```

tobio (commented):
Couple of comments, but they all stem back to using a richer input schema. You'll need to regenerate the docs again after the schema change too :) 

Otherwise, LGTM

internal/kibana/export_saved_objects/schema.go
```diff
@@ -0,0 +1,52 @@
...
+			},
+		},
+	}
+}
+
+// dataSourceModel maps the data source schema data.
+type dataSourceModel struct {
+	ID                    types.String `tfsdk:"id"`
+	SpaceID               types.String `tfsdk:"space_id"`
+	Objects               types.String `tfsdk:"objects"`
```
tobio:
```suggestion
	Objects               types.List `tfsdk:"objects"`
```

tobio (approved):
:tada: thanks for adding this

PR #1296, opened 2025-09-10
Install prebuilt rules

### LINKED ISSUE #750: [Feature] Add support for prebuilt rules
I would like to manage all rules using terraform to include prebuilt ones from Elastic.

Add another resource like `elasticstack_kibana_prebuilt_rule` that installs or updates all prebuilt rules from elastic. Optionally enable or disable rules based on `alert.attribute.tags` using  `POST api/detection_engine/rules/_bulk_action`

Enable rules by `alert.attribute.tags`
```hcl
resource "elasticstack_kibana_prebuilt_rule" "example" {
  tags = [
    "OS: Linux",
    "OS: Windows",
    "Data Source: GCP"
  ]
}
```

This would enable all rules
```hcl
resource "elasticstack_kibana_prebuilt_rule" "example" {
  tags = ["all"]
}
```

This would install the rules but not enable them
```hcl
resource "elasticstack_kibana_prebuilt_rule" "example" {
  tags = []
}
```

Only run if `tags` is updated or if `GET api/detection_engine/rules/prepackaged/_status` returns  `rules_not_installed` or `rules_not_updated` is greater than or equal to 1.

Using `ansible.builtin.uri` or curl as an alternative

**Additional context**

tehbooom:
This PR adds a resource to install and enable prebuilt rules. It will also update the prebuilt rules if it detects there is an update from Elastic.

This will install and enabled rules with the following tags
```tf
resource "elasticstack_kibana_prebuilt_rule" "example" {
  tags = [
    "OS: Linux",
    "OS: Windows",
    "Data Source: GCP"
  ]
}
```

This will only install rules and not enable any of them
```tf
resource "elasticstack_kibana_prebuilt_rule" "example" {
  tags = []
}
```

The initial issue wanted to support enabling all rules at once but this is not feasible and highly unlikely that a user will want to enable 1500 rules. Thus, one can only enable 1000 rules at a time.

Set the minimum version to 8.0.0 as I don't think this was supported in version 7.X and the EOL for 7.17.x is coming soon. 

This PR will close #750

### DIFF
--- a/internal/kibana/prebuilt_rules/acc_test.go
+++ b/internal/kibana/prebuilt_rules/acc_test.go
@@ -91,10 +91,10 @@ func testAccResourcePrebuiltRules(t *testing.T, spaceID string) {
 }
 
 func deleteSingleDetectionRule(t *testing.T, spaceID string) {
-	supported, err := versionutils.CheckIfVersionIsUnsupported(minVersionPrebuiltRules)()
+	unsupported, err := versionutils.CheckIfVersionIsUnsupported(minVersionPrebuiltRules)()
 	require.NoError(t, err)
 
-	if !supported {
+	if unsupported {
 		return
 	}
 


internal/kibana/prebuilt_rules/acc_test.go
```diff
@@ -0,0 +1,78 @@
...
+					resource.TestCheckResourceAttr("elasticstack_kibana_prebuilt_rule.test", "tags.#", "2"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_prebuilt_rule.test", "tags.0", "OS: Linux"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_prebuilt_rule.test", "tags.1", "OS: Windows"),
+				),
+			},
+		},
+	})
+}
+
+func testAccPrebuiltRuleDestroy(s *terraform.State) error {
```
tobio: If this is a no-op we can just omit it entirely from the test definition.

internal/kibana/prebuilt_rules/models.go
```diff
@@ -0,0 +1,245 @@
...
+type prebuiltRulesStatus struct {
+	RulesInstalled        int `json:"rules_installed"`
+	RulesNotInstalled     int `json:"rules_not_installed"`
+	RulesNotUpdated       int `json:"rules_not_updated"`
+	TimelinesInstalled    int `json:"timelines_installed"`
+	TimelinesNotInstalled int `json:"timelines_not_installed"`
+	TimelinesNotUpdated   int `json:"timelines_not_updated"`
+}
+
+func (model *prebuiltRuleModel) populateFromStatus(ctx context.Context, status *prebuiltRulesStatus) diag.Diagnostics {
```
tobio:
```suggestion
func (model *prebuiltRuleModel) populateFromStatus(ctx context.Context, status *prebuiltRulesStatus) {
```

This function never returns a non-nil diagnostic

internal/kibana/prebuilt_rules/models.go
```diff
@@ -0,0 +1,245 @@
...
+	var err error
+
+	if spaceID != "default" {
+		resp, err = client.API.ReadPrebuiltRulesAndTimelinesStatusWithResponse(ctx, func(ctx context.Context, req *http.Request) error {
+			req.Header.Set("kbn-space-id", spaceID)
+			return nil
+		})
+	} else {
+		resp, err = client.API.ReadPrebuiltRulesAndTimelinesStatusWithResponse(ctx)
+	}
```
tobio:
```suggestion
	resp, err = client.API.ReadPrebuiltRulesAndTimelinesStatusWithResponse(ctx, func(ctx context.Context, req *http.Request) error {
		if spaceID != "default" {
			req.Header.Set("kbn-space-id", spaceID)
			return nil
		}
	})
```

internal/kibana/prebuilt_rules/schema.go
```diff
@@ -0,0 +1,70 @@
...
+				Description: "An identifier for the space. If space_id is not provided, the default space is used.",
+				Optional:    true,
+				Computed:    true,
+				Default:     stringdefault.StaticString("default"),
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"tags": schema.ListAttribute{
+				Description: "A list of tag names to filter prebuilt rules for enabling/disabling. Use ['all'] to enable all prebuilt rules, or an empty list to install but not enable any rules.",
```
tobio:
```suggestion
				Description: "A list of tag names to filter prebuilt rules for enabling/disabling.",
```

We're explicitly returning an error for `all`, it doesn't seem helpful to suggest it in the docs. 
Is there a reason we expect this to be set to an empty list to install but not enable, the attribute is marked as optional so I'd expect that just omitting it entirely should be fine?

internal/kibana/prebuilt_rules/schema.go
```diff
@@ -0,0 +1,70 @@
...
+				ElementType: types.StringType,
+				Optional:    true,
+				Validators: []validator.List{
+					listvalidator.SizeAtLeast(0),
```
tobio:
```suggestion
					listvalidator.SizeAtLeast(1),
```

Linked to the above comment, I'd expect that if a user just wants to install the rules then they can leave `tags` undefined, and if they want to enable rules then they should include some tags. I don't really feel strongly about enforcing a non-empty list, but if we're going to maintain support for an empty list of tags then we can just remove this validator altogether, it's impossible to have a list with a negative size.

internal/kibana/prebuilt_rules/update.go
```diff
@@ -0,0 +1,81 @@
...
+
+	spaceID := model.SpaceID.ValueString()
+
+	// Check if we need to install/update rules
+	if needsRuleUpdate(ctx, client, spaceID) {
+		resp.Diagnostics.Append(installPrebuiltRules(ctx, client, spaceID)...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+	}
```
tobio: I'm curious if there's any real benefit to checking if we need an update vs just always running `installPrebuiltRules` regardless?

internal/kibana/prebuilt_rules/models.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+type prebuiltRulesStatus struct {
+	RulesInstalled        int `json:"rules_installed"`
+	RulesNotInstalled     int `json:"rules_not_installed"`
+	RulesNotUpdated       int `json:"rules_not_updated"`
+	TimelinesInstalled    int `json:"timelines_installed"`
+	TimelinesNotInstalled int `json:"timelines_not_installed"`
+	TimelinesNotUpdated   int `json:"timelines_not_updated"`
+}
```
tobio: I don't think we need this model and can just use the generated model directly. 
 
```suggestion
```

internal/kibana/prebuilt_rules/models.go
```diff
@@ -0,0 +1,245 @@
...
+
+	model.RulesInstalled = types.Int64Value(int64(status.RulesInstalled))
+	model.RulesNotInstalled = types.Int64Value(int64(status.RulesNotInstalled))
+	model.RulesNotUpdated = types.Int64Value(int64(status.RulesNotUpdated))
+	model.TimelinesInstalled = types.Int64Value(int64(status.TimelinesInstalled))
+	model.TimelinesNotInstalled = types.Int64Value(int64(status.TimelinesNotInstalled))
+	model.TimelinesNotUpdated = types.Int64Value(int64(status.TimelinesNotUpdated))
+
+	return nil
+}
```
tobio: You'll need to update `getPrebuiltRulesStatus` to return the full generated response, but we shouldn't need to duplicate the response parsing. 

```suggestion
func (model *prebuiltRuleModel) populateFromStatus(ctx context.Context, status *kbapi.ReadPrebuiltRulesAndTimelinesStatusResponse) {
	if status == nil || status.{
		return
	}
	
	model.RulesInstalled = types.Int64Value(int64(status.JSON200.RulesInstalled))
	model.RulesNotInstalled = types.Int64Value(int64(status.JSON200.RulesNotInstalled))
	model.RulesNotUpdated = types.Int64Value(int64(status.JSON200.RulesNotUpdated))
	model.TimelinesInstalled = types.Int64Value(int64(status.JSON200.TimelinesInstalled))
	model.TimelinesNotInstalled = types.Int64Value(int64(status.JSON200.TimelinesNotInstalled))
	model.TimelinesNotUpdated = types.Int64Value(int64(status.JSON200.TimelinesNotUpdated))
}
```

internal/kibana/prebuilt_rules/models.go
```diff
@@ -0,0 +1,245 @@
...
+
+func (model *prebuiltRuleModel) getTags(ctx context.Context) ([]string, diag.Diagnostics) {
+	if model.Tags.IsNull() || model.Tags.IsUnknown() {
```
tobio:
```suggestion
	if !utils.IsKnown(model.Tags) {
```

internal/kibana/prebuilt_rules/models.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	return &status, nil
+}
+
+func installPrebuiltRules(ctx context.Context, client *kibana_oapi.Client, spaceID string) diag.Diagnostics {
+	var resp *kbapi.InstallPrebuiltRulesAndTimelinesResponse
+	var err error
+
+	if spaceID != "default" {
```
tobio: Same as above, can we move the if inside the request editor.

internal/kibana/prebuilt_rules/models.go
```diff
@@ -0,0 +1,245 @@
...
+	} else {
+		resp, err = client.API.InstallPrebuiltRulesAndTimelinesWithResponse(ctx)
+	}
+
+	if err != nil {
+		return utils.FrameworkDiagFromError(err)
+	}
+
+	if resp.StatusCode() != 200 {
+		return utils.FrameworkDiagFromError(fmt.Errorf("failed to install prebuilt rules: %s", resp.Status()))
```
tobio:
```suggestion
		return utils.FrameworkDiagFromError(fmt.Errorf("failed to install prebuilt rules: %s - %s", resp.Status(), string(resp.Body)))
```

internal/kibana/prebuilt_rules/models.go
```diff
@@ -0,0 +1,245 @@
...
+		return nil
+	}
+
+	// Enable rules matching the specified tags
+	diags := performBulkActionByTags(ctx, client, spaceID, "enable", tags)
+	if diags.HasError() {
+		return diags
+	}
+
+	return nil
```
tobio:
```suggestion
	return performBulkActionByTags(ctx, client, spaceID, "enable", tags)
```

In theory there could be warning diags that are getting dropped here.

internal/kibana/prebuilt_rules/models.go
```diff
@@ -0,0 +1,245 @@
...
+	// Disable rules for removed tags
+	if len(removedTags) > 0 {
+		diags := performBulkActionByTags(ctx, client, spaceID, "disable", removedTags)
+		if diags.HasError() {
+			return diags
+		}
+	}
+
+	// Enable rules for added tags
+	if len(addedTags) > 0 {
```
tobio: We only enable newly added tags here, is it possible this misses potential new rules:

* User create this resource with `os:Windows` 
* New rules are added matching `os:Windows` (Stack release? do we push rulesets outside of Stack releases IDK?) 
* User re-runs this resource with the same tag set. The new rules are presumably installed, but they'll never be updated because no tags were added.

internal/kibana/prebuilt_rules/models.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	bodyJSON, err := json.Marshal(bulkActionBody)
+	if err != nil {
+		return utils.FrameworkDiagFromError(err)
+	}
+
+	var resp *kbapi.PerformRulesBulkActionResponse
+
+	if spaceID != "default" {
```
tobio: Same, can we move the if inside the request editor

examples/resources/elasticstack_kibana_prebuilt_rule/import.sh
```diff
@@ -0,0 +1 @@
+terraform import elasticstack_kibana_prebuilt_rule.enable "default"
```
tobio: The resource doesn't implement an import handler, so this command will fail.

IMO I'm not sure import actually makes sense for this resource though, it's quite difficult for us to know that all rules matching a given tag set have been enabled. I don't think there's a real impact to just running create again if the rules have been installed and the expected rules are already enabled?

internal/kibana/prebuilt_rules/models.go
```diff
@@ -0,0 +1,245 @@
...
+		diags := performBulkActionByTags(ctx, client, spaceID, "enable", addedTags)
+		if diags.HasError() {
+			return diags
+		}
+	}
+
+	return nil
+}
+
+func performBulkActionByTags(ctx context.Context, client *kibana_oapi.Client, spaceID, action string, tags []string) diag.Diagnostics {
```
tobio: I'm certainly unfamiliar with this API, so apologies if I'm missing something obvious. What prevents this resource from operating on user defined rules that also match the given tag set? That is, what restricts the resource to only enabling 'pre-built' rules, and not user defined rules which also match a given tag?

tobio (commented):
`tags = ["OS: Linux", "OS: Windows"]`

I'm not sure how I feel about this resource as it's written. 

We're kind of implementing two things (install, and enable) within a single resource. That's not necessarily bad, but I wonder if it would be simpler to have two independent resources in this case. 

There's also a reasonable amount of complexity in managing the tag set, and I'm not sure there's a tangible benefit for users in supporting multiple tags within a single resource. If we reduced this to a single tag then Terraform would be able to fully manage the state transitions for us. 

WDYT about something like:

```hcl
resource "elasticstack_kibana_install_prebuilt_rules" "rules" {}

resource "elasticstack_kibana_enable_prebuilt_rules" "linux" {
  key = "OS"
  value = "Linux"
}

resource "elasticstack_kibana_enable_prebuilt_rules" "windows" {
  key = "OS"
  value = "Windows"
}
```

Issue #395, opened 2023-08-08
Expand elasticstack_fleet_output resource to support Kafka


taylor-swanson:
The output API was recently updated (8.10) to support Kafka as an output type. Currently, the fleet output resource only supports `elasticsearch` and `logstash` as output types. The resource should be updated to include support for Kafka.

The OpenAPI spec around outputs has also changed significantly to provide a more precise schema for the different output types. Unfortunately, this is not a drop-in replacement for the existing output resource and may require significant changes/rework to support the new code generated by the OpenAPI code generator.

DaneViking7:
Is there any timeline on when this may be implemented? Thanks!

SimonKoetting:
Any news on this Topic? Really need this feature

MikePaquette:
@Kushmaro any update on this since adding to milestone 0.12.0?

Kushmaro:
Heya, @MikePaquette not as of yet. The hosted platform team is delayed in working on the providers, unfortunately.

PR #1302, opened 2025-09-14
Add support for Kafka Fleet output types.

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/395

This includes a breaking change to the current provider schema. Existing outputs defining SSL attributes will require an update as part of upgrading. SSL has been moved from a block to an attribute, so `ssl {}` becomes `ssl = {}`. 

For example: 

```
resource "elasticstack_fleet_output" "test_output" {
  name = "Test Output"
  type = "elasticsearch"
  config_yaml = yamlencode({
    "ssl.verification_mode" : "none"
  })
  default_integrations = false
  default_monitoring   = false
  hosts = [
    "https://elasticsearch:9200"
  ] 
  ssl {
    certificate_authorities = [
      file("${path.module}/ca.crt")
    ]
    certificate = file("${path.module}/client.crt")
    key         = file("${path.module}/client.key")
  } 
}
```

becomes:

```
resource "elasticstack_fleet_output" "test_output" {
  name = "Test Output"
  type = "elasticsearch"
  config_yaml = yamlencode({
    "ssl.verification_mode" : "none"
  })
  default_integrations = false
  default_monitoring   = false
  hosts = [
    "https://elasticsearch:9200"
  ] 
  ssl = {
    certificate_authorities = [
      file("${path.module}/ca.crt")
    ]
    certificate = file("${path.module}/client.crt")
    key         = file("${path.module}/client.key")
  } 
}
```


internal/fleet/output/models.go
```diff
@@ -161,22 +366,227 @@ func (model outputModel) toAPICreateModel(ctx context.Context) (union kbapi.NewO
...
+			Random *bool   `json:"random,omitempty"`
+		} {
+			if utils.IsKnown(kafkaModel.Hash) {
+				hashModels := utils.ListTypeAs[outputHashModel](ctx, kafkaModel.Hash, path.Root("kafka").AtName("hash"), &diags)
+				if len(hashModels) > 0 {
+					return &struct {
+						Hash   *string `json:"hash,omitempty"`
+						Random *bool   `json:"random,omitempty"`
+					}{
+						Hash:   hashModels[0].Hash.ValueStringPointer(),
```
nick-benoit: I'm not sure I follow why we only consider `hashModels[0]`. Don't we need to consider all entries? https://www.elastic.co/docs/reference/fleet/kafka-output

nick-benoit: Same question applies to the other complex type handling in this function and the similar update function.

nick-benoit: I see we have this validation [here](https://github.com/elastic/terraform-provider-elasticstack/pull/1302/files). I don't see that in the [docs](https://www.elastic.co/docs/solutions/observability/apm/configure-kafka-output#_partition) or the open api spec.

internal/utils/validators/conditional.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	if !isAllowed {
+		if v.failureMessage != "" {
+			response.Diagnostics.AddAttributeError(
+				request.Path,
+				"Invalid Configuration",
+				v.failureMessage,
+			)
+		} else {
```
nick-benoit, line 74: Is there a reason these validations are exclusive? Would you ever want the diagnostic to include both of the following?

```
"Invalid Configuration: <message>"
"Invalid Configuration: Attribute %s can only be set when %s equals %q, but %s is %q"
```

nick-benoit: Nit: Also is it possible to reduce duplication in this file by creating a shared function that accepts a `*conditionalRequirement`?

nick-benoit (commented):
Just a few clarifying questions. Looks good! 🚀

internal/utils/validators/conditional.go
tobio: I've removed the `failureMessage` part here. We're just duplicating what the generated message would be anyway.

internal/fleet/output/models.go
tobio: I think this was me badly driving an AI agent, but I should have picked these up in review. There's no reason for these to be lists (the schema was restricting them to max length or 1). I've switched them to a single object and tidied this up.

nick-benoit (dismissed):

Issue #903, opened 2024-11-06
Fleet package_policy API use deprecated property policy_id


nchaulet:
## Description 

Related PR for fleet https://github.com/elastic/kibana/pull/184636

Fleet has introduced a change that allow a package policy to belong to multiple agent policy, for this the property `policy_id` has been deprecated in favor of the new `policy_ids` one in both request and response of the fleet `package_policies` API.

This should be changed before the property is removed.

PR #1311, opened 2025-09-16
update integration_policy to include agent_policy_ids

### LINKED ISSUE #1310: [Feature] policy_ids in elasticstack_fleet_integration_policy
The Fleet API was updated to allow support for an integration policy being attached to multiple agent policies. 

We should port this to terraform.

orangecola:
Fixes #1310  
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/903

This is my first PR, so please let me know if there’s anything I can improve.
* policy_id moved to optional, deprecated
* added policy_ids field
* Added unit tests
* Updated kibana API

Some hacking has happened on populateFromAPI and toAPIModel as the kibana API behaves a bit differently with its relation to policy_id and policy_ids.

* If policy_id is not set, and policy_ids is, it will set policy_id to the first value within policy_id.


tobio:
@orangecola thanks for the PR. It looks like the email address used to author the commits isn't attached to your GitHub profile which is causing the CLA check to fail. 

You could either add that email to your profile, or re-author those commits with an email already linked to your profile.

internal/fleet/integration_policy/schema.go
```diff
@@ -50,8 +51,14 @@ func getSchemaV1() schema.Schema {
 				Required:    true,
 			},
 			"agent_policy_id": schema.StringAttribute{
-				Description: "ID of the agent policy.",
-				Required:    true,
+				Description:        "ID of the agent policy.",
+				DeprecationMessage: "Use agent_policy_ids instead. This field will be removed in a future version.",
```
tobio:
```suggestion
```

The provider supports a long list of Elastic stack versions. Throwing a deprecation warning out for people using pre 8.15 clusters isn't a great UX.

internal/fleet/integration_policy/schema.go
```diff
@@ -50,8 +51,14 @@ func getSchemaV1() schema.Schema {
...
+				Optional:           true,
```
tobio: Let's enforce that only one of these can be set at the schema level.

```suggestion
				Optional:           true,
				Validators: []validator.String{
					stringvalidator.ConflictsWith(path.Root("agent_policy_ids").Expression()),
				},
```

internal/fleet/integration_policy/schema.go
```diff
@@ -50,8 +51,14 @@ func getSchemaV1() schema.Schema {
...
+			},
+			"agent_policy_ids": schema.ListAttribute{
+				Description: "List of agent policy IDs.",
+				ElementType: types.StringType,
+				Optional:    true,
```
tobio, line 67:
```suggestion
				Optional:    true,
				Validators: []validator.List{
					listvalidator.ConflictsWith(path.Root("agent_policy_ids").Expression()),
					listvalidator.SizeAtLeast(1),
				},
```

internal/fleet/integration_policy/models.go
```diff
@@ -45,7 +50,38 @@ func (model *integrationPolicyModel) populateFromAPI(ctx context.Context, data *
 	model.PolicyID = types.StringValue(data.Id)
 	model.Name = types.StringValue(data.Name)
 	model.Namespace = types.StringPointerValue(data.Namespace)
-	model.AgentPolicyID = types.StringPointerValue(data.PolicyId)
+
+	// Only populate the agent policy field that was originally configured
+	// to avoid Terraform detecting inconsistent state
+	originallyUsedAgentPolicyID := !model.AgentPolicyID.IsNull() && !model.AgentPolicyID.IsUnknown()
+	originallyUsedAgentPolicyIDs := !model.AgentPolicyIDs.IsNull() && !model.AgentPolicyIDs.IsUnknown()
```
tobio:
```suggestion
	originallyUsedAgentPolicyID := utils.IsKnown(model.AgentPolicyID)
	originallyUsedAgentPolicyIDs := utils.IsKnown(model.AgentPolicyIDs)
```

internal/fleet/integration_policy/models.go
```diff
@@ -45,7 +50,38 @@ func (model *integrationPolicyModel) populateFromAPI(ctx context.Context, data *
...
+		if data.PolicyIds != nil && len(*data.PolicyIds) > 1 {
+			// Multiple policy IDs, use agent_policy_ids
+			agentPolicyIDs, d := types.ListValueFrom(ctx, types.StringType, *data.PolicyIds)
+			diags.Append(d...)
+			model.AgentPolicyIDs = agentPolicyIDs
+		} else {
+			// Single policy ID, use agent_policy_id
+			model.AgentPolicyID = types.StringPointerValue(data.PolicyId)
+		}
+	}
```
tobio: We can tidy this if/else block up with the schema validation enforcing that only one of these is set 

```suggestion
	if originallyUsedAgentPolicyID  {
		// Only set agent_policy_id if it was originally used
		model.AgentPolicyID = types.StringPointerValue(data.PolicyId)
	}
	if originallyUsedAgentPolicyIDs {
		// Only set agent_policy_ids if it was originally used
		if data.PolicyIds != nil {
			agentPolicyIDs, d := types.ListValueFrom(ctx, types.StringType, *data.PolicyIds)
			diags.Append(d...)
			model.AgentPolicyIDs = agentPolicyIDs
		} else {
			model.AgentPolicyIDs = types.ListNull(types.StringType)
		}
	} 
```

internal/fleet/integration_policy/models.go
```diff
@@ -94,7 +130,21 @@ func (model integrationPolicyModel) toAPIModel(ctx context.Context, isUpdate boo
 			Version: model.IntegrationVersion.ValueString(),
 		},
 		PolicyId: model.AgentPolicyID.ValueStringPointer(),
-		Vars:     utils.MapRef(utils.NormalizedTypeToMap[any](model.VarsJson, path.Root("vars_json"), &diags)),
+		PolicyIds: func() *[]string {
```
tobio, line 149: Can we add a check prior to building this request body to return an error diagnostic if `agent_policy_ids` is configured and the current stack version doesn't support it.

internal/fleet/integration_policy/models.go
orangecola: When I removed the final else block, it breaks the test when it is performing an import on the policy:

When doing imports, the model is set to null for both items, which would then cause originallyUsed to be false, even though the policy id / ids are declared in the resource. 

```
DEBUG: originallyUsedAgentPolicyID=false, originallyUsedAgentPolicyIDs=false
DEBUG: model.AgentPolicyID.IsNull()=true, model.AgentPolicyID.IsUnknown()=false
DEBUG: model.AgentPolicyIDs.IsNull()=true, model.AgentPolicyIDs.IsUnknown()=false
DEBUG: data.PolicyId=84486fe4-d947-48fb-b65a-ca8b5ab9bc43, data.PolicyIds=nil

    resource_test.go:178: Step 3/3 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.
        
          map[string]string{
        -       "agent_policy_id": "d821dddf-90c9-4c2f-8300-6110b7c26138",
          }
--- FAIL: TestAccResourceIntegrationPolicySecrets (18.02s)
```

https://github.com/orangecola/terraform-provider-elasticstack/actions/runs/17937030274/job/51005029950

agent_policy_id will always appear in the response, even if you only defined agent_policy_ids. It will always be the first entry in the agent_policy_ids array.

Unless, we just leave agent_policy_id as mandatory and remove the if else block. Let me know if you think of another way.

tobio: Makes sense, thanks for clarifying :)

tobio (commented):
LGTM. Are you able to add an entry to `CHANGELOG.md` for this PR?

orangecola:
I couldn't reproduce the validation conflicts, for some reason, so i'll revert back to the LGTM commit.

tobio (approved):
LGTM, thanks for following up. I'll get this merged.

Issue #1232, opened 2025-08-05
[Bug] elasticstack_fleet_integration_policy unable to handle IDs on isSecretRef


kenwaikinwongfwd:
**Describe the bug**
resource "elasticstack_fleet_agent_policy" unable to handle field "ids" on "isSecretRef" for some integration policies that have a list of IDs instead of single ID.
 
Here is my example of SQL input integration policy from API showing secret ref of "ids" instead of "id":
```
{
  "package": {
    "name": "sql",
    "version": "1.1.0"
  },
  "name": "test-name",
  "namespace": "test-namespace",
  "description": "Test SQL integration",
  "policy_ids": [
    "my_agent_policy_id"
  ],
  "vars": {},
  "inputs": {
    "sql-sql/metrics": {
      "enabled": true,
      "streams": {
        "sql.sql": {
          "enabled": true,
          "vars": {
            "hosts": {
              "ids": [
                "Zgr3ZJgBz9OcmZgWR1vo"
              ],
              "isSecretRef": true
            },
            "period": "5m",
            "driver": "mssql",
            "sql_queries": "- query: SHOW GLOBAL STATUS LIKE 'Innodb_system%'\n  response_format: variables\n        \n",
            "merge_results": false,
            "ssl": "#",
            "processors": "#",
            "data_stream.dataset": "sql"
          }
        }
      }
    }
  }
}
```

Error when running `terraform apply`
 ```
╷│ Error: Plugin did not respond
│ 
│ The plugin encountered an error, and failed to respond to the
│ plugin6.(*GRPCProvider).ApplyResourceChange call. The plugin logs may
│ contain more details.
╵

Stack trace from the terraform-provider-elasticstack_v0.11.15 plugin:

panic: interface conversion: interface {} is nil, not string

github.com/elastic/terraform-provider-elasticstack/internal/fleet/integration_policy.HandleReqRespSecrets({0x1624568, 0x4000750a80}, {0x4000774b78, 0x0, 0x0, 0x400043ad88, {0x4000604d98, 0x11}, 0x4000774b90, 0x0, ...}, ...)
	github.com/elastic/terraform-provider-elasticstack/internal/fleet/integration_policy/secrets.go:172 +0x1390
[...]

PR #1316, opened 2025-09-18
Support secret ref arrays in integration policies

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1232

### DIFF
--- a/internal/fleet/integration_policy/resource_test.go
+++ b/internal/fleet/integration_policy/resource_test.go
@@ -20,7 +20,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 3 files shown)


gigerdo:
One thing I noticed with this is that the bug-report also includes a second bug that comes from defining `vars_json = "{}",`

Because this is defined as `nil` in the API, there will always be a diff causing the error

```
│
│ When applying changes to elasticstack_fleet_integration_policy.sample_sql_input, provider "provider[\"registry.terraform.io/elastic/elasticstack\"]" produced an unexpected new value:
│ .input[0].vars_json: inconsistent values for sensitive attribute.
│
```

However this is a separate bug and there is a workaround which is to not define `vars_json` at all in terraform.

nick-benoit (approved):

PR #1320, opened 2025-09-19
fix: nil reference in api.kibana_synthetics.go:418 in case of unmarshable arrays

biscout42:


libs/go-kibana-rest/kbapi/api.kibana_logstash_pipeline.go
```diff
@@ -73,7 +73,7 @@ func newKibanaLogstashPipelineGetFunc(c *resty.Client) KibanaLogstashPipelineGet
 			if resp.StatusCode() == 404 {
 				return nil, nil
 			}
-			return nil, NewAPIError(resp.StatusCode(), resp.Status())
+			return nil, NewAPIError(resp.StatusCode(), "%s", resp.Status())
```
tobio, line 76: This pattern (`NewAPIError(resp.StatusCode(), resp.Status())`) is still present in other parts of the `go-kibana-rest` code, and seems equivalent to what we're changing it to here. 

Am I missing something where the old code isn't working properly?

biscout42: hm... not sure yet. I thought I have updated all the cases where we don't have a **string constant**. 

Could you share which code do you mean?

tobio: I think I must have been looking at something stale.

tobio (dismissed):
LGTM

Can we add a CHANGELOG entry for this one too?

Issue #520, opened 2023-12-30
[Feature] Anomaly Detection Jobs 


nicholas-r-king:
**Is your feature request related to a problem? Please describe.**
We are an enterprise customer (20+ units) transitioning from AWS CloudWatch to Elastic (via ECK). We currently have hundreds of developers and utilize terraform (CDK) to provision and manage our alerting and dashboarding in CloudWatch. We utilize CloudWatch for both its simple functionality (threshold based alarms) and its more advanced (Anomaly Detection). More recently we can now provision simple threshold based alerts via terraform for Elastic and teams are able to do that without issue, but as we've investigated porting over our AD based alarms we keep running into fact that anomaly detection jobs themselves are NOT manageable by the elasticstack terraform provider or any other providers. This means developers cannot provision AD alarms. This is blocking us from completely transitioning from CloudWatch to Elastic. 

**Describe the resource you would like to have implemented.**
A clear and concise description of the API you want to target.
The Anomaly Detection API's. 
https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-ad-apis.html

https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-put-job.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-delete-job.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-close-job.html
https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-get-job.html

**Describe the solution you'd like**
For the official terraform provider for the elasticstack to support the creation, reading, updating and deletion of Anomaly Detection jobs and allow for them to be referenced in alert rules. 

**Describe alternatives you've considered**
We've designed and tested our own bash script type solution (via null provider) that creates and deletes AD jobs. This does not scale.

holy-moly-555:
Would be great to provision ml jobs through the provider!

jakubpetrovic:
+1 on this feature. We heavily utilize elastic for observability and want to start utilizing anomaly detection. Not having native terraform resources for this is very unfortunate for us as we need to manage jobs in many clusters, and link anomaly detection alerts.

sophiec20:
cc @arisonl @dimkots

heespi:
Hi @dimkots ... any idea on whether the team would be in a position to prioritize this work in the medium term?

sophiec20:
[...]

PR #1329, opened 2025-09-22
Add resource to manage ML anomaly detection jobs

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/520

This resource only manages the _existence_ of an anomaly detection job, it does not manage the state (open/closed). I'm planning on adding another resource to purely manage that state, it's an independent API. IMO it's simpler for practitioners if these are separate resources altogether.


internal/elasticsearch/ml/anomaly_detection_job/acc_test.go
```diff
@@ -0,0 +1,214 @@
...
+					resource.TestCheckResourceAttr("elasticstack_elasticsearch_ml_anomaly_detection_job.test", "analysis_config.bucket_span", "15m"),
+					resource.TestCheckResourceAttr("elasticstack_elasticsearch_ml_anomaly_detection_job.test", "analysis_config.detectors.0.function", "count"),
+					resource.TestCheckResourceAttr("elasticstack_elasticsearch_ml_anomaly_detection_job.test", "data_description.time_field", "@timestamp"),
+					resource.TestCheckResourceAttr("elasticstack_elasticsearch_ml_anomaly_detection_job.test", "data_description.time_format", "epoch_ms"),
+					resource.TestCheckResourceAttrSet("elasticstack_elasticsearch_ml_anomaly_detection_job.test", "create_time"),
+					resource.TestCheckResourceAttr("elasticstack_elasticsearch_ml_anomaly_detection_job.test", "job_type", "anomaly_detector"),
+				),
+			},
+			{
+				Config: testAccResourceAnomalyDetectionJobComprehensive(jobID),
```
nick-benoit: Would it be worth separating into a basic create / update and "comprehensive" create / update test so we get coverage for things like updating the minimal example and creating the comprehensive example?

internal/elasticsearch/ml/anomaly_detection_job/create.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// Create the ML job
+	res, err := esClient.ML.PutJob(jobID, bytes.NewReader(body), esClient.ML.PutJob.WithContext(ctx))
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to create ML anomaly detection job", err.Error())
+		return
+	}
+	defer res.Body.Close()
+
+	if diags := diagutil.CheckErrorFromFW(res, fmt.Sprintf("Unable to create ML anomaly detection job: %s", jobID)); diags.HasError() {
```
nick-benoit: Is it worth adding non-error diags coming from `diagutil.CheckErrorFromFW`? Currently only error diags are added, but in the future I suppose warnings could be added.

internal/elasticsearch/ml/anomaly_detection_job/delete.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// Delete the ML job
+	res, err := esClient.ML.DeleteJob(jobID, esClient.ML.DeleteJob.WithContext(ctx))
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to delete ML anomaly detection job", err.Error())
+		return
+	}
+	defer res.Body.Close()
+
+	if diags := diagutil.CheckErrorFromFW(res, fmt.Sprintf("Unable to delete ML anomaly detection job: %s", jobID)); diags.HasError() {
```
nick-benoit: Do we need to handle non-error diag here?

internal/elasticsearch/ml/anomaly_detection_job/read.go
```diff
@@ -0,0 +1,68 @@
...
+		diags.AddError("Failed to get ML anomaly detection job", err.Error())
+		return false, diags
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == http.StatusNotFound {
+		return false, nil
+	}
+
+	if d := diagutil.CheckErrorFromFW(res, fmt.Sprintf("Unable to get ML anomaly detection job: %s", jobID)); d.HasError() {
```
nick-benoit: Do we need to handle non-error diag here? (same as others)

internal/elasticsearch/ml/anomaly_detection_job/read.go
```diff
@@ -0,0 +1,68 @@
...
+		diags.AddError("Failed to decode job response", err.Error())
+		return false, diags
+	}
+
+	if len(response.Jobs) == 0 {
+		return false, nil
+	}
+
+	// Convert API response back to TF model
+	diags.Append(job.fromAPIModel(ctx, &response.Jobs[0])...)
```
nick-benoit, line 71: Would it be worth adding an error diag here if there is more than one job in `response.Jobs`? This would probably only be future proofing against the read API call being updated without the corresponding change here.

internal/elasticsearch/ml/anomaly_detection_job/update.go
```diff
@@ -0,0 +1,175 @@
...
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	jobID := state.JobID.ValueString()
+
+	tflog.Debug(ctx, fmt.Sprintf("Updating ML anomaly detection job: %s", jobID))
+
+	// Note: Many ML job properties cannot be updated after creation.
```
nick-benoit, line 37: Would it make sense to mark all of the non-editable fields as `Computed: true` and `Optional: false` so they can only be set on creation instead of manually diffing the updatable fields?

internal/elasticsearch/ml/anomaly_detection_job/update.go
```diff
@@ -0,0 +1,175 @@
...
+		hasChanges = true
+	}
+
+	if !plan.ResultsRetentionDays.Equal(state.ResultsRetentionDays) && !plan.ResultsRetentionDays.IsNull() {
+		updateBody.ResultsRetentionDays = utils.Pointer(plan.ResultsRetentionDays.ValueInt64())
+		hasChanges = true
+	}
+
+	// Only proceed with update if there are changes
+	if !hasChanges {
```
nick-benoit, line 53: I'm not familiar with this pattern of early exiting from the update. If we ever hit this case won't the resource always report that it requires an update in future plans due to the desired state not matching the resource? I wonder if we should at least report a warning diag or something here.

internal/elasticsearch/ml/anomaly_detection_job/update.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// Update the ML job
+	res, err := esClient.ML.UpdateJob(jobID, bytes.NewReader(body), esClient.ML.UpdateJob.WithContext(ctx))
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to update ML anomaly detection job", err.Error())
+		return
+	}
+	defer res.Body.Close()
+
+	if diags := diagutil.CheckErrorFromFW(res, fmt.Sprintf("Unable to update ML anomaly detection job: %s", jobID)); diags.HasError() {
```
nick-benoit: Do we need to handle no-error diag here?

internal/elasticsearch/ml/anomaly_detection_job/update.go
```diff
@@ -0,0 +1,175 @@
...
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+
+	// Read the updated job to get the current state
+	found, diags := r.read(ctx, &plan)
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
```
nick-benoit: Should we handle non-error diags here?

internal/elasticsearch/ml/anomaly_detection_job/resource.go
```diff
@@ -0,0 +1,75 @@
...
+func (r *anomalyDetectionJobResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	var state AnomalyDetectionJobTFModel
+	diags := req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	found, diags := r.read(ctx, &state)
+	if diags.HasError() {
```
nick-benoit: Do we need to append non-error diags here?

internal/elasticsearch/ml/anomaly_detection_job/schema.go
```diff
@@ -0,0 +1,456 @@
...
+				},
+			},
+			"job_id": schema.StringAttribute{
+				MarkdownDescription: "The identifier for the anomaly detection job. This identifier can contain lowercase alphanumeric characters (a-z and 0-9), hyphens, and underscores. It must start and end with alphanumeric characters.",
+				Required:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+				Validators: []validator.String{
+					stringvalidator.LengthBetween(1, 64),
```
nick-benoit, line 51: Is this really capped at 64 characters? https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ml-put-job#operation-ml-put-job-job_id 

I didn't see this in the docs

internal/elasticsearch/ml/anomaly_detection_job/schema.go
```diff
@@ -0,0 +1,456 @@
...
+			"analysis_config": schema.SingleNestedAttribute{
+				MarkdownDescription: "Specifies how to analyze the data. After you create a job, you cannot change the analysis configuration; all the properties are informational.",
+				Required:            true,
+				PlanModifiers: []planmodifier.Object{
+					objectplanmodifier.RequiresReplace(),
+				},
+				Attributes: map[string]schema.Attribute{
+					"bucket_span": schema.StringAttribute{
+						MarkdownDescription: "The size of the interval that the analysis is aggregated into, typically between 15m and 1h. If the anomaly detector is expecting to see data at near real-time frequency, then the bucket_span should be set to a value around 10 times the time between ingested documents. For example, if data comes every second, bucket_span should be 10s; if data comes every 5 minutes, bucket_span should be 50m. For sparse or batch data, use larger bucket_span values.",
+						Required:            true,
```
nick-benoit: Is this actually required? In the request body docs it makes it seem not required: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ml-put-job#operation-ml-put-job-body-application-json

internal/elasticsearch/ml/anomaly_detection_job/schema.go
```diff
@@ -0,0 +1,456 @@
...
+					},
+					"detectors": schema.ListNestedAttribute{
+						MarkdownDescription: "Detector configuration objects. Detectors identify the anomaly detection functions and the fields on which they operate.",
+						Required:            true,
+						Validators: []validator.List{
+							listvalidator.SizeAtLeast(1),
+						},
+						NestedObject: schema.NestedAttributeObject{
+							Attributes: map[string]schema.Attribute{
+								"function": schema.StringAttribute{
```
nick-benoit, line 97: Is this actually required?

internal/elasticsearch/ml/anomaly_detection_job/schema.go
```diff
@@ -0,0 +1,456 @@
...
+					"time_format": schema.StringAttribute{
+						MarkdownDescription: "The time format, which can be epoch, epoch_ms, or a custom pattern.",
+						Optional:            true,
+					},
+					"format": schema.StringAttribute{
+						MarkdownDescription: "Only JSON format is supported at this time.",
+						Optional:            true,
+						Computed:            true,
+						Validators: []validator.String{
+							// TODO: TB If only json is supported then why are we allowing delimited?
```
nick-benoit: Leftover TODO

internal/elasticsearch/ml/anomaly_detection_job/acc_test.go
tobio: Yep, that's a good idea.

internal/elasticsearch/ml/anomaly_detection_job/schema.go
tobio: Screenshot 2025-10-21 at 14 15 42

Yea, I can't seem to find this in any docs, but the API limits the length.

internal/elasticsearch/ml/anomaly_detection_job/schema.go
tobio: Same thing, this appears to be a docs issue (or maybe older versions of ES allowed jobs without a defined function). There's no way to define a job without setting this parameter, though some of the error messages suggest you can. 

For example, if you define a detector with just `over_field_name` defined, the error is `field_name must be set when the 'metric' function is used`, but then adding `field_name` to that gets `field_name cannot be used with function 'null'`.

internal/elasticsearch/ml/anomaly_detection_job/schema.go
tobio: You're absolutely right, added the 5m default.

internal/elasticsearch/ml/anomaly_detection_job/read.go
tobio: I've added this as a warning. I'm a bit torn on if this should actually prevent users from moving forward. WDYT?

internal/elasticsearch/ml/anomaly_detection_job/schema.go
tobio: I've removed all these options since ES removed support for delimited formats in 8.x

internal/elasticsearch/ml/anomaly_detection_job/update.go
tobio: Doing that would make them read-only, and so not settable during creation. They should all be marked `RequiresReplace`, and so we _should_ be able to assume an updateable field has changed if we get here.

internal/elasticsearch/ml/anomaly_detection_job/update.go
tobio: Kind of the same as https://github.com/elastic/terraform-provider-elasticstack/pull/1329#discussion_r2446901288

I've included a warning message in this case so if practitioners do hit this case hopefully we'll get notified and can investigate.

internal/elasticsearch/ml/anomaly_detection_job/read.go
nick-benoit: Yeah I think that makes sense 👍

internal/elasticsearch/ml/anomaly_detection_job/update.go
nick-benoit: It seems like i'm missing something obvious here, but i'm not quite sure I follow why we have all this diffing logic in this file

eg 
```
	if !plan.RenormalizationWindowDays.Equal(state.RenormalizationWindowDays) && !plan.RenormalizationWindowDays.IsNull() {
		updateBody.RenormalizationWindowDays = utils.Pointer(plan.RenormalizationWindowDays.ValueInt64())
		hasChanges = true
	}
```

If we have `RequiresReplacement`  in the right places shouldn't be able to jut do: `plan.toAPIModel(ctx)`  and know that we are only updating fields that are allowed to be updated?

nick-benoit: Per discussion in meeting, this is because the PUT api should only be sent fields that have actually changed instead of a full update payload.

nick-benoit (dismissed):
🚀

PR #1337, opened 2025-09-28
Add ML job state resource

tobio:
Requires https://github.com/elastic/terraform-provider-elasticstack/pull/1329

This resource is intended to manage the state of an ML anomaly detection job. Specifically, whether the job is opened of closed. The intention here is to ultimately have four resources:

* The anomaly detection job
* The data feed
* The job state (this PR) 
* The data feed state

This is intended to better align with the lifecycle of the underlying stack resources, whilst also simplifying the job and data feed resources.


.env
```diff
@@ -3,7 +3,7 @@ ELASTICSEARCH_CONTAINER_NAME=terraform-elasticstack-es
 ELASTICSEARCH_PASSWORD=password
 ELASTICSEARCH_PORT=9200
 ELASTICSEARCH_URL=http://localhost:${ELASTICSEARCH_PORT}
-ELASTICSEARCH_JAVA_OPTS="-Xms128m -Xmx2g"
+ELASTICSEARCH_JAVA_OPTS="-Xms128m -Xmx1g"
```
nick-benoit, line 6: Did this help with CI or something?

docs/resources/elasticsearch_ml_job_state.md
```diff
@@ -0,0 +1,145 @@
+---
+page_title: "elasticstack_elasticsearch_ml_job_state Resource - terraform-provider-elasticstack"
+subcategory: ""
+description: |-
+  ML Job State Resource
+  Manages the state of an Elasticsearch Machine Learning (ML) job, allowing you to open or close ML jobs.
+  This resource uses the following Elasticsearch APIs:
+  Open ML Job API https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-open-job.htmlClose ML Job API https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-close-job.htmlGet ML Job Stats API https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-get-job-stats.html
+  Important Notes
+  This resource manages the state of an existing ML job, not the job configuration itself.The ML job must already exist before using this resource.Opening a job allows it to receive and process data.Closing a job stops data processing and frees up resources.Jobs can be opened and closed multiple times throughout their lifecycle.
```
nick-benoit, line 10: Nit: For some reason there aren't any spaces after periods in this snippet.

internal/clients/elasticsearch/ml_job.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, diags
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == http.StatusNotFound {
+		return nil, diags
+	}
+
+	if fwDiags := diagutil.CheckErrorFromFW(res, fmt.Sprintf("Unable to get ML job stats: %s", jobId)); fwDiags.HasError() {
+		diags.Append(fwDiags...)
```
nick-benoit: Append diags outside of error check

internal/elasticsearch/ml/job_state/update.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// First, get the current job stats to check if the job exists and its current state
+	currentJob, fwDiags := elasticsearch.GetMLJobStats(ctx, client, jobId)
+	diags.Append(fwDiags...)
+	if diags.HasError() {
+		return diags
+	}
+
+	if currentJob == nil {
+		diags.AddError(
```
nick-benoit, line 67: In read when we encounter `currentJob == nil` we remove the resource from state: https://github.com/elastic/terraform-provider-elasticstack/pull/1337/files#diff-226ab1183831f21c5bcc39d3c0bac967d9487f86061de61a2f585cfcb36d6d58R41

Should we do that here too?

internal/elasticsearch/ml/job_state/update.go
```diff
@@ -0,0 +1,181 @@
...
+	data.Id = types.StringValue(compId.String())
+	data.JobId = types.StringValue(jobId)
+	data.State = types.StringValue(desiredState)
+
+	diags.Append(state.Set(ctx, data)...)
+	return diags
+}
+
+// waitForJobStateTransition waits for an ML job to reach the desired state
+func waitForJobStateTransition(ctx context.Context, client *clients.ApiClient, jobId, desiredState string) error {
```
nick-benoit: Can this be replaced with: https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/asyncutils/state_waiter.go?

.env
tobio: In main, we have a 2G container memory limit and a 2G JRE heap. ML models are stored outside of the heap, and so these values left essentially no space for ML. ES verifies the model fits in the available memory and so would return an error as there wasn't any available space. 

The general recommendation is to give half the available ram to ES heap, so that's what this change does. We're not really looking for ideal performance from this instance through, so I do want to do a little more testing to make sure this doesn't create instability in acceptance test runs in Buildkite. Is 1G enough for all acceptance tests to pass consistently, I don't know but I'd like to have some confidence.

docs/resources/elasticsearch_ml_job_state.md
tobio: I'm not sure where this frontmatter description is used, but the formatting is stripped here in all the docs. In this case, it looks like the result for stripping a markdown list back to plain text.

internal/elasticsearch/ml/job_state/update.go
tobio: No, this should be an error. 

If the job doesn't exist then the user has provided an incorrect ID, or there's some race condition in their TF setup. In either case, this resource can't manage the current state of a non-existent job. 

In the read case, we've previously successfully applied a change to the job state, and so the job exists in Terraform state. The job has then been deleted from Elasticsearch outside of the Terraform flow and so we're just reconciling the Terraform state with reality.

internal/elasticsearch/ml/job_state/update.go
tobio: Yep, definitely. Thanks for the reminder, I'd planned to do this when the ML resource was merged but should have added a comment here so it didn't get missed.

.env
nick-benoit: Oh interesting. I didn't know that. Thanks for sharing.

nick-benoit (approved):

Issue #813, opened 2024-10-01
[Feature] Add ML Datafeed Management


wandergeek:
**Describe the resource you would like to have implemented.**
I'd like to manage datafeeds using the provider: 

*create*: https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-put-datafeed.html. 
*update*: https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-update-datafeed.html
*delete*: https://www.elastic.co/guide/en/elasticsearch/reference/current/ml-delete-datafeed.html

**Describe the solution you'd like**
This will require some orchestration on updates since datafeeds can not be updated if they are running. 

**Describe alternatives you've considered**
N/A

sophiec20:
Related https://github.com/elastic/terraform-provider-elasticstack/issues/520

PR #1340, opened 2025-09-29
Add ML Datafeed resource

tobio:
Requires https://github.com/elastic/terraform-provider-elasticstack/pull/1329
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/813

Adds a resource to manage the existence of an ML Datafeed. I'm planning on adding an additional resource which manages the state of that data feed (started/stopped) which will close out #813


internal/clients/elasticsearch/ml_job.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode == http.StatusNotFound {
+		return nil, diags
+	}
+
+	fwDiags := diagutil.CheckErrorFromFW(res, fmt.Sprintf("Unable to get ML datafeed: %s", datafeedId))
+	if fwDiags.HasError() {
+		diags.Append(fwDiags...)
```
nick-benoit: Should we append diags outside of error check?

internal/elasticsearch/ml/datafeed/models.go
```diff
@@ -0,0 +1,446 @@
...
+func (m *Datafeed) ToAPIModel(ctx context.Context) (*models.Datafeed, fwdiags.Diagnostics) {
+	var diags fwdiags.Diagnostics
+
+	apiModel := &models.Datafeed{
+		DatafeedId: m.DatafeedID.ValueString(),
+		JobId:      m.JobID.ValueString(),
+	}
+
+	// Convert indices
+	if !m.Indices.IsNull() && !m.Indices.IsUnknown() {
```
nick-benoit: Could checks like this just be: `utils.isKnown`?

internal/elasticsearch/ml/datafeed/resource.go
```diff
@@ -0,0 +1,87 @@
...
+	var state Datafeed
+	diags := req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	found, diags := r.read(ctx, &state)
+	if diags.HasError() {
+		resp.Diagnostics.Append(diags...)
```
nick-benoit: Append diags outside of error check

internal/elasticsearch/ml/datafeed/schema.go
```diff
@@ -0,0 +1,266 @@
...
+			"job_id": schema.StringAttribute{
+				MarkdownDescription: "Identifier for the anomaly detection job. The job must exist before creating the datafeed.",
+				Required:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"indices": schema.ListAttribute{
+				MarkdownDescription: "An array of index names. Wildcards are supported. If any of the indices are in remote clusters, the machine learning nodes must have the `remote_cluster_client` role.",
+				Required:            true,
```
nick-benoit, line 67: Is this actually required? https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ml-put-datafeed#operation-ml-put-datafeed-body-application-json-indices

internal/elasticsearch/ml/datafeed/schema.go
```diff
@@ -0,0 +1,266 @@
...
+					listvalidator.SizeAtLeast(1),
+				},
+			},
+			"query": schema.StringAttribute{
+				MarkdownDescription: "The Elasticsearch query domain-specific language (DSL). This value corresponds to the query object in an Elasticsearch search POST body. All the options that are supported by Elasticsearch can be used, as this object is passed verbatim to Elasticsearch. By default uses `{\"match_all\": {\"boost\": 1}}`.",
+				Optional:            true,
+				Computed:            true,
+				CustomType:          jsontypes.NormalizedType{},
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
```
nick-benoit, line 79: Why do we want this here? IIUC this would typically be a configured value right so this wouldn't really apply?

internal/elasticsearch/ml/datafeed/schema.go
```diff
@@ -0,0 +1,266 @@
...
+			"scroll_size": schema.Int64Attribute{
+				MarkdownDescription: "The size parameter that is used in Elasticsearch searches when the datafeed does not use aggregations. The maximum value is the value of `index.max_result_window`, which is 10,000 by default.",
+				Optional:            true,
+				Computed:            true,
+				PlanModifiers: []planmodifier.Int64{
+					int64planmodifier.UseStateForUnknown(),
+				},
+				Validators: []validator.Int64{
+					int64validator.AtLeast(1),
+					int64validator.AtMost(10000),
```
nick-benoit: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ml-put-datafeed#operation-ml-put-datafeed-body-application-json-scroll_size

It looks like this max value can be updated in a given ES instance. Given that I wonder if we should validate the max value?

internal/elasticsearch/ml/datafeed/schema.go
```diff
@@ -0,0 +1,266 @@
...
+						},
+					},
+					"time_span": schema.StringAttribute{
+						MarkdownDescription: "The time span for each chunk. Only applicable and required when mode is `manual`. Must be a valid duration.",
+						Optional:            true,
+						Computed:            true,
+						PlanModifiers: []planmodifier.String{
+							stringplanmodifier.UseStateForUnknown(),
+						},
+						Validators: []validator.String{
```
nick-benoit, line 165: If we were excited to use our fancy new validators we could add a validation for:

> This setting is applicable only when the mode is set to manual.

internal/elasticsearch/ml/datafeed/schema.go
```diff
@@ -0,0 +1,266 @@
...
+			},
+			"delayed_data_check_config": schema.SingleNestedAttribute{
+				MarkdownDescription: "Specifies whether the datafeed checks for missing data and the size of the window. The datafeed can optionally search over indices that have already been read in an effort to determine whether any data has subsequently been added to the index. If missing data is found, it is a good indication that the `query_delay` is set too low and the data is being indexed after the datafeed has passed that moment in time. This check runs only on real-time datafeeds.",
+				Optional:            true,
+				Computed:            true,
+				PlanModifiers: []planmodifier.Object{
+					objectplanmodifier.UseStateForUnknown(),
+				},
+				Attributes: map[string]schema.Attribute{
+					"enabled": schema.BoolAttribute{
```
nick-benoit, line 180: It looks like this is required: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ml-put-datafeed#operation-ml-put-datafeed-body-application-json-delayed_data_check_config-enabled

internal/models/ml.go
```diff
@@ -0,0 +1,131 @@
+package models
+
+import "time"
+
+// Datafeed represents the complete datafeed as returned by the Elasticsearch API
+type Datafeed struct {
+	DatafeedId             string                  `json:"datafeed_id"`
```
nick-benoit, line 7: Why do not all fields have `omitempty` like the Kibana client?

nick-benoit (commented):
Looking good! Just a few questions

internal/models/ml.go
tobio: `omitempty` is only set on optional fields, which should be consistent with the Kibana client (it doesn't set `omitempty` on fields marked as required in the OpenAPI spec).

internal/elasticsearch/ml/datafeed/schema.go
tobio: `query` is optional, but with a server applied default. So if a user leaves it unset, ES will return `{"match_all": {"boost": 1}}` as the query.

Without `UseStateForUnknown`, when query is unset it will be listed in a `terraform plan` as "known after apply" which creates noise for users, which we can avoid.

internal/elasticsearch/ml/datafeed/schema.go
tobio: Screenshot 2025-10-22 at 22 04 13

Yep, there's an API error if this is left unspecified.

PR #1365, opened 2025-10-14
Fixup docs frontmatter

tobio:
The subcategory generation was leaving leading whitespace which broke the frontmatter of these generated docs. This leads to all resources/datasources being lumped together, and generally things just not being how we want.


KastenMike:
@tobio am I missing something or this does not support the hidden/system indices option that the UI has? If not, was it intentional?

image

PR #1370, opened 2025-10-15
Synthetics tidy

tobio:
Closes https://github.com/elastic/terraform-provider-elasticstack/issues/610

This issue is essentially already done, I've moved some code around to better align with our normal organisation.


biscout42 (dismissed):
Thanks a lot for the clean-up. Good catch that Geo data is only private location related, using of existing `utils.MapValueFrom` and moving api client to a separate file. 

I wish we could migrate kibana client to OpenApi / generated one. 

One nit question: what is the convention for acceptance test file names? We have  `internal/kibana/synthetics/parameter/resource_test.go`  and e.g. `internal/kibana/synthetics/monitor/acc_test.go` .

Issue #1309, opened 2025-09-16
Define a Buildkite pipeline to update the Kibana OpenAPI spec


tobio:
We want to be able to validate the Kibana OpenAPI spec is valid within the Kibana PR checks. This Buildkite pipeline should:

* Accept a target ref to pull the spec from
* Regenerate the Kibana client from that ref
* Build the provider
* Execute the acceptance tests

There's some things we need to figure out:
* Do we automatically apply valid spec updates to the provider. IMO the pipeline should commit client changes based on spec updates which pass CI directly to `main`.
* How are valid breaking changes handled. For example, we update a field to accept both a string and an integer. This isn't a breaking change at the API level, but will potentially break the provider CI as the generated client structure will change. 
  * If we make this check informational then we lose most of the utility, the first breakage will cause all further Kibana PRs to also fail this check and it will quickly be ignored. 
  * One option here is that CI creates a provider PR based on the spec in the Kibana PR. We (i.e the author of the Kibana PR) can then update the provider codebase to handle the changing API, and then merge both provider and Kibana PR's together.

PR #1378, opened 2025-10-17
Reliably run acceptance tests in Kibana client buildkite pipeline

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/1309


internal/kibana/synthetics/private_location/acc_test.go
```diff
@@ -29,14 +30,15 @@ var (
 
 func TestSyntheticPrivateLocationResource(t *testing.T) {
 	resourceId := "elasticstack_kibana_synthetics_private_location.test"
+	randomSuffix := sdkacctest.RandStringFromCharSet(4, sdkacctest.CharSetAlphaNum)
```
nick-benoit, line 33: I'm not quite sure I follow why we need the random suffix for this resource but not for others. Should we add a comment to explain if there is something special about them?

Issue #1368, opened 2025-10-14
Kibana Security Detection Rule Schema Validation Improvements


nick-benoit:
Issue to track a series of validation improvements for the Kibana Security Detection Rule resource. 

* We should move validation occurring in model specific files (eg [this](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/kibana/security_detection_rule/models_machine_learning.go#L63)) with schema based validations ([example](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/utils/validators/conditional.go#L122-L144))
* We should add a schema based validation to ensure that `index` or `data_view_id` are set to mirror the UI validation

andrew-moldovan:
@Kushmaro I don't think this issue is done yet, Nick has a PR open for it though

Kushmaro:
Ah! my bad, closed the wrong issue.

PR #1381, opened 2025-10-21
Security Detection Rule schema validation improvements

nick-benoit:
* Extends existing `conditionalValidation` to support: 
  * `valueRequired` - Requires a value exists if condition is satisfied
  * `valueAllowed` (existing prior use case) - Allows a value to be set only if condition is satisfied
  * `valueForbidden` - Prevents a value from being set if condition is satisfied
* Adds new Asserts validations eg `ListAssert ` to verify dependent value matches allowed values (useful for composing with other validations to build conditional validations based on dependent properties)
* Removes rule type specific validations in `models_*` files in favor of schema validations
* Adds validation to enforce `index` or `data_view_id` is set (except in cases where they are not supported `machine_learning`, `esql`)

Rel: https://github.com/elastic/terraform-provider-elasticstack/issues/1368

### DIFF
--- a/internal/kibana/security_detection_rule/schema.go
+++ b/internal/kibana/security_detection_rule/schema.go
@@ -958,13 +958,7 @@ func (r securityDetectionRuleResource) ValidateConfig(ctx context.Context, req r
 		return
 	}
 
-	t, diag := data.Type.ToStringValue(ctx)
-	resp.Diagnostics.Append(diag...)
-	if resp.Diagnostics.HasError() {
-		return
-	}
-
-	if t.ValueString() == "esql" || t.ValueString() == "machine_learning" {
+	if data.Type.ValueString() == "esql" || data.Type.ValueString() == "machine_learning" {
 		return
 	}
 


internal/utils/validators/conditional.go
tobio, line 1: So I really struggled parsing these validations. There's a couple of things I think we should improve in this file:

* We have duplicate `ListAssert`/`StringAssert` style functions which just return a different interface. Idiomatic go would return the struct and allow the consumer to use it to fulfill the required interface which means we can have a single `Assert` function, and use it for both List and String validators. 
* The `condition` and `conditionValidator` behaviour is similar, but different, and the methods for each are interleaved. So it's hard to track what code matches each struct, but also feels weird that `condition` performs one sort of validation, whilst `conditionValidator` performs 3 others.
* `conditionValidator` providing three different validation paths feels pretty messy. 
* The `Assert`/`Allowance`/`Requirement` naming isn't obvious about the actual validator behaviour. I think we could be more explicit about what's being validated. 

Most of this was like it when you found it, it was probably quite unwieldy before, and feels like we need to improve things now. 

I ended up playing around with this locally, mostly to just focus myself on what was going on. I committed what I'd changed [here](https://github.com/elastic/terraform-provider-elasticstack/commit/2c4fdd65a108d84d3cf0a059bb698c76180f3c19).
* Changes the naming to `DependantPathOneOf`/`AllowedIfDependantPathOneOf`/`ForbiddenIfDependantPathOneOf`...
* Removes the `conditionValidator` entirely, all the different rules are provided as validate functions to the `condition` struct
* Return structs, accept interfaces

I'm not expecting my commit is ready to merge, and can almost certainly be improved upon, and maybe you've got an entirely different, better idea. But I think we should tidy up this file a bit.

nick-benoit: > Most of this was like it when you found it, it was probably quite unwieldy before, and feels like we need to improve things now.

I think I took the existing unwieldyness and scaled it up quite a bit 😅. ++ on this path forward.

nick-benoit: After spending a bit more time here I realized composing validations at the property level results in pretty terrible error messages (example when both `index` / `data_view_id` are missing)
```
│ Error: Invalid Attribute Combination
│ 
│   with elasticstack_kibana_security_detection_rule.test,
│   on deployment.tf line 90, in resource "elasticstack_kibana_security_detection_rule" "test":
│   90: resource "elasticstack_kibana_security_detection_rule" "test" {
│ 
│ No attribute specified when one (and only one) of [index,data_view_id] is required
╵
╷
│ Error: Invalid Configuration
│ 
│   with elasticstack_kibana_security_detection_rule.test,
│   on deployment.tf line 90, in resource "elasticstack_kibana_security_detection_rule" "test":
│   90: resource "elasticstack_kibana_security_detection_rule" "test" {
│ 
│ Attribute 'type' is not one of [machine_learning esql], type is "query"
╵
╷
│ Error: Invalid Attribute Combination
│ 
│   with elasticstack_kibana_security_detection_rule.test,
│   on deployment.tf line 90, in resource "elasticstack_kibana_security_detection_rule" "test":
│   90: resource "elasticstack_kibana_security_detection_rule" "test" {
│ 
│ No attribute specified when one (and only one) of [index,data_view_id] is required
╵
╷
│ Error: Invalid Configuration
│ 
│   with elasticstack_kibana_security_detection_rule.test,
│   on deployment.tf line 90, in resource "elasticstack_kibana_security_detection_rule" "test":
│   90: resource "elasticstack_kibana_security_detection_rule" "test" {
│ 
│ Attribute 'type' is not one of [machine_learning esql], type is "query"
```


At least for the `data_view_id` / `index` validation I added a custom resource level validation: https://github.com/elastic/terraform-provider-elasticstack/pull/1381/commits/a8d9de35e79f8e72e7d3adbeaa4f329ff9a7d110

I think this case is complicated enough it is worth using the custom validation. Though we are still using the property based validation for preventing these values being set for `esql` and `machine_learning` rules. 

For the same example above now we get:
```
│ Error: Invalid Configuration
│ 
│   with elasticstack_kibana_security_detection_rule.test,
│   on deployment.tf line 90, in resource "elasticstack_kibana_security_detection_rule" "test":
│   90: resource "elasticstack_kibana_security_detection_rule" "test" {
│ 
│ One of 'index' or 'data_view_id' must be set.
```

internal/kibana/security_detection_rule/schema.go
```diff
@@ -913,3 +940,47 @@ func getThreatSubtechniqueElementType() attr.Type {
...
+		return
+	}
+
+	t, diag := data.Type.ToStringValue(ctx)
+	resp.Diagnostics.Append(diag...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if t.ValueString() == "esql" || t.ValueString() == "machine_learning" {
```
tobio: Isn't it already and StringValue?

```suggestion
	if data.Type..ValueString() == "esql" || data.Type..ValueString() == "machine_learning" {
```

nick-benoit: https://github.com/elastic/terraform-provider-elasticstack/pull/1381/commits/b2cce8e963b9ee0a1f6e4c127acca5958660e834

Issue #1304, opened 2025-09-15
[Feature] Add support for space-aware Agent Policies in Fleet


jalogisch:
**Is your feature request related to a problem? Please describe.**
Yes. When managing Fleet, I'm unable to fully automate our setup using Terraform because there's no way to assign an Agent Policy to a specific Kibana space. This prevents us from using an Infrastructure-as-Code approach for environments that rely on space-aware policies, forcing us to perform manual configuration steps in the UI after running Terraform.

**Describe the resource you would like to have implemented.**
The request is to enhance the existing `elasticstack_fleet_agent_policy` resource.

**Describe the solution you'd like**
I would like a new attribute, such as `space_id`, to be added to the `elasticstack_fleet_agent_policy` resource. This would allow a user to specify the ID of the Kibana space that the policy should belong to, enabling the creation and management of space-aware policies directly from Terraform.

For example:

```hcl
resource "elasticstack_fleet_agent_policy" "space_aware_policy" {
  name        = "Policy for Marketing Space"
  namespace   = "default"
  
  # Proposed new attribute
  space_id    = "marketing" 
}
```

**Describe alternatives you've considered**
The only alternative is to create the policies manually in the Kibana UI after the rest of the infrastructure has been deployed with Terraform. This is not a scalable or desirable solution as it circumvents the benefits of Infrastructure-as-Code.

**Additional context**
This feature is crucial for organizations that use Kibana spaces for multi-tenancy or to separate resources by team. This was observed in provider version `0.11.17` but appears to be a limitation in the current version as well.

tobio:
Fixed by #1390

PR #1390, opened 2025-10-23
Added space awareness to Agent Policies

jamesagarside:
## Overview

This PR implements comprehensive Kibana Spaces support across all Fleet resources, enabling multi-tenant Fleet management within the Terraform provider. Users can now create and manage Fleet resources (agent policies, integrations, outputs, server hosts) in specific Kibana spaces instead of being limited to the default space.

Completes #1304 

### New Feature: `space_ids` Attribute

All Fleet resources now support an optional `space_ids` field:
- **Default**: `["default"]` (backward compatible)
- **Affects**: Agent policies, integration policies, outputs, server hosts, integrations, enrollment tokens

### Resources Updated

- `elasticstack_fleet_agent_policy`
- `elasticstack_fleet_integration_policy`  
- `elasticstack_fleet_output`
- `elasticstack_fleet_server_host`
- `elasticstack_fleet_integration`
- `elasticstack_fleet_enrollment_tokens` (data source)

### Implementation

**1. Space-Aware API Methods** ([internal/clients/fleet/fleet.go](internal/clients/fleet/fleet.go))

Added 650+ lines of space-aware client methods that construct space-specific API paths (`/s/{space_id}/api/fleet/...`):
- Agent Policy: `GetAgentPolicyInSpace`, `CreateAgentPolicyInSpace`, `UpdateAgentPolicyInSpace`, `DeleteAgentPolicyInSpace`
- Outputs, Server Hosts, Package Policies: Similar space-aware CRUD methods

**2. CRUD Updates Across All Resources**

- Modified create/read/update/delete operations to use space-aware methods when `space_ids` is set
- Integration policies inherit `space_ids` from parent agent policy
- Added state preservation logic to prevent drift

**3. State Management Fixes**

- Added `UseStateForUnknown` plan modifiers to prevent spurious diffs
- Fixed `space_ids` preservation during update operations
- Resolved state drift showing non-empty plans on refresh

**4. Testing & Documentation**

- Added version checks (Stack 9.1.0+ required)
- Updated all resource documentation with `space_ids` examples
- Enhanced acceptance tests with space validation

## Example

```hcl
resource "elasticstack_fleet_agent_policy" "my_policy" {
  name      = "my-space-policy"
  space_ids = ["my-custom-space"]
}

resource "elasticstack_fleet_integration_policy" "nginx" {
  name             = "nginx-policy"
  agent_policy_id  = elasticstack_fleet_agent_policy.my_policy.policy_id
  space_ids        = ["my-custom-space"]
  integration_name = "nginx"
}
```

### Added
* space_ids field to all Fleet resources and enrollment tokens data source
* Space-aware API client methods for Fleet operations
* Version-based testing for Stack 9.1.0+ features

### Fixed
* State drift issues with space_ids causing non-empty plans
* Integration policy import failures in space-scoped configurations
* Output space_ids preservation during updates
* Backward Compatibility
* Fully backward compatible - space_ids defaults to ["default"]

## Requirements
* Minimum Stack Version: 9.1.0 for space-aware Fleet features
* Older versions gracefully fall back to default space

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Fix `elasticstack_elasticsearch_snapshot_lifecycle` metadata type conversion causing terraform apply to fail ([#1409](https://github.com/elastic/terraform-provider-elasticstack/issues/1409))
 - Add new `elasticstack_elasticsearch_ml_anomaly_detection_job` resource ([#1329](https://github.com/elastic/terraform-provider-elasticstack/pull/1329))
 - Add new `elasticstack_elasticsearch_ml_datafeed` resource ([1340](https://github.com/elastic/terraform-provider-elasticstack/pull/1340))
+- Add `space_ids` attribute to all Fleet resources to support space-aware Fleet resource management ([#1390](https://github.com/elastic/terraform-provider-elasticstack/pull/1390))
 
 ## [0.12.1] - 2025-10-22
 - Fix regression restricting the characters in an `elasticstack_elasticsearch_role_mapping` `name`. ([#1373](https://github.com/elastic/terraform-provider-elasticstack/pull/1373))


internal/clients/fleet/fleet.go
```diff
@@ -16,6 +19,16 @@ var (
 	ErrPackageNotFound = errors.New("package not found")
 )
 
+// buildSpaceAwarePath constructs an API path with space awareness.
+// If spaceID is empty or "default", returns the basePath unchanged.
+// Otherwise, prepends "/s/{spaceID}" to the basePath.
+func buildSpaceAwarePath(spaceID, basePath string) string {
```
tobio, line 24: The client generator is injecting this `space_id` parameter for a subset of [kibana paths](https://github.com/elastic/terraform-provider-elasticstack/blob/main/generated/kbapi/transform_schema.go#L684-L722). The benefit there is that we don't have to redefine al the API operations and manually build up the HTTP requests. 

Are you able to:
* Extract the code which injects the space id parameter into it's own dedicated function. You'll need to add that function into the list of transformers [here](https://github.com/elastic/terraform-provider-elasticstack/blob/main/generated/kbapi/transform_schema.go#L562)
* Add the fleet paths to the `spaceIdPaths` slice so the path parameter is injected into the client for us. 
* Refactor this code to remove the space id variants you've added, and instead just supply the expected space to the API client.

internal/fleet/agent_policy/delete.go
```diff
@@ -30,6 +34,19 @@ func (r *agentPolicyResource) Delete(ctx context.Context, req resource.DeleteReq
...
 	}
 
-	diags = fleet.DeleteAgentPolicy(ctx, client, policyID)
+	// If space_ids is set in state, use space-aware DELETE request
+	if !stateModel.SpaceIds.IsNull() && !stateModel.SpaceIds.IsUnknown() {
+		var tempDiags diag.Diagnostics
+		spaceIDs := utils.ListTypeAs[types.String](ctx, stateModel.SpaceIds, path.Root("space_ids"), &tempDiags)
+		if !tempDiags.HasError() && len(spaceIDs) > 0 {
+			spaceID := spaceIDs[0].ValueString()
+			diags = fleet.DeleteAgentPolicyInSpace(ctx, client, policyID, spaceID)
```
tobio: Let's assume there are multiple space IDs. 

Does this entirely delete the policy, or does it simply remove the policy from the specified space? From the perspective of an API consumer I would expect the latter, in which case I'd assume we'd need to call this for each space in the slice.

internal/fleet/agent_policy/create.go
```diff
@@ -41,6 +46,41 @@ func (r *agentPolicyResource) Create(ctx context.Context, req resource.CreateReq
 		return
 	}
 
+	// The CREATE response may not include all fields (e.g., space_ids can be null in the response
+	// even when specified in the request). Read the policy back to get the complete state.
+	// Only do this if we got a valid ID from the create response.
+	if policy != nil && policy.Id != "" {
```
tobio, line 52: IIUC this is the same as the core of the `Read` function. Are we able to extract this and re-use it here, in read, and presumably in `Update` as well?

internal/fleet/agent_policy/read.go
```diff
@@ -23,7 +28,22 @@ func (r *agentPolicyResource) Read(ctx context.Context, req resource.ReadRequest
...
-	policy, diags := fleet.GetAgentPolicy(ctx, client, policyID)
+
+	// If space_ids is set in state, use space-aware GET request
+	var policy *kbapi.AgentPolicy
+	if !stateModel.SpaceIds.IsNull() && !stateModel.SpaceIds.IsUnknown() {
+		var tempDiags diag.Diagnostics
+		spaceIDs := utils.ListTypeAs[types.String](ctx, stateModel.SpaceIds, path.Root("space_ids"), &tempDiags)
+		if !tempDiags.HasError() && len(spaceIDs) > 0 {
+			spaceID := spaceIDs[0].ValueString()
+			policy, diags = fleet.GetAgentPolicyInSpace(ctx, client, policyID, spaceID)
```
tobio: What happens if we prepend a space to the list of space IDs after the policy is created. IIUC at this point the policy won't exist in the space, and this call will 404, which would cause Terraform to attempt to create the policy again. 

To be clear, what's the behaviour when:
1. Create a new policy with `space_ids = ["default", "non-default"]`
2. Update the policy to `space_ids = ["another-space", "default", "non-default"]
3. `terraform apply`

I'd imagine the correct behaviour would be to just add the existing policy to `another-space`. As the code is written I think what will happen is:
* The original policy will get removed from TF state since it doesn't exist in `another-space`. The actual policy in Kibana will be become orphaned.
* A new policy will be created in all three spaces

internal/fleet/enrollment_tokens/read.go
```diff
@@ -27,10 +27,36 @@ func (d *enrollmentTokensDataSource) Read(ctx context.Context, req datasource.Re
...
+	var spaceID string
+	if !model.SpaceIds.IsNull() && !model.SpaceIds.IsUnknown() {
+		var spaceIDs []types.String
+		diags = model.SpaceIds.ElementsAs(ctx, &spaceIDs, false)
+		resp.Diagnostics.Append(diags...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+		if len(spaceIDs) > 0 {
+			spaceID = spaceIDs[0].ValueString()
```
tobio: Shouldn't we query all the defined spaces here, not just the first? Or change this to `space_id` so users can only supply a single value?

internal/fleet/integration_policy/models.go
```diff
@@ -91,12 +92,45 @@ func (model *integrationPolicyModel) populateFromAPI(ctx context.Context, data *
...
+	// 3. If model.Input is known and has values: user configured inputs → populate from API
+
+	isInputKnown := utils.IsKnown(model.Input)
+	isInputNullOrEmpty := model.Input.IsNull() || (isInputKnown && len(model.Input.Elements()) == 0)
+
+	// Case 1: Unknown (import/fresh read) - always populate
+	if !isInputKnown {
+		// Import or fresh read - populate everything from API
+		// (continue to normal population below)
+	} else if isInputNullOrEmpty {
```
tobio, line 126:
```suggestion
	if model.Input.IsNull() {
```

Couple of questions:
* Can we just do away with the first condition here? It's not clear what it's preventing from falling through to the empty case
* `ifInputKnown && len() == 0`. `input` is currently defined as a block, so I'm not sure what the behaviour is here. I'd expect `isInputKnown` means that the user has defined an `input` somewhere, and that if that collection is empty then it's been explicitly defined as empty and we should populate the model. I assume that's wrong though since you've written this code, is there an example of a TF config which hits each of these cases?

tobio (commented):
Thanks for taking the time to get this added, it'll be really useful :) 

All the comments on `agent_policy` apply to the other resources as well. I think we need to explore the behaviour here when the list of spaces changes.

internal/clients/fleet/fleet.go
jamesagarside: Looks like this is Blocked by upstream Kibana OpenAPI [spec](https://www.elastic.co/docs/api/doc/kibana.json)
Confirmed Fleet endpoints exist (/api/fleet/agent_policies, /api/fleet/outputs, etc.)
No space-aware variants (/s/{spaceId}/api/fleet/*) exist in the spec yet
The Kibana team hasn't added space aware Fleet support to their official OpenAPI specification

## Current Approach:
The *InSpace() wrapper functions work correctly with the default-space-first fix
These wrappers provide necessary error handling and response parsing
Removing them would require either:
Kibana adding /s/{spaceId}/api/fleet/* endpoints to their OpenAPI spec
Manually adding these paths to the transform_schema.go (causes duplicate type definitions)

## Recommendation
Keep current wrapper functions until Kibana updates their OpenAPI spec upstream.

tobio: I've added a commit to this PR demonstrating a different approach which removes all this duplication.

internal/fleet/agent_policy/delete.go
```diff
@@ -30,6 +31,18 @@ func (r *agentPolicyResource) Delete(ctx context.Context, req resource.DeleteReq
 		return
 	}
 
-	diags = fleet.DeleteAgentPolicy(ctx, client, policyID)
+	// Extract space IDs from state and determine operational space
+	// NOTE: DELETE removes the policy from ALL spaces (global delete)
+	// To remove from specific spaces only, UPDATE space_ids instead of deleting
+	stateSpaceIDs := fleetutils.ExtractSpaceIDs(ctx, stateModel.SpaceIds)
```
tobio:
```suggestion
	stateSpaceIDs :=  utils.ListTypeToSlice_String(ctx, stateModel.SpaceIds, path.Root("space_ids"), &resp.Diagnostics)
```

We can delete `ExtractSpaceIDs` entirely and just use the existing util function.

internal/fleet/agent_policy/update.go
```diff
@@ -36,14 +38,39 @@ func (r *agentPolicyResource) Update(ctx context.Context, req resource.UpdateReq
 	}
 
 	policyID := planModel.PolicyID.ValueString()
-	policy, diags := fleet.UpdateAgentPolicy(ctx, client, policyID, body)
+
+	// Extract space IDs from plan and determine operational space
+	// Using default-space-first model for stable multi-space updates
+	planSpaceIDs := fleetutils.ExtractSpaceIDs(ctx, planModel.SpaceIds)
+	spaceID := fleetutils.GetOperationalSpace(planSpaceIDs)
```
tobio: There's still a bug here:

1. Create a policy with  `space_ids = ["space-a"]`
2. Update the policy to `space_ids = ["space-b", "space-a"]`
3. 💥 

```suggestion
	// Read the existing spaces from state to avoid updating the policy 
	// in a space where it's not yet visible.     
	var stateSpaces basetypes.ListValue
	resp.Diagnostics.Append(req.State.GetAttribute(ctx, path.Root("space_ids"), &stateSpaces)...)
	if resp.Diagnostics.HasError() {
		return
	}
	
	// Extract space IDs from plan and determine operational space
	// Using default-space-first model for stable multi-space updates
	stateSpaceIDs := fleetutils.ExtractSpaceIDs(ctx, stateSpaces)
	spaceID := fleetutils.GetOperationalSpace(stateSpaceIDs)
```

We'll have the same behaviour across all these resources so you could extract that entire snippet to `fleetutils` 

```
func GetOperationalSpace(ctx context.Context, state tfsdk.State) (string, diag.Diagnostics) {
	// Read the existing spaces from state to avoid updating the policy 
	// in a space where it's not yet visible.     
	var stateSpaces basetypes.ListValue
	diags := state.GetAttribute(ctx, path.Root("space_ids"), &stateSpaces)
	if diags.HasError() {
		return "", diags
	}
	
	// Extract space IDs from plan and determine operational space
	// Using default-space-first model for stable multi-space updates
	...
}
```

In fact, using spaces from state means we don't need to prefer reading from the `default` space and could go back to just reading from the first space in the collection since there's no real benefit either way.

internal/fleet/agent_policy/update.go
```diff
@@ -36,14 +38,39 @@ func (r *agentPolicyResource) Update(ctx context.Context, req resource.UpdateReq
...
+	// Preserve the space_ids order from plan before populating from API
+	// The Kibana API may return space_ids in a different order (sorted),
+	// but we need to maintain the user's configured order to avoid false drift detection
+	originalSpaceIds := planModel.SpaceIds
+
 	planModel.populateFromAPI(ctx, policy)
 
+	// Restore the original space_ids order if appropriate
+	// See ShouldPreserveSpaceIdsOrder documentation for edge case handling
+	if fleetutils.ShouldPreserveSpaceIdsOrder(policy.SpaceIds, originalSpaceIds, planModel.SpaceIds) {
```
tobio: We can do away with this `ShouldPreserveSpaceIdsOrder` palaver if we switch `space_ids` to be a `SetAttribute` rather than a `ListAttribute`.

internal/fleet/agent_policy/update.go
jamesagarside: Created new GetOperationalSpaceFromState() helper that reads from STATE (current) instead of PLAN (desired):
```
// GetOperationalSpaceFromState extracts the operational space ID from Terraform state.
// This helper reads space_ids from state (not plan) to determine which space to use
// for API operations, preventing errors when space_ids changes (e.g., prepending a new space).
func GetOperationalSpaceFromState(ctx context.Context, state tfsdk.State) (string, diag.Diagnostics) {
    var stateSpaces types.Set
    diags := state.GetAttribute(ctx, path.Root("space_ids"), &stateSpaces)
    // ... extracts first space from STATE
    return spaceIDs[0], nil
}
```
Applied to ALL CRUD operations:

agent_policy: [read.go:30](https://github.com/elastic/terraform-provider-elasticstack/blob/space-aware-fleet/internal/fleet/agent_policy/read.go#L30), [update.go:43](https://github.com/elastic/terraform-provider-elasticstack/blob/space-aware-fleet/internal/fleet/agent_policy/update.go#L43), [delete.go:34](https://github.com/elastic/terraform-provider-elasticstack/blob/space-aware-fleet/internal/fleet/agent_policy/delete.go#L34)
server_host: read.go, update.go, delete.go
output: read.go, update.go, delete.go
integration_policy: read.go, update.go, delete.go

Now when prepending spaces, we query the existing space where the resource is known to exist, then the API handles adding it to the new space. No more 404s! 

Code reference: [space_utils.go:15-58](https://github.com/elastic/terraform-provider-elasticstack/blob/space-aware-fleet/internal/fleet/space_utils.go#L15-L58)

internal/fleet/agent_policy/update.go
jamesagarside: Good spot. Should be sorted now

internal/fleet/agent_policy/delete.go
jamesagarside: removed in latest

internal/clients/fleet/fleet.go
jamesagarside: @tobio Thanks for the request editor example with enrolment tokens! Makes total sense - I'll apply that pattern to all the remaining Fleet resources in this PR.

Also fixed two small bugs I spotted:
- Removed extra `UseStateForUnknown()` from integration schema that shouldn't have been there (2b13928)
- Fixed a test assertion still using List indexing instead of Set matching (1e423ad)

Issue #1393, opened 2025-10-24
[Bug] Multiple Authorization headers sent in some cases


nick-benoit:
**Describe the bug**

Multiple Authorization headers sent when Kibana or Fleet is configured with a differing authoriation mechanism than Elasticsearch.

**To Reproduce**

1. Configure provider to use Elasticsearch with apikey and Kibana with basic auth 

```hcl
provider "elasticstack" {
  elasticsearch {
    endpoints = ["https://<redacted-es-endpoint>:9200"]
    api_key   = var.elasticsearch_api_key
  }
  kibana {
    endpoints = ["https://<redacted-kibana-endpoint>"]
    username = "elastic"
    password = "<redacted>"
  }
}
```

2. Apply plan to create or modify a resource with debug logs eg (`TF_LOG=debug terraform apply `)

3. Verify in logs multiple Authorization headers are sent in requests to Kibana

Note that if username / password are excluded from the `kibana` block only a single Authorization is sent. 

**Expected behavior**

A single authorization header is sent regardless of authorization configuration. This aligns with [RFC 7230](https://www.rfc-editor.org/rfc/rfc7230#section-3.2.2). This is reported to have caused 400 responses when Elasticsearch is behind some proxies. 

**Debug output**

```
2025-10-21T16:57:42.860+0200 [DEBUG] provider.terraform-provider-elasticstack_v0.12.0: Fleet API Request Details:
---[ REQUEST ]---------------------------------------
POST /api/fleet/outputs HTTP/1.1
Host: <redacted>
User-Agent: Go-http-client/1.1
Content-Length: 4611
Authorization: ***************************************************
Authorization: ********************************************************************
Content-Type: application/json
Kbn-Xsrf: true
Accept-Encoding: gzip
{
 "config_yaml": "{}\n",
 "hosts": [
  "logstash:5444"
 ],
 "id": "",
 "is_default": false,
 "is_default_monitoring": false,
 "name": "Logstash test",
 "ssl": {
  "certificate": "-----BEGIN CERTIFICATE-----\nMII(...)=\n-----END CERTIFICATE-----\n",
  "certificate_authorities": [
   "-----BEGIN CERTIFICATE-----\nMII(...)=\n-----END CERTIFICATE-----\n"
  ],
  "key": "-----BEGIN PRIVATE KEY-----\nMII(...)=\n-----END PRIVATE KEY-----\n"
 },
 "type": "logstash"
}
```

[...]

PR #1394, opened 2025-10-25
Respect Kibana / Fleet config schema when inheriting ES credentials

nick-benoit:
Fixes: https://github.com/elastic/terraform-provider-elasticstack/issues/1393

Before this PR we were using the ES Config as a starting point and applying Kibana provider block and environment configs on top of that. This can lead to invalid Configs where both Username / Password and ApiKey are both configured. 

This PR starts with an empty Kibana / Fleet config, then applies provider settings, environment overrides, then only applies defaults for unknown values (while respecting mutually exclusive values like Username / ApiKey)

Additionally we add `fleetConfigKeys` and `kibanaConfigKeys` which allows us to track explicitly set fields. This adds a bit of overhead to maintain state, but allows us to correctly differentiate between explicit empty values and unset values in a way that supports both the sdkv2 and plugin framework implementations. Technically the Kibana config doesn't require tracking the keys because it only inherits string values from ES, but I thought it was more straightforward for Fleet and Kibana to use a similar pattern.

### DIFF
--- a/generated/kbapi/Makefile
+++ b/generated/kbapi/Makefile
@@ -2,7 +2,7 @@
 SHELL := /bin/bash
 ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
-github_ref ?= 5c4f76696e63bf9e9a53d55521f4c18faa02ccf2
+github_ref ?= dcff88f37b458e385b7d2b1b82d765d5ae887ca3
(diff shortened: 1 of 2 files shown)


nick-benoit:
Woops, not sure why this doesn't seem to be playing nice with the SLO test. I'll switch back to a draft while I sort that out.

PR #1422, opened 2025-11-06
Add resource to manage ML datafeed state

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/813

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,10 @@
 
 - Fix `elasticstack_elasticsearch_snapshot_lifecycle` metadata type conversion causing terraform apply to fail ([#1409](https://github.com/elastic/terraform-provider-elasticstack/issues/1409))
 - Add new `elasticstack_elasticsearch_ml_anomaly_detection_job` resource ([#1329](https://github.com/elastic/terraform-provider-elasticstack/pull/1329))
-- Add new `elasticstack_elasticsearch_ml_datafeed` resource ([1340](https://github.com/elastic/terraform-provider-elasticstack/pull/1340))
+- Add new `elasticstack_elasticsearch_ml_datafeed` resource ([#1340](https://github.com/elastic/terraform-provider-elasticstack/pull/1340))
 - Add `space_ids` attribute to all Fleet resources to support space-aware Fleet resource management ([#1390](https://github.com/elastic/terraform-provider-elasticstack/pull/1390))
+- Add new `elasticstack_elasticsearch_ml_job_state` resource ([#1337](https://github.com/elastic/terraform-provider-elasticstack/pull/1337))
+- Add new `elasticstack_elasticsearch_ml_datafeed_state` resource ([#1422](https://github.com/elastic/terraform-provider-elasticstack/pull/1422))
 
 ## [0.12.1] - 2025-10-22
 - Fix regression restricting the characters in an `elasticstack_elasticsearch_role_mapping` `name`. ([#1373](https://github.com/elastic/terraform-provider-elasticstack/pull/1373))


internal/elasticsearch/ml/datafeed_state/read.go
```diff
@@ -0,0 +1,72 @@
...
+	data.State = types.StringValue(datafeedStats.State)
+
+	if datafeedStats.State == "started" {
+		if datafeedStats.RunningState == nil {
+			diags.AddWarning("Running state was empty for a started datafeed", "The Elasticsearch API returned an empty running state for a Datafeed which was successfully started. Ignoring start and end response values.")
+		}
+
+		data.Start = timetypes.NewRFC3339TimeValue(time.UnixMilli(datafeedStats.RunningState.SearchInterval.StartMS))
+		if datafeedStats.RunningState.RealTimeConfigured {
+			data.End = timetypes.NewRFC3339Null()
```
nick-benoit: I don't feel strongly here, but I wonder if it is worth adding a warning diag here if we are in this case and and `datafeedStats.RunningState.SearchInterval.EndMS` is configured?

internal/elasticsearch/ml/datafeed/state_utils.go
```diff
@@ -2,35 +2,65 @@ package datafeed
...
 	}
 
-	return statsResponse.State, nil
+	return &statsResponse.State, nil
 }
 
-// waitForDatafeedState waits for a datafeed to reach the desired state
-func (r *datafeedResource) waitForDatafeedState(ctx context.Context, datafeedId, desiredState string) error {
+var terminalDatafeedStates = map[string]struct{}{
+	"stopped": {},
```
nick-benoit: Nit: I wonder if these should be constants defined somewhere?

internal/elasticsearch/ml/datafeed_state/acc_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/acctest"
+	"github.com/hashicorp/terraform-plugin-testing/config"
+	sdkacctest "github.com/hashicorp/terraform-plugin-testing/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+)
+
```
nick-benoit, line 13: Is there a good way to add test coverage for `datafeed_timeout` and `timeouts`? I guess since they aren't written to any particular resource we read back from ES it might be a bit tricky. It would be nice to have some test coverage though.

nick-benoit (approved):
Nice. Having the resource to manage the state is a cool idea. Just a few optional questions / thoughts.

internal/elasticsearch/ml/datafeed_state/acc_test.go
tobio: Yea, good question. I've added a test to verify the timeouts on the *job state* resource, but I can't think of a way to do the same on the datafeed (i.e artificially block the datafeed from starting).

internal/elasticsearch/ml/datafeed_state/update.go
```diff
@@ -0,0 +1,220 @@
...
+		return nil, diags
+	}
+
+	// It's possible that the datafeed starts, and then immediately stops if there is no (or very little) data to process.
+	// In this case, the state transition may occur too quickly to be detected by the wait function.
+	// To handle this, we check if the search count has increased to determine if the datafeed actually started since the update.
+	if statsAfterUpdate.TimingStats == nil || datafeedStats.TimingStats == nil {
+		diags.AddWarning("Expected Datafeed to contain timing stats",
+			fmt.Sprintf("Stats for datafeed %s did not contain timing stats either before or after the update. Before %v - After %v", datafeedId, datafeedStats, statsAfterUpdate))
+	} else if statsAfterUpdate.TimingStats.SearchCount <= datafeedStats.TimingStats.SearchCount {
```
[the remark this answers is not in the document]
nick-benoit: Based on your comment above it seems like your code is actually correct. IIUC If the count increases we know the state briefly transitioned. If the count does not increase we don't actually know if there was a state transition.

internal/elasticsearch/ml/job_state/update.go
```diff
@@ -72,7 +77,7 @@ func (r *mlJobStateResource) update(ctx context.Context, plan tfsdk.Plan, state
 	}
 
 	// Perform state transition if needed
-	fwDiags = r.performStateTransition(ctx, client, data, *currentState, operationTimeout)
+	fwDiags = r.performStateTransition(ctx, client, data, *currentState)
```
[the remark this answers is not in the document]
nick-benoit: I guess in this case we are using `data.Timeout.Parse()` in `performStateTransition` instead?

nick-benoit (approved):
Changes from my comments look good 👍 Pending you resolve the couple of new copilot comments. 

Sidenote, it seems to me this is a notably more useful review than we were seeing before.

internal/elasticsearch/ml/job_state/update.go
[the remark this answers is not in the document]
tobio: The context being passed to `performStateTransition` already has the a timeout applied, so passing this value is redundant.

PR #1429, opened 2025-11-07
First take at documented coding standards

tobio:
The intention here is to try and increase the value we get out of Copilot code review. Overall, the aim is to allow human reviewers to focus attention on higher level concerns, with confidence the AI reviews are enforcing more formulaic coding style/expectations, kind of lint on steroids.

### DIFF
--- a/CODING_STANDARDS.md
+++ b/CODING_STANDARDS.md
@@ -52,6 +52,9 @@ This document outlines the coding standards and conventions used in the terrafor
   - Further details may be found in the [API documentation](https://www.elastic.co/docs/api/doc/kibana/v9/)
 - Elasticsearch resources will be backed by the [go-elasticsearch](https://github.com/elastic/go-elasticsearch) client. 
   - Further details may be found in the [API documentation](https://www.elastic.co/docs/api/doc/elasticsearch/)
+- Use `EnforceMinVersion` to ensure the backing Elastic Stack applications support the defined fields. 
+  - The provider supports a wide range of Stack versions, and so newer features will not be available in all versions. 
+  - See [`assertKafkaSupport`](./internal/fleet/output/models.go) for an example of how to handle the use of unsupported attributes.
 
 
 ## JSON Handling


CODING_STANDARDS.md
```diff
@@ -0,0 +1,64 @@
...
+
+- Write idiomatic Go.
+  - [Effective Go](https://go.dev/doc/effective_go)
+  - [Code Review Comments](https://go.dev/wiki/CodeReviewComments)
+  - The [Google Styleguide](https://google.github.io/styleguide/go/index#about) 
+
+## Project Structure
+
+- Use the Plugin Framework for all new resources (not SDKv2)
+- Follow the code organization pattern of `internal/elasticsearch/security/system_user` for new Plugin Framework resources
```
nick-benoit: I wonder if it's worth listing all the files in that directory and explaining what they are for to make it more straightforward for copilot to understand the structure? Alternatively adding a file like this to the directory and listing that here could be useful. This is probably not necessary for all resources, but I think could be good for our "canonical" resource. 

It seems like giving copilot the summary might yield more consistent performance than having copilot parse the files and comprehend how they all fit together in one go.

CODING_STANDARDS.md
```diff
@@ -0,0 +1,64 @@
...
+
+- Use custom types to model attribute specific behaviour.
+    - Use `jsontypes.NormalizedType{}` custom type for string attributes containing JSON blobs.
+    - Use `customtypes.DurationType{}` for duration-based string attributes.
+    - Use `customtypes.JSONWithDefaultsType{}` to allow users to specify only a subset of a JSON blob.
+- Always include comprehensive descriptions for all resources, and attributes. 
+- Long, multiline descriptions should be stored in an external markdown file, which is imported via Golang embedding. See `internal/elasticsearch/security/system_user/resource-description.md` for an example location.
+- Use schema validation wherever possible. Only perform validation within create/read/update functions as a last resort. 
+  - For example, any validation that relies on the actual Elastic Stack components (e.g Elasticsearch version)
+    can only be performed during the create/read/update phase.
```
nick-benoit, line 50: I wonder if we should be explicitly asking co-pilot to generate the OAS and consider this when generating schema definitions?

nick-benoit: I guess we could also consider committing this file? I suppose currently we are ignoring the client oas file. I'm sure this would lead to some annoying merge conflicts etc, but maybe is worth it in order to make it easier for copilot to consider the oas?

CODING_STANDARDS.md
```diff
@@ -0,0 +1,64 @@
...
+    can only be performed during the create/read/update phase. 
+
+## JSON Handling
+
+- Use `jsontypes.NormalizedType{}` for JSON string attributes to ensure proper normalization and comparison.
+- Use `customtypes.JSONWithDefaultsType{}` if API level defaults may be applied automatically. 
+
+## Resource Implementation
+
+- Follow the pattern: `resource.go`, `schema.go`, `models.go`, `create.go`, `read.go`, `update.go`, `delete.go`
```
nick-benoit: This seems like it maybe overlaps with `## Project Structure`. I wonder if we should combine both of those here? Or move higher level details to "Project Structure" and reserve this section for more fiddly details like "`utils.IsKnown(val)` instead of `!val.IsNull() && !val.IsUnknown()`"

I don't feel strongly about this, just a thought.

nick-benoit (approved):
Nice, i'm excited to see how this works. Thanks for pushing this forward.

CODING_STANDARDS.md
tobio: I know I've started this within the context of Copilot, but this is also pretty helpful for actual humans coming into the project :) 

Either way, this was a good call.

CODING_STANDARDS.md
tobio: I've combined this with the sections above. If they start getting too big then we can look at restructuring things.

CODING_STANDARDS.md
tobio: > I'm sure this would lead to some annoying merge conflicts etc,

Now Renovate is handling the updates we shouldn't have humans cherry-picking changes, so I think these updates should be pretty clean. I'll add it in.

Issue #847, opened 2024-10-15
[Bug] `.gen-ai` and AWS Bedrock `.bedrock` missing as valid connector ids for `elasticstack_kibana_action_connector`


ebeahan:
**Describe the bug**

Same issue identified for `.gemini` in #727 also affecting other LLM provider Kibana connectors:

* OpenAI/Azure OpenAI - `.gen-ai`
* AWS Bedrock - `.bedrock`

Trying to use these values as `connector_type_id` in a `elasticstack_kibana_action_connector` will fail with the error:

```
Error: unknown connector type [.bedrock]
```

PR #1467, opened 2025-11-20
Support .bedrock and .gen-ai connectors

dimuon:
Closes https://github.com/elastic/terraform-provider-elasticstack/issues/847


internal/kibana/connectors/acc_test.go
tobio, line 1: Can we restructure this test file to be consistent with [this pattern](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/fleet/integration_policy/acc_test.go#L50-L53) in a follow up PR?

That is:
* Move all the Terraform HCL into external files referenced by `NamedTestCaseDirectory`. 
* Move the name (and any other variables) into a `ConfigVariables` map.

PR #1473, opened 2025-11-21
Streams support  POC

rStelmach:
## Summary

 Add a new `elasticstack_kibana_stream` Terraform resource backed by the main `kbapi` Kibana client, plus a small internal condition model and Streams API helpers, to validate the first end‑to‑end integration between Terraform and Streams.  
This POC focuses on **group streams**, base‑stream creation, and a **read‑only view of ingest**, without committing yet to a final public Terraform schema for all Streams features.

## What this PR changes

- **Streams API helpers** (`internal/clients/kibana_oapi/streams.go`)

  - Wrap the existing `kbapi.ClientWithResponses` on top of the Kibana HTTP client.
  - Provide focused helpers for:
    - `EnableStreams` / `DisableStreams` (`POST /api/streams/_enable|_disable`),
    - `GetStreamJSON`, `PutStreamRaw`, `DeleteStream` (with the `"Expected undefined, received null"` quirk handled),
    - `GetStreamIngestJSON` / `PutStreamIngest`,
    - `GetStreamGroupJSON` / `PutStreamGroup`.

- **New Terraform resource: `elasticstack_kibana_stream`** (`internal/kibana/streams`)

  - Registered in `provider/plugin_framework.go` via `streams.NewResource`.
  - Schema + models:
    - Top‑level: `id`, `name`, `space_id`, `description`, `type`, `create_if_missing`.
    - `group { members, metadata, tags }` – managed via `/api/streams/{name}/_group`.
    - `ingest { type }` – computed, read‑only from `/api/streams/{name}/_ingest` in this POC.
  - CRUD behavior:
    - Group streams: optionally create the base stream, upsert `_group`, round‑trip `members`/`metadata`/`tags`, and delete via `DELETE /api/streams/{name}` with quirks handled.
    - Ingest streams: read‑only (Terraform reads `_ingest` but does not write it yet) .

- **Internal condition expression model** (`internal/kibana/streams/conditions.go`)

  - Small internal tree (`Condition`, `FieldComparison`, `And`, `Or`) plus `MarshalCondition` helper.
  - Unit tests in `conditions_test.go` verify JSON output for simple and nested trees.

#### Kibana's side :
- generate logs 

#### Terraform side

- Unit tests:
  - `go test ./internal/kibana/streams/...` (condition model + flatten/expand helpers).
- Manual POC validation:
  - Adjust kibana url in new `main.tf`
  - Build the provider: `make build`.
  - cd streams then run:
    - `terraform init`
    - `terraform apply`
  - Confirm:
    - Group streams: base stream creation via `create_if_missing = true`, `_group` upsert, and round‑trip of `members` / `metadata` / `tags`.
    - Ingest streams: read‑only ingest behavior (`ingest.type` populated from `_ingest`, no writes to `_ingest`).


insukcho:
@CoenWarmer @rStelmach https://github.com/elastic/streams-program/issues/683 has been merged a month ago! Are we resuming this issue now? Any rough ETA I can share to the customer?

insukcho:
@tobio It seems this issue has been closed. Did we provide the feature described in this ER?

tobio:
Streams support was added in https://github.com/elastic/terraform-provider-elasticstack/pull/2027 but is still marked as experimental @CoenWarmer are we able to make this resource available to customers now?

CoenWarmer:
> Streams support was added in https://github.com/elastic/terraform-provider-elasticstack/pull/2027 but is still marked as experimental @CoenWarmer are we able to make this resource available to customers now? 

@3kt @gbamparop @flash1293 @mdbirnstiehl thoughts?

3kt:
Yes, I don't see any reason to keep it away from them - while the feature isn't GA I don't anticipate the controls to change. Some things will need polishing in Elasticsearch (query for unmapped fields, etc...) but the backend plumbing is good and should be exposed through TF ASAP.
@CoenWarmer shall I open another issue for this or is it a "quick fix"?

CoenWarmer:
> Yes, I don't see any reason to keep it away from them - while the feature isn't GA I don't anticipate the controls to change. Some things will need polishing in Elasticsearch (query for unmapped fields, etc...) but the backend plumbing is good and should be exposed through TF ASAP.
> 
> @CoenWarmer shall I open another issue for this or is it a "quick fix"?

@3kt the TF docs only publish experimental Kibana features (i.e., Streams) when the generate-docs function inside this repo is run with a particular env variable set.


So there's two ways we can go about this:
1. We (the Streams team) mark Streams as GA -> the next time TF docs are generated, Streams are included

Or:

2. We ask the owners of terraform-provider-elasticstack if they are willing to run the docs with the env variable so experimental features are included -> the next time TF docs are generated Streams are included as an experimental feature.


Which way do you want to go with this? It then also follows where the issue should be opened and who should follow up.

3kt:
GA is a separate issue https://github.com/elastic/ingest-dev/issues/8015 - I'm not comfortable with "just GA-ing" wired streams at this stage.

Let's go with option 2, who do we need to ~gently ask~ shamelessly bribe to make it happen @CoenWarmer ? ;)
Happy to support any way I can!

CoenWarmer:
> GA is a separate issue https://github.com/elastic/ingest-dev/issues/8015 - I'm not comfortable with "just GA-ing" wired streams at this stage.
> 
> 
> 
> Let's go with option 2, who do we need to ~gently ask~ shamelessly bribe to make it happen @CoenWarmer ? ;)
> 
> Happy to support any way I can!

@tobio, is this something you could help us with? 

Cc @3kt

tobio:
That's not a change worth making, we've got two options for resources at the moment:

1. Publicly available resources. They're documented and available by default in the provider release.
2. Experimental resources (where streams sits now). They're undocumented and only available behind an environment variable. 

Including Streams in the public docs, but still only available behind an environment variable is a foot gun not worth pulling. 

Dashboards sits in a similar space (Tech Preview in the product, but available in the provider). The [docs](https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/kibana_dashboard) call this out. 

IMO the two options really are:
1. Going down the same path as dashboards, calling out 'wired streams' as tech preview or whatever framing makes sense there. FWIW wired streams are already fully documented in the [Kibana API docs](https://www.elastic.co/docs/api/doc/kibana/v9/operation/operation-put-streams-name) so this seems consistent.
2. Hiding the wired stream section of the schema behind the experimental flag. Wired streams wouldn't be documented and would only be available behind an environment variable. Should be quite simple, but not something we've done before.

CoenWarmer:
> That's not a change worth making, we've got two options for resources at the moment:
> 
> 1. Publicly available resources. They're documented and available by default in the provider release.
> 2. Experimental resources (where streams sits now). They're undocumented and only available behind an environment variable.
> 
> Including Streams in the public docs, but still only available behind an environment variable is a foot gun not worth pulling.
> 
> Dashboards sits in a similar space (Tech Preview in the product, but available in the provider). The [docs](https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/kibana_dashboard) call this out.
> 
> IMO the two options really are:
> 
> 1. Going down the same path as dashboards, calling out 'wired streams' as tech preview or whatever framing makes sense there. FWIW wired streams are already fully documented in the [Kibana API docs](https://www.elastic.co/docs/api/doc/kibana/v9/operation/operation-put-streams-name) so this seems consistent.
> 2. Hiding the wired stream section of the schema behind the experimental flag. Wired streams wouldn't be documented and would only be available behind an environment variable. Should be quite simple, but not something we've done before.

@tobio @3kt Makes sense. We've discussed and think following Dashboards is the right way forward. 

Created: https://github.com/elastic/terraform-provider-elasticstack/pull/3782

Issue #1150, opened 2025-05-28
elasticstack_fleet_integration_policy's streams_json param should not be sensitive by default.


dotspec:
**Is your feature request related to a problem? Please describe.**

Using the `elasticstack_fleet_integration_policy` the `streams_json` parameter is sensitive by default. This makes reviewing terraform plans as part of a PR impossible as the reviewers (well anyone really) cannot see the output.  In addition if you hit an error on apply you have no way to review the output because, again, sensitive by default.  And in addition to that using terraform's `nonsensitive` function doesn't seem to have any effect, i.e. I still can't read the plan.

**Describe the resource you would like to have implemented.**

N/A I just want to be able to see what my plans are going to do ahead of applying them.

**Describe the solution you'd like**

See above

**Describe alternatives you've considered**

A few alternatives but non terraform related and they're all bad.

**Additional context**

tobio:
> by default

This is a technical limitation within the Terraform API. An attribute is either sensitive or not, there's no finer grained mechanism available here. 

Since `streams_json` commonly includes sensitive values, not marking it as sensitive isn't a realistic option. As part of https://github.com/elastic/terraform-provider-elasticstack/issues/889 we'll look at supporting an individual stream as an isolated block attribute, which would mean that only specific sensitive fields would be marked as sensitive, rather than the entire JSON string. 

For example:

```hcl
elasticstack_fleet_integration_policy {
  stream {
    id = "foo"
    ...
  }

  stream {
    id = "bar"
  }
}
```

Issue #617, opened 2024-04-29
Improve fleet integration management 


Kushmaro:
**Situation 1: Mandatory Definition of All Integration Inputs**

Customer is encountering an "issue" where they must explicitly define all integration inputs within their Fleet Terraform configurations, even for those that are not enabled. This is necessary to prevent Terraform from showing diffs between plans. For example, in the AWS integration, which offers 50 different inputs, they only enable 4. Despite this, they need to define every input, importing their default values, to avoid plan discrepancies.

Here's an example of the issue with a disabled input still needing to be defined:

```hcl
resource "elasticstack_fleet_integration_policy" "server-system" {
  name              = "system"
  namespace         = "default"
  description       = "example"
  agent_policy_id   = elasticstack_fleet_agent_policy.server_policy.id
  integration_name  = elasticstack_fleet_integration.this["system"].name
  integration_version = elasticstack_fleet_integration.this["system"].version

  input {
    enabled       = true
    input_id      = "system-logfile"
    streams_json  = jsonencode(jsondecode(file("integration_configs/server/system/system-logfile.json")))
  }
  ...
  input {
    enabled       = false
    input_id      = "system-httpjson"
    streams_json  = jsonencode(jsondecode(file("integration_configs/server/system/system-httpjson.json")))
  }
}
```

If customer remove the last input and its associated streams, the plan continuously shows a diff suggesting the removal of the undefined input. This requires customer to "backport" the default streams for all inputs, even disabled ones, and to update them with each integration version change.

psypuff:
This seems to also affect multiple streams within a single input (e.g. the [Prometheus integration](https://www.elastic.co/docs/current/integrations/prometheus)).
Would love to omit the disabled-by-default streams from the `streams_json` argument. Not sure how easy would that be given this is a json string and probably not part of the schema. Maybe worth having streams as blocks just for the `enabled` argument and the rest of their config could be kept as json.

PR #1482, opened 2025-11-24
Move the input blocks to an inputs map

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/617

By itself this is just a bit of a painful breaking change, I'm splitting this overall change up to make things more easily reviewed though. The end goal here is:
* Reduce the scope of sensitive attributes. https://github.com/elastic/terraform-provider-elasticstack/issues/1150 highlights some of the pain here, where `streams_json` is a big blob which is treated as a single sensitive string. Decomposing this into separate stream attributes allows practitioners to better understand the changes being applied. 
* Better consistency between API and TF schemas. Inputs and streams are maps in the API, it makes much more sense to treat them as maps in TF land as well.
* Only require what's actually required. Currently all inputs and streams must be fully defined, otherwise the provider breaks. Decomposing the streams JSON blob allows us to more easily merge user provided config with the integration defaults so users only need to specify vars/streams etc *if they're overriding a default value* 

All of that is coming in follow up PRs though. This PR is intended to only be a schema change. 

### Breaking change
* The `input` block has been removed, and replaced with an `inputs` map.
* `streams_json` has been replaced by a `streams` map attribute
* See the changes to the acceptance test definitions ([example](https://github.com/elastic/terraform-provider-elasticstack/compare/input-attribute?expand=1#diff-3346189f5ed24da90a529a5fa0d06be1745ee9013775c4b8e42a0e909155e5b6)) for examples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,72 @@
 ## [Unreleased]
 
+### Breaking changes
+
+#### `elasticstack_fleet_integration_policy` input block has changed to a map attribute. 
+
+The `input` block in the `elasticstack_fleet_integration_policy` resource has been restructured into the `inputs` map attribute. 
+This transition:
+* Allows the provider to implement semantic equality checking across all inputs within the integration policy. This change:
+  * Prevents several state consistency errors experienced whilst using this resource
+  * Allow practitioners to only define configuration for the inputs, streams, and variables that differ from the package defined defaults.
+* Reduces the scope of the large `streams_json` string. Instead allowing each stream to be defined as it's own object for Terraform drift checking. 
+
+Existing usage of the `input` block must be migrated to the attribute syntax. Some [examples](https://github.com/elastic/terraform-provider-elasticstack/compare/input-attribute?expand=1#diff-3346189f5ed24da90a529a5fa0d06be1745ee9013775c4b8e42a0e909155e5b6) of this migration can be seen in the changes to the provider automated tests. As a step-by-step guide however:
+
+1. `input` blocks are merged together into a single `inputs` attribute
+2. The `input_id` attribute is removed, and instead used as the map key when defining an input
+3. `streams_json` is removed, with the contents becoming a `streams` map attribute
+
+Combined, this looks like:
+
+```hcl
+input {
+  input_id = "tcp-tcp"
+  enabled  = false
+  streams_json = jsonencode({
+    "tcp.generic" : {
+      "enabled" : false
+      "vars" : {
+        "listen_address" : "localhost"
+        "listen_port" : 8085
+        "data_stream.dataset" : "tcp.generic"
+        "tags" : []
+        "syslog_options" : "field: message"
+        "ssl" : ""
+        "custom" : ""
+      }
+    }
+  })
+}
+```
+
+becoming
+
+```hcl
+inputs = {
+  "tcp-tcp" = {
+    enabled = false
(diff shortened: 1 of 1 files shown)


docs/resources/fleet_integration_policy.md
dimuon: The example should be updated.

internal/fleet/integration_policy/models.go
```diff
@@ -107,20 +110,19 @@ func (model *integrationPolicyModel) populateFromAPI(ctx context.Context, data *
...
 
 	return diags
 }
 
-func (model *integrationPolicyModel) populateInputFromAPI(ctx context.Context, inputs map[string]kbapi.PackagePolicyInput, diags *diag.Diagnostics) {
+func (model *integrationPolicyModel) populateInputsFromAPI(ctx context.Context, inputs map[string]kbapi.PackagePolicyInput, diags *diag.Diagnostics) {
 	// Handle input population based on context:
 	// 1. If model.Input is unknown: we're importing or reading fresh state → populate from API
 	// 2. If model.Input is known and null/empty: user explicitly didn't configure inputs → don't populate (avoid inconsistent state)
 	// 3. If model.Input is known and has values: user configured inputs → populate from API
```
dimuon:
```suggestion
	// 1. If model.Inputs is unknown: we're importing or reading fresh state → populate from API
	// 2. If model.Inputs is known and null/empty: user explicitly didn't configure inputs → don't populate (avoid inconsistent state)
	// 3. If model.Inputs is known and has values: user configured inputs → populate from API
```

internal/fleet/integration_policy/testdata/TestAccResourceIntegrationPolicySecretsFromSDK/current/integration_policy.tf
```diff
@@ -0,0 +1,115 @@
...
+  description     = "IntegrationPolicyTest Agent Policy"
+  monitor_logs    = true
+  monitor_metrics = true
+  skip_destroy    = false
+}
+
+data "elasticstack_fleet_enrollment_tokens" "test_policy" {
+  policy_id = elasticstack_fleet_agent_policy.test_policy.policy_id
+}
+
```
dimuon: I guess it can be removed:

```suggestion
```

dimuon: I guess the same change can be appied to internal/fleet/integration_policy/testdata/TestAccResourceIntegrationPolicySecretsFromSDK/legacy/integration_policy.tf

dimuon (approved):

docs/resources/fleet_integration_policy.md
tobio: https://github.com/elastic/terraform-provider-elasticstack/pull/1482/commits/ac544c376c724b952da7462295d0ae6bb4527e54

internal/fleet/integration_policy/testdata/TestAccResourceIntegrationPolicySecretsFromSDK/current/integration_policy.tf
tobio: https://github.com/elastic/terraform-provider-elasticstack/pull/1482/commits/60ba6b309ff098eabb633b9af89ba029a1951fc6

Issue #1332, opened 2025-09-24
[Feature]Support Elastic Security Exceptions


approksiu:
**Is your feature request related to a problem? Please describe.**
Support managing Elastic security exceptions with Terraform, without this feature users are limited to managing only rules-connected exceptions.

**Describe the resource you would like to have implemented.**
https://www.elastic.co/docs/api/doc/kibana/group/endpoint-security-exceptions-api

**Describe the solution you'd like**
A resource to manage security exceptions is added, or the existing detection rules resource is extended - please advise on pros and cons.

PR #1489, opened 2025-11-25
Add security list resource

nick-benoit:
Adds Security List resource [Docs](https://www.elastic.co/docs/api/doc/kibana/group/endpoint-security-lists-api).

This is supporting work for [Security Exceptions](https://github.com/elastic/terraform-provider-elasticstack/issues/1332). 

On it's own this resource is not tremendously useful, but will quickly be followed by PRs for the following:
* [Security List Items ](https://www.elastic.co/docs/api/doc/kibana/operation/operation-createlistitem)
* [Security Exceptions](https://www.elastic.co/docs/api/doc/kibana/group/endpoint-security-exceptions-api)

### DIFF
--- a/internal/kibana/security_list/models.go
+++ b/internal/kibana/security_list/models.go
@@ -133,23 +133,11 @@ func (m *SecurityListModel) fromAPI(ctx context.Context, apiList *kbapi.Security
 	m.UpdatedBy = types.StringValue(apiList.UpdatedBy)
 
 	// Set optional _version field
-	if apiList.UnderscoreVersion != nil {
-		m.VersionID = utils.StringishPointerValue(apiList.UnderscoreVersion)
-	} else {
-		m.VersionID = types.StringNull()
-	}
+	m.VersionID = utils.StringishPointerValue(apiList.UnderscoreVersion)
 
-	if apiList.Deserializer != nil {
-		m.Deserializer = utils.StringishPointerValue(apiList.Deserializer)
-	} else {
-		m.Deserializer = types.StringNull()
-	}
+	m.Deserializer = utils.StringishPointerValue(apiList.Deserializer)
 
-	if apiList.Serializer != nil {
-		m.Serializer = utils.StringishPointerValue(apiList.Serializer)
-	} else {
-		m.Serializer = types.StringNull()
-	}
+	m.Serializer = utils.StringishPointerValue(apiList.Serializer)
 
 	if apiList.Meta != nil {
 		metaBytes, err := json.Marshal(apiList.Meta)


internal/kibana/security_list/schema.go
```diff
@@ -0,0 +1,122 @@
...
+				MarkdownDescription: "Determines how uploaded list item values are parsed. By default, list items are parsed using named regex groups based on the type.",
+				Optional:            true,
+				Computed:            true,
+			},
+			"meta": schema.StringAttribute{
+				MarkdownDescription: "Placeholder for metadata about the value list as JSON string.",
+				Optional:            true,
+			},
+			"version": schema.Int64Attribute{
+				MarkdownDescription: "The document version number.",
```
nick-benoit, line 89: From the OAS "The version id, normally returned by the API when the document is retrieved. Use it ensure updates are done against the latest version." Add this instead

CODING_STANDARDS.md
```diff
@@ -36,6 +36,7 @@ This document outlines the coding standards and conventions used in the terrafor
 - Prefer using existing util functions over longer form, duplicated code:
   - `utils.IsKnown(val)` instead of `!val.IsNull() && !val.IsUnknown()`
   - `utils.ListTypeAs` instead of `val.ElementsAs` or similar for other collection types
+- The final state for a resource should be derived from a read request following a mutative request (eg create or update). We should not use the response from a mutative request to build the final resource state.
```
nick-benoit, line 39: Copilot kept suggesting that I use the create / update response instead of doing a secondary read. Adding this seems to help with that.

internal/clients/kibana_oapi/security_lists.go
```diff
@@ -0,0 +1,154 @@
...
+// GetList reads a security list from the API by ID
+func GetList(ctx context.Context, client *Client, spaceId string, params *kbapi.ReadListParams) (*kbapi.ReadListResponse, diag.Diagnostics) {
+	resp, err := client.API.ReadListWithResponse(ctx, kbapi.SpaceId(spaceId), params)
+	if err != nil {
+		return nil, diagutil.FrameworkDiagFromError(err)
+	}
+
+	switch resp.StatusCode() {
+	case http.StatusOK:
+		return resp, nil
```
tobio, line 37:
```suggestion
		return resp.JSON200, nil
```

Does it make sense to return the parsed response models from these helpers?

internal/kibana/security_list/resource-description.md
```diff
@@ -0,0 +1,27 @@
...
+  type        = "keyword"
+}
+```
+
+## Notes
+
+- Security lists define the type of data they can contain via the `type` attribute
+- Once created, the `type` of a list cannot be changed
+- Lists can be referenced by exception items to create more sophisticated matching rules
+- The `list_id` is auto-generated if not provided
```
tobio:
```suggestion
## Notes

- Security lists define the type of data they can contain via the `type` attribute
- Once created, the `type` of a list cannot be changed
- Lists can be referenced by exception items to create more sophisticated matching rules
- The `list_id` is auto-generated if not provided
```

The docs generator will include the examples automatically, we don't need to duplicate it here. 

We probably need to add some more details to the copilot instructions to stop it generating these huge descriptions.

internal/kibana/security_list/resource-description.md
```diff
@@ -0,0 +1,27 @@
+Manages Kibana security lists (also known as value lists). Security lists are used by exception items to define sets of values for matching or excluding in security rules.
```
tobio, line 1: Can we add a link to some security list docs here too.

internal/kibana/security_list/schema.go
tobio: Was this actually resolved? What's the difference between `version` and `version_id`?

internal/kibana/security_list/schema.go
```diff
@@ -0,0 +1,121 @@
...
+				Optional:            true,
+				Computed:            true,
+			},
+			"serializer": schema.StringAttribute{
+				MarkdownDescription: "Determines how uploaded list item values are parsed. By default, list items are parsed using named regex groups based on the type.",
+				Optional:            true,
+				Computed:            true,
+			},
+			"meta": schema.StringAttribute{
+				MarkdownDescription: "Placeholder for metadata about the value list as JSON string.",
```
tobio, line 84:
```suggestion
				MarkdownDescription: "Placeholder for metadata about the value list as JSON string.",
				CustomType: jsontypes.NormalizedType,
```

internal/kibana/security_list/models.go
```diff
@@ -0,0 +1,156 @@
...
+	m.CreatedBy = types.StringValue(apiList.CreatedBy)
+	m.UpdatedAt = types.StringValue(apiList.UpdatedAt.Format(time.RFC3339))
+	m.UpdatedBy = types.StringValue(apiList.UpdatedBy)
+
+	// Set optional _version field
+	if apiList.UnderscoreVersion != nil {
+		m.VersionID = types.StringValue(string(*apiList.UnderscoreVersion))
+	} else {
+		m.VersionID = types.StringNull()
+	}
```
tobio: We have this pattern a lot where the API returns a custom string type and so we can't use `types.StringPointerValue`. 

Want to create a new `typesutil` package with something like:

```go
func StringishPointerValue[T ~string](ptr *T) types.String {
	if ptr == nil {
		return types.StringNull()
	}
	return types.StringValue(string(*ptr))
}
```

so this code can just be

```suggestion
	m.VersionID = typesutil.StringishPointerValue(apiList.UnderscoreVersion)
```

internal/kibana/security_list/update.go
```diff
@@ -0,0 +1,75 @@
...
+	readResp, diags := kibana_oapi.GetList(ctx, client, spaceID, readParams)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if readResp == nil || readResp.JSON200 == nil {
+		resp.State.RemoveResource(ctx)
+		return
+	}
```
tobio, line 66: IMO we should likely return an error here, we just updated the rule and should expect it to exist.

internal/kibana/security_list/create.go
```diff
@@ -0,0 +1,69 @@
...
+	readResp, diags := kibana_oapi.GetList(ctx, client, spaceID, readParams)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if readResp == nil || readResp.JSON200 == nil {
+		resp.State.RemoveResource(ctx)
+		return
+	}
```
tobio, line 60: Similarly, we just successfully created the list and should expect to be able to read it back.

internal/kibana/security_list/resource-description.md
nick-benoit: https://github.com/elastic/terraform-provider-elasticstack/pull/1489/commits/924f1d766b906b39bad2cdd9f8b9cba53fb5741b

internal/kibana/security_list/resource-description.md
nick-benoit: https://github.com/elastic/terraform-provider-elasticstack/pull/1489/commits/924f1d766b906b39bad2cdd9f8b9cba53fb5741b

internal/kibana/security_list/schema.go
nick-benoit: https://github.com/elastic/terraform-provider-elasticstack/pull/1489/commits/09091597ad177fbe8c92c7a45b434f514ab1330b

internal/kibana/security_list/models.go
nick-benoit: It seemed like these utilities fit pretty well into the existing `utils/utils.go` file. Happy to add a dedicated package if you prefer that though.

https://github.com/elastic/terraform-provider-elasticstack/pull/1489/commits/52c50ba915d3c6dc1f8aa212c19080223836393e

internal/kibana/security_list/update.go
nick-benoit: https://github.com/elastic/terraform-provider-elasticstack/pull/1489/commits/d89ae58b6f857e77cc083c62f5fa593a94683c17

internal/kibana/security_list/create.go
nick-benoit: https://github.com/elastic/terraform-provider-elasticstack/pull/1489/commits/d89ae58b6f857e77cc083c62f5fa593a94683c17

internal/clients/kibana_oapi/security_lists.go
nick-benoit: I think it does. I may opt to do that refactor in the list item PR though for the sake of easily pulling these changes into that dependent PR.

internal/kibana/security_list/read.go
```diff
@@ -0,0 +1,61 @@
...
+		resp.Diagnostics.AddError("Failed to get Kibana client", err.Error())
+		return
+	}
+
+	// Parse composite ID to get space_id and list_id
+	spaceID := state.SpaceID.ValueString()
+	listID := state.ListID.ValueString()
+
+	// Try to parse as composite ID from state.ID
+	if compId, diags := clients.CompositeIdFromStrFw(state.ID.ValueString()); !diags.HasError() {
```
nick-benoit, line 31: @tobio FYI I realized while adding an import test that we were not using the internal composite ID properly. I just pushed that up in this commit: https://github.com/elastic/terraform-provider-elasticstack/pull/1489/commits/20b443dffba4eee6c420b70701b59d038fbd058a

internal/kibana/security_list/models.go
tobio: Happy to leave this for a follow up PR. A catch-all utils package is [discouraged](https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common) by the Golang gods, so I was thinking we could start moving these TF type specific helpers to a dedicated `typeutils` package.

internal/kibana/security_list/models.go
```diff
@@ -0,0 +1,166 @@
...
+		m.VersionID = utils.StringishPointerValue(apiList.UnderscoreVersion)
+	} else {
+		m.VersionID = types.StringNull()
+	}
+
+	if apiList.Deserializer != nil {
+		m.Deserializer = utils.StringishPointerValue(apiList.Deserializer)
+	} else {
+		m.Deserializer = types.StringNull()
+	}
```
tobio: We don't need the if/else for these with this new helper 

```suggestion
	m.Deserializer = utils.StringishPointerValue(apiList.Deserializer)
```

nick-benoit: Ahh good call: https://github.com/elastic/terraform-provider-elasticstack/pull/1489/commits/1907787391ea1cea2990ca40f16ff18fe78a05b2

internal/kibana/security_list/models.go
nick-benoit: I'll make a note and add that to the security list item PR 👍

PR #1496, opened 2025-11-27
Add security exception item

nick-benoit:
Adds security exception item resource ([docs](https://www.elastic.co/docs/api/doc/kibana/operation/operation-createexceptionlistitem))

Rel: https://github.com/elastic/terraform-provider-elasticstack/issues/1332

### DIFF
--- a/internal/kibana/security_exception_item/models.go
+++ b/internal/kibana/security_exception_item/models.go
@@ -683,13 +683,144 @@ func getCommentAttrTypes() map[string]attr.Type {
 	}
 }
 
+// convertEntriesToAPIWithDiags converts entries and handles diagnostics
+func (m *ExceptionItemModel) convertEntriesToAPIWithDiags(ctx context.Context, diags *diag.Diagnostics) kbapi.SecurityExceptionsAPIExceptionListItemEntryArray {
+	entries, d := convertEntriesToAPI(ctx, m.Entries)
+	diags.Append(d...)
+	return entries
+}
+
+// CommonExceptionItemProps holds pointers to common fields across create/update requests
+type CommonExceptionItemProps struct {
+	NamespaceType *kbapi.SecurityExceptionsAPIExceptionNamespaceType
+	OsTypes       *[]kbapi.SecurityExceptionsAPIExceptionListOsType
+	Tags          *kbapi.SecurityExceptionsAPIExceptionListItemTags
+	Meta          *kbapi.SecurityExceptionsAPIExceptionListItemMeta
+	ExpireTime    *kbapi.SecurityExceptionsAPIExceptionListItemExpireTime
+}
+
+// setCommonProps sets common fields across create and update requests
+func (m *ExceptionItemModel) setCommonProps(
+	ctx context.Context,
+	props *CommonExceptionItemProps,
+	diags *diag.Diagnostics,
+	client clients.MinVersionEnforceable,
+) {
+	// Set optional namespace_type
+	if utils.IsKnown(m.NamespaceType) {
+		nsType := kbapi.SecurityExceptionsAPIExceptionNamespaceType(m.NamespaceType.ValueString())
+		*props.NamespaceType = nsType
+	}
+
+	// Set optional os_types
+	if utils.IsKnown(m.OsTypes) {
+		osTypes := utils.SetTypeAs[kbapi.SecurityExceptionsAPIExceptionListOsType](ctx, m.OsTypes, path.Empty(), diags)
+		if diags.HasError() {
+			return
+		}
+		if len(osTypes) > 0 {
+			*props.OsTypes = osTypes
+		}
+	}
+
+	// Set optional tags
+	if utils.IsKnown(m.Tags) {
+		tags := utils.SetTypeAs[string](ctx, m.Tags, path.Empty(), diags)
+		if diags.HasError() {
+			return
+		}
+		if len(tags) > 0 {
+			tagsArray := kbapi.SecurityExceptionsAPIExceptionListItemTags(tags)
+			*props.Tags = tagsArray
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


internal/kibana/security_exception_item/schema.go
```diff
@@ -0,0 +1,251 @@
...
+							Computed:            true,
+						},
+						"comment": schema.StringAttribute{
+							MarkdownDescription: "The comment text.",
+							Required:            true,
+						},
+					},
+				},
+			},
+			"expire_time": schema.StringAttribute{
```
nick-benoit, line 223: It seems there is an issue where if a round second is provided then the go client doesn't send the milliseconds and generates a 400 from kibana

So eg `"2027-12-31T23:59:59.001Z"` works and `"2027-12-31T23:59:59.000Z"` does not. Still trying to find an elegant solution to this.

tobio: Golang will format times as [RFC3339 in JSON](https://pkg.go.dev/time#Time.MarshalJSON) which isn't strictly compatible with the API spec. I wonder if we should remove the `format: date-time` from `Security_Exceptions_API_ExceptionListItemExpireTime` so we can take control of the string formatting ourselves?

nick-benoit: Yeah I think that is probably the best bet

internal/kibana/security_exception_item/models.go
```diff
@@ -0,0 +1,976 @@
...
+		entry.Values = nestedEntry.Values
+	case "exists":
+		nestedEntry := convertNestedExistsFromMap()
+		entry.Value = nestedEntry.Value
+		entry.Values = nestedEntry.Values
+	}
+
+	return entry, diags
+}
+
```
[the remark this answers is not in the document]
nick-benoit: https://github.com/elastic/terraform-provider-elasticstack/pull/1496/commits/2ae421df9bf86d8512fbb3e574e8fe6a69590d2b

internal/kibana/security_exception_item/schema.go
nick-benoit: I added that here: https://github.com/elastic/terraform-provider-elasticstack/pull/1496/commits/4e4b02d2d7d0cc67e525ee10662f7b8d97623e96

I added custom serialization that always includes the millis. It seems that since this is a subset of RFC3339 we can still use that for deserialization.

internal/kibana/security_exception_item/models.go
```diff
@@ -0,0 +1,994 @@
...
+		if diags.HasError() {
+			return nil, diags
+		}
+		if len(osTypes) > 0 {
+			osTypesArray := make(kbapi.SecurityExceptionsAPIExceptionListItemOsTypeArray, len(osTypes))
+			for i, osType := range osTypes {
+				osTypesArray[i] = kbapi.SecurityExceptionsAPIExceptionListOsType(osType)
+			}
+			req.OsTypes = &osTypesArray
+		}
```
tobio:
```suggestion
		osTypes := utils.SetTypeAs[kbapi.SecurityExceptionsAPIExceptionListOsType](ctx, m.OsTypes, path.Empty(), &diags)
		if diags.HasError() {
			return nil, diags
		}

		if len(osTypes) > 0 {
			req.OsTypes = &osTypes
		}
```

internal/kibana/security_exception_item/models.go
```diff
@@ -0,0 +1,994 @@
...
```
tobio:
```suggestion
		osTypes := utils.SetTypeAs[kbapi.SecurityExceptionsAPIExceptionListOsType](ctx, m.OsTypes, path.Empty(), &diags)
		if diags.HasError() {
			return nil, diags
		}

		if len(osTypes) > 0 {
			req.OsTypes = &osTypes
		}
```

internal/kibana/security_exception_item/models.go
```diff
@@ -0,0 +1,994 @@
...
+
+		expireTimeAPI := kbapi.SecurityExceptionsAPIExceptionListItemExpireTime(expireTime.Format("2006-01-02T15:04:05.000Z"))
+		req.ExpireTime = &expireTimeAPI
+	}
+
+	return req, diags
+}
+
+// toUpdateRequest converts the Terraform model to API update request
+func (m *ExceptionItemModel) toUpdateRequest(ctx context.Context, resourceId string, client clients.MinVersionEnforceable) (*kbapi.UpdateExceptionListItemJSONRequestBody, diag.Diagnostics) {
```
tobio, line 889: There's a lot of duplication between `toCreate` and `toUpdate`. I wonder if we can just build the update request from the create request, or reuse some of the common code.

internal/kibana/security_exception_item/models.go
nick-benoit: https://github.com/elastic/terraform-provider-elasticstack/pull/1496/commits/f7fd99e891b9030fa3f94d4a34b8eab353a37d85

internal/kibana/security_exception_item/models.go
nick-benoit: https://github.com/elastic/terraform-provider-elasticstack/pull/1496/commits/f7fd99e891b9030fa3f94d4a34b8eab353a37d85

internal/kibana/security_exception_item/models.go
nick-benoit: What do you think of this approach? https://github.com/elastic/terraform-provider-elasticstack/pull/1496/commits/a4da71b6f082ad86a35e5ac26b21e362befb8e4d

This is kind of similar to what we did in security detection rules

tobio:
It looks like the acceptance tests have become pretty flaky with these exception list resources?

Issue #1488, opened 2025-11-25
[Bug] elasticstack_kibana_security_detection_rule validation fails when index comes from for_each


girtsLv:
**Describe the bug**
The `elasticstack_kibana_security_detection_rule` resource's schema validation incorrectly reports that `index` (or `data_view_id`) is not set when the `index` value comes from a `for_each` loop.

The error:
```
Error: Invalid Configuration

  with elasticstack_kibana_security_detection_rule.rules_eck,
  on rules_eck.tf line 1, in resource "elasticstack_kibana_security_detection_rule" "rules_eck":
   1: resource "elasticstack_kibana_security_detection_rule" "rules_eck" {

One of 'index' or 'data_view_id' must be set.
```

**To Reproduce**
1. Create a YAML file with rule definition including `index`:

```yaml
name: Example rule
rule_id: example-rule
type: query
index:
  - logs-*
query: event.code: 1116
```

2. Create Terraform configuration that reads YAML files and creates resources:

```hcl
locals {
  rules_dir   = "./rules_eck"
  rules_files = fileset(local.rules_dir, "*.yml")
  rules = {
    for f in local.rules_files :
    f => yamldecode(file("${local.rules_dir}/${f}"))
  }
}

resource "elasticstack_kibana_security_detection_rule" "rules_eck" {
  for_each = local.rules

  name        = each.value.name
  description = each.value.description
  type        = each.value.type
  rule_id     = each.value.rule_id
  index       = each.value.index
  
  # this works
  # index       = ["logs-*"]
}
```

3. Run `terraform validate`

**Expected behavior**
Validation should pass.
Verification shows the value exists: `terraform console` confirms `local.rules["test_rule_eck.yml"].index` contains the expected list

**Debug output**
n/a

**Screenshots**
n/a

**Versions (please complete the following information):**
- **Provider**: elasticstack (elastic/elasticstack)
- **Provider Version**: 0.12.2
- **OS**: Linux (WSL2)

**Additional context**
n/a

PR #1499, opened 2025-11-30
Allow index or data_view_id to be unknown

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1488

Updates the validation to fail only if both `index` and `data_view_id` are null, allowing at least one of them to be unknown. I'm not failing validation if they're both unknown since the `unknown` value may become null once known.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ alias = [
 
 ### Changes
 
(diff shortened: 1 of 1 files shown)


nick-benoit (approved):
🚀

Issue #601, opened 2024-04-05
[Feature] [terraform] Elastic Defend integration configuration


3kt:
**Is your feature request related to a problem? Please describe.**
Using `elasticstack_fleet_integration_policy` works well for standard TCP / logs integration, as showcased in the [documentation](https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/fleet_integration_policy), but for the Elastic [Defend integration](https://www.elastic.co/guide/en/security/current/install-endpoint.html) no custom configuration can be applied because of its structure. 

**Describe the resource you would like to have implemented.**
We'd need some extension of the capabilities of the `/api/fleet/package_policies/[integration ID]` endpoint. We basically need an extension from the capabilities of https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/fleet/integration_policy_resource.go to support customization for Elastic Defend integration.

**Describe the solution you'd like**
I'd like to be able to customize the Elastic Defend integration with the `elasticstack_fleet_integration_policy` resource, the same way the UI offers:
<img width="392" alt="image" src="https://github.com/elastic/terraform-provider-elasticstack/assets/12101744/e9dd8b07-c413-4831-b732-65b102924254">

**Describe alternatives you've considered**
There is currently no workaround with pure Terraform, as the API endpoint are simply not instrumented in the provider.

**Additional context**
Looking at what Kibana does when saving the integration through the UI, the custom settings of the Defend Integration look like this:

```
{
  "version": "WzU0NTAsMV0=",
  "name": "Defend integration",
  "namespace": "default",
  "description": "A sample integration policy",
  "package": {
    "name": "endpoint",
    "title": "Elastic Defend",
    "version": "8.13.0"
  },
  "enabled": true,
  "policy_id": "85ea0949-8d7b-4da0-bf2c-bbff166bbb72",
  "vars": {},
  "inputs": [
    {
      "type": "endpoint",
      "enabled": true,
      "config": {
        "integration_config": {},
        "artifact_manifest": {[some stuff]},
        "policy": {[some other stuff]}
      },
      "streams": []
    }
  ]
}
```

We don't currently have support for `artifact_manifest` and `policy`, but only the `streams` bit which is unused by the Defend integration.

nimarezainia:
@andrewkroh would you be able to help route this request. I believe your team developed the terraform provider. TIA

cc: @cmacknz

andrewkroh:
[...]

PR #1500, opened 2025-12-02
Support both mapped and typed input schemas in Fleet integration polies

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/601

Integration policies support two body formats, a map (`{"input-id": {...}}`) and a list (`[{"type":"endpoint"...}]`). Only the latter form supports the `config` field used by both Endpoint and APM integrations (even though that form is apparently deprecated 🤷). 

Functionally, this PR should be a no-op. It adds support for the latter input form in the API client, and updates the integration policy resource to match the new generated API client. It doesn't make any resource schema changes, and the behaviour should be identical.

### DIFF
--- a/generated/kbapi/transform_schema.go
+++ b/generated/kbapi/transform_schema.go
@@ -17,6 +17,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/google/go-cmp/cmp"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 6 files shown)


tobio:
There's some serious issues adding support for the endpoint integration using this resource. There's still a good chance we'll want to add this support in the API but ignore this PR for now.

Issue #836, opened 2024-10-09
[Bug] elasticstack_fleet_output produces inconsistent result after 0.11.8 upgrade


mag-mkorn:
**Describe the bug**
After upgrading the provider from 0.11.7 to 0.11.8 the provider plans to perform an in-place update to a pre-existing elasticstack_fleet_output-resource and ultimately fails.

```
Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # elasticstack_fleet_output.<redacted> will be updated in-place
  ~ resource "elasticstack_fleet_output" "cribl" {
        id                     = "<redacted>"
        name                   = "<redacted>"
        # (7 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
elasticstack_fleet_output.cribl: Modifying... [id=<redacted>]

Error: Provider produced inconsistent result after apply

When applying changes to elasticstack_fleet_output.<redacted>, provider
"provider[\"registry.terraform.io/elastic/elasticstack\"]" produced an
unexpected new value: .ca_trusted_fingerprint: was null, but now
cty.StringVal("").

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
Error: Terraform exited with code 1.
Error: Process completed with exit code 1.
```

**To Reproduce**
Resource config:

```
resource "elasticstack_fleet_output" "<redacted>" {
  name = "<redacted>"
  type = "elasticsearch"
  config_yaml = yamlencode({
    "ssl.verification_mode" : "full",
    "ssl.certificate_authorities" : <<EOF
-----BEGIN CERTIFICATE-----
<redacted>
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
<redacted>
-----END CERTIFICATE-----
EOF
  })
  default_integrations = true
  default_monitoring   = true
  hosts                = [var.<redacted>]
}
```

**Expected behavior**
Don't perform any changes without showing the reason for the change. Don't fail on comparing null values to empty strings.

**Debug output**
Currently not possible to provide, as TF is executed as CI Pipeline with not easily accessible secrets.

**Versions (please complete the following information):**
 - OS: [e.g. Linux]
 - Terraform Version tag "latest" on GHA hashicorp/setup-terraform@v2
 - Provider version 0.11.8
 - Elasticsearch Version 8.15.2

daemitus:
[...]

PR #1509, opened 2025-12-03
Migrate fleet output acceptance tests

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/836

Bringing this resource inline with the current expectations whilst I'm here.


nick-benoit (approved):
🚀

PR #1525, opened 2025-12-08
Add Security List Data Stream Resource

nick-benoit:
Changes inspired by using all security exception adjacent resources together

* Adds  simple resource (`elasticstack_kibana_security_list_data_streams`) to manage lifecycle of[ create list data streams](https://www.elastic.co/docs/api/doc/kibana/operation/operation-createlistindex)
* Adds more extensive test coverage for different list types within list type security exception item entries
  * Add tests for all types of security value list exception items including value list updates
  * Add tests for multiple exception items
  * Add tests for multiple entries in exception items of each type
  * Add tests for managing inner nested item entries
* Improve exception item schema validations

Rel: https://github.com/elastic/terraform-provider-elasticstack/issues/1332

### DIFF
--- a/internal/kibana/security_list_data_streams/resource-description.md
+++ b/internal/kibana/security_list_data_streams/resource-description.md
@@ -1,39 +1,3 @@
-# Resource: elasticstack_kibana_security_list_data_streams
-
 Creates `.lists` and `.items` data streams in the relevant Kibana space. These data streams are required before you can start using security lists and exceptions that reference value lists.
 
 Before you can start working with exceptions that use value lists, you must create the `.lists` and `.items` data streams for the relevant Kibana space. Once these data streams are created, your role needs privileges to manage rules.
-
-## Example Usage
-
-```terraform
-# Create list data streams in the default space
-resource "elasticstack_kibana_security_list_data_streams" "default" {
-}
-
-# Create list data streams in a custom space
-resource "elasticstack_kibana_security_list_data_streams" "custom" {
-  space_id = "my-space"
-}
-```
-
-## Argument Reference
-
-* `space_id` - (Optional) An identifier for the space. If `space_id` is not provided, the default space is used. Changing this forces a new resource to be created.
-
-## Attribute Reference
-
-In addition to the arguments above, the following attributes are exported:
-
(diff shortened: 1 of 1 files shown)


internal/kibana/security_list_data_streams/read.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	// Check if the data streams exist
+	listIndex, listItemIndex, diags := kibana_oapi.ReadListIndex(ctx, client, spaceID)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if !listIndex || !listItemIndex {
```
tobio, line 39:
```suggestion
	if !listIndex && !listItemIndex {
```

Should we only completely remove it if they're both missing?

internal/kibana/security_list_data_streams/resource-description.md
```diff
@@ -0,0 +1,39 @@
+# Resource: elasticstack_kibana_security_list_data_streams
```
tobio: Can we trim the extra template content from this one

tobio (commented):
Couple of small comments

internal/kibana/security_list_data_streams/read.go
nick-benoit: I think we may want OR. If only one of them is false I don't think we would ever call the create API which I think is how we would get back to the desired end state.

internal/kibana/security_list_data_streams/resource-description.md
nick-benoit: https://github.com/elastic/terraform-provider-elasticstack/pull/1525/commits/d8f4be821434295e3adac868371714746bc1b367

internal/kibana/security_list_data_streams/read.go
tobio: So another approach would be to have `list_index` and `list_item_index` set to `true` during planning (IIRC a default would work, otherwise via a plan modifier), which would then trigger an update if either of them were `false` in state. Which might be more informative for users, or give us more flexibility in the future. 

I'm not fussed though, this approach works and we can adjust it if required.

Issue #1523, opened 2025-12-08
[Bug] Provider injects internal __tf_provider_connector_type_id into config JSON causing perpetual plan drift and HTTP 400 rejection


StephanBenning:
**Describe the bug**  
The `elasticstack_kibana_action_connector` resource injects an internal `__tf_provider_connector_type_id` field into the `config` JSON after creation or UI changes. This causes perpetual Terraform plan drift (Terraform wants to remove the field) and fails with HTTP 400 if manually added to config.

**To Reproduce**  
1. TF configuration used:  
```terraform
resource "elasticstack_kibana_action_connector" "webhook" {
  connector_id      = "webhook-connector-id"
  connector_type_id = ".webhook"
  name              = "Hello webhook connector. (Terraform managed)"
  space_id          = "foo"

  config = jsonencode({
    authType = "webhook-authentication-basic"
    hasAuth  = true
    method   = "post"
    url      = "https://foo.com/webhook"
  })

  secrets = (sensitive value)
}
```
2. TF operations to execute to get the error:  
   - `terraform apply` (succeeds)  
   - Make manual UI change (e.g. update secret)  
   - `terraform plan` (shows drift)  
   - Or add `__tf_provider_connector_type_id = ".webhook"` to config and `terraform apply`  
3. See the error/drift in the output:  
**Plan drift:**  
```
~ resource "elasticstack_kibana_action_connector" "webhook" {
  ~ config = jsonencode(
    ~ {
      - __tf_provider_connector_type_id = ".webhook"
      # (other attributes unchanged)
    }
  )
}
```
**Apply error when manually adding the field:**  
```
Error: Unexpected status code from server: got HTTP 400

{"statusCode":400,"error":"Bad Request","message":"error validating action type config: [__tf_provider_connector_type_id]: definition for this key is missing"}
```

**Expected behavior**  
The provider should not inject `__tf_provider_connector_type_id` into user-visible `config` JSON. Internal metadata should be tracked separately (computed attribute) and ignored in state equality checks. Terraform plan should show "no changes" after UI secret updates.[1]

**Debug output**  
`TF_LOG=trace` shows provider reading Kibana connector and injecting `__tf_provider_connector_type_id` into state config during refresh. Field rejected by Kibana API when sent back. Sensitive data redacted.

**Screenshots**  
N/A

**Versions (please complete the following information):**  
- Terraform Version: 1.13.0
- Provider version: elastic/elasticstack >= 0.12.2
[...]

PR #1529, opened 2025-12-09
Add docs around the internal __tf_provider_connector_type_id property

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1523

There's not a lot we can do to entirely prevent this issue without a much bigger change here:
* The connector type is required to populate the correct defaults for a specific type. 
* A custom type doesn't have access to the value of any other sibling/parent attributes within the same resource. 
* We could strip the type in `ToTerraformValue` but that doesn't work since the type gets stripped out before `StringSemanticEquals` is called. 
* We could populate the type in a plan modifier rather than store it in state. That doesn't work since the result of a modified plan *must* equal either the value in config, or in state. Since that comparison happens on a plain `StringValue` the custom `StringSemanticEquals` isn't called, it's a byte for byte equality.  

Definitely happy to try out other ideas here, but I don't see another path forward right now.

### DIFF
--- a/docs/resources/kibana_action_connector.md
+++ b/docs/resources/kibana_action_connector.md
@@ -118,7 +118,7 @@ resource "elasticstack_kibana_action_connector" "webhook" {
 ### Optional
 
 - `config` (String) The configuration for the connector. Configuration properties vary depending on the connector type.
-				
+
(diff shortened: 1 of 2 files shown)


tobio:
> Could we rely on plan modifiers instead to achieve the same behavior? 

No. Plan modifiers *must* follow the core [data consistency rules](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#terraform-data-consistency-rules), so we can't change the contents of user provided `config` attribute at the point since it would be inconsistent. The consistency rules applied after plan modification don't consider semantic equality and so we can't ignore the injected `__tf_provider_connector_type_id`.

dimuon (approved):

Issue #650, opened 2024-05-31
[Feature]


Anri-Lombard:
**Is your feature request related to a problem? Please describe.**
Can't create Kibana Dashboards with provider.

**Describe the resource you would like to have implemented.**
Kibana Dashboards.

**Describe the solution you'd like**
Provisioning Kibana dashboards directly with terraform.

**Describe alternatives you've considered**
Doing it manually, which is the only way I'm aware of.

tobio:
> I was wondering when would be the earliest you think we can integrate a resource for the dashboard APIs?

@Kushmaro I'm going to spend some time with the current dashboards spec next week, so hopefully we can track alongside the Stack development and have something ready shortly after release.

andrew-moldovan:
@tobio @Kushmaro @teresaalvarezsoler I'm going to call this issue closed and we will do a 0.14 release asap. Any new API changes, bugs, etc, should be filed under new issues 🙇

PR #1530, opened 2025-12-09
Initial Kibana dashboard resource implementation

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/650

This isn't intended to be a full implementation of a dashboard resource, currently it will only manage the existence of a dashboard. 

This PR is intended to provide a starting point so that we can continue to track the development of the dashboard API with the aim to have provider support for dashboards released alongside, or closely after the initial release of the dashboard API.

### DIFF
--- a/internal/kibana/dashboard/models.go
+++ b/internal/kibana/dashboard/models.go
@@ -6,12 +6,12 @@ import (
 
 	"github.com/elastic/terraform-provider-elasticstack/generated/kbapi"
 	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
+	"github.com/elastic/terraform-provider-elasticstack/internal/diagutil"
 	"github.com/elastic/terraform-provider-elasticstack/internal/utils"
 	"github.com/hashicorp/terraform-plugin-framework-jsontypes/jsontypes"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/types"
-	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 )
 
 // dashboardModel is the top-level Terraform model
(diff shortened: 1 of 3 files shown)


examples/resources/elasticstack_kibana_dashboard/import.sh
```diff
@@ -0,0 +1,4 @@
+# Dashboard can be imported using the composite ID format: <space_id>/<dashboard_id>
+# For example, to import a dashboard with ID "my-dashboard-id" from the default space:
```
nick-benoit, line 2: IIUC we aren't setup to support these examples in the automated docs?

internal/kibana/dashboard/create.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// Create the dashboard
+	createResp, diags := kibana_oapi.CreateDashboard(ctx, kibanaClient, spaceID, apiReq)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Populate state from API response
+	if createResp.JSON200 != nil {
```
nick-benoit: Should we do a secondary read instead of using the create response here?

internal/kibana/dashboard/models.go
```diff
@@ -0,0 +1,418 @@
...
+			Tags            *[]string
+			TimeRange       kbapi.KbnEsQueryServerTimeRangeSchema
+			Title           string
+			Version         *float32
+		}
+		References []kbapi.KbnContentManagementUtilsReferenceSchema
+	}
+
+	// Use type switch to handle different response types
+	switch v := resp.(type) {
```
nick-benoit: This seems like it could quickly get unwieldy. I wonder if we should split each case into its own function? I also wonder if we do followup reads after create / update if we need to handle post / put responses here.

internal/kibana/dashboard/models.go
```diff
@@ -0,0 +1,418 @@
...
+	if data.Data.Description != nil {
+		m.Description = types.StringValue(*data.Data.Description)
+	} else {
+		m.Description = types.StringNull()
+	}
+
+	// Map time range
+	m.TimeFrom = types.StringValue(data.Data.TimeRange.From)
+	m.TimeTo = types.StringValue(data.Data.TimeRange.To)
+	// TODO: Dashboards
```
nick-benoit, line 67: 🚀

internal/kibana/dashboard/models.go
```diff
@@ -0,0 +1,418 @@
...
+	// } else {
+	// 	m.TimeRangeMode = types.StringNull()
+	// }
+
+	// Map refresh interval
+	m.RefreshIntervalPause = types.BoolValue(data.Data.RefreshInterval.Pause)
+	m.RefreshIntervalValue = types.Int64Value(int64(data.Data.RefreshInterval.Value))
+
+	// Map query
+	m.QueryLanguage = types.StringValue(data.Data.Query.Language)
```
nick-benoit, line 80: Does query language give us any hints about if the query is a string or json?  I suppose KQL and Lucene can both come back as either a string or JSON?

internal/kibana/dashboard/models.go
```diff
@@ -0,0 +1,418 @@
...
+	SyncTooltips    *bool `json:"syncTooltips,omitempty"`
+	UseMargins      *bool `json:"useMargins,omitempty"`
+}, diags *diag.Diagnostics) types.Object {
+	if options == nil {
+		return types.ObjectNull(getOptionsAttrTypes())
+	}
+
+	model := optionsModel{}
+
+	if options.HidePanelTitles != nil {
```
nick-benoit: Maybe this pattern should be adapted to a new utility in `typeutils`?

internal/kibana/dashboard/models.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+
+	// Set time range mode
+	if utils.IsKnown(m.TimeRangeMode) {
+		mode := kbapi.KbnEsQueryServerTimeRangeSchemaMode(m.TimeRangeMode.ValueString())
+		req.Data.TimeRange.Mode = &mode
+	}
+
+	// Set query text - Query is a union type with json.RawMessage
+	if utils.IsKnown(m.QueryText) {
```
nick-benoit: It would be nice to extract this into something shared between update / create

internal/kibana/dashboard/testdata/TestAccResourceDashboardInSpace/in_space/main.tf
```diff
@@ -0,0 +1,27 @@
...
+  description = "Test dashboard in custom space"
+
+  time_from = "now-15m"
+  time_to   = "now"
+
+  refresh_interval_pause = true
+  refresh_interval_value = 60000
+
+  query_language = "kuery"
+  query_text     = ""
```
nick-benoit, line 26: Would it be worth adding a json example as well?

internal/kibana/dashboard/models.go
tobio: > I also wonder if we do followup reads after create / update if we need to handle post / put responses here.

💯 removed this stuff.

internal/kibana/dashboard/models.go
tobio: https://github.com/elastic/terraform-provider-elasticstack/pull/1530/commits/e6cef258970af5def809b1c092b1521465ad4852

internal/kibana/dashboard/models.go
tobio: https://github.com/elastic/terraform-provider-elasticstack/pull/1530/commits/e6cef258970af5def809b1c092b1521465ad4852

internal/kibana/dashboard/create.go
tobio: https://github.com/elastic/terraform-provider-elasticstack/pull/1530/commits/e6cef258970af5def809b1c092b1521465ad4852

examples/resources/elasticstack_kibana_dashboard/import.sh
tobio: The standard template will pull in this script by default if the resource implements import.

internal/kibana/dashboard/models.go
tobio: Yep, I'll work on getting some understanding of the requirements from the dashboard folk. Added a TODO list to the original [issue](https://github.com/elastic/terraform-provider-elasticstack/issues/650) to track this stuff

tobio: IIUC KQL, Lucene, and ESQL queries would only come back as text, whilst Query DSL would be JSON

internal/kibana/dashboard/testdata/TestAccResourceDashboardInSpace/in_space/main.tf
tobio: Probably wrapped up with https://github.com/elastic/terraform-provider-elasticstack/pull/1530#discussion_r2605142473 at the moment. I can experiment with the API, but it seems easier to get some better docs from the API folk.

internal/kibana/dashboard/models.go
tobio: https://github.com/elastic/terraform-provider-elasticstack/pull/1530/commits/e9f65c58ab37c3fa65f77b6bfcc4159f1a4537d8

internal/kibana/dashboard/models.go
nick-benoit: Cool, yeah it would be nice to be able to base this on the type 👍

nick-benoit (approved):
🚀

Issue #1543, opened 2025-12-11
Inconsistent state errors with ML jobs


tobio:
This resource results in a provider error. 

```hcl
resource "elasticstack_elasticsearch_ml_anomaly_detection_job" "ml_test" {
  job_id = "ml_test3"
  description = null
  groups      = []

  analysis_config = {
    detectors = [
      {
        function             = "sum"
        field_name           = "bytes"
        detector_description = "Sum of bytes"
        by_field_name        = null
        over_field_name      = null
        partition_field_name = null
        use_null             = false
        exclude_frequent     = null
        custom_rules         = []
      }
    ]

    bucket_span = "15m"
    latency     = null
    period      = null
    influencers = []

    categorization_field_name = null
    categorization_filters    = []

    per_partition_categorization = {
      enabled      = false
      stop_on_warn = false
    }

    summary_count_field_name = null
    multivariate_by_fields   = false
    model_prune_window       = null
  }

  data_description = {
    time_field  = "timestamp"
    time_format = "epoch_ms"
  }

  analysis_limits = {
    model_memory_limit            = "11MB"
    categorization_examples_limit = null
  }

  model_plot_config = {
    enabled             = true
    annotations_enabled = true
    terms               = null
  }
  allow_lazy_open                           = false
  background_persist_interval               = null
  custom_settings                           = null
  daily_model_snapshot_retention_after_days = null
  model_snapshot_retention_days             = null
  renormalization_window_days               = null
  results_index_name  = "test-job1"
  results_retention_days = null
}
```

Error:
```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to
│ module.ml_job_config.elasticstack_elasticsearch_ml_anomaly_detection_job.ml_test,
│ provider "provider[\"registry.terraform.io/elastic/elasticstack\"]"
│ produced an unexpected new value: .description: was null, but now
│ cty.StringVal("").
│ 
│ This is a bug in the provider, which should be reported in the
│ provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to
│ module.ml_job_config.elasticstack_elasticsearch_ml_anomaly_detection_job.ml_test,
│ provider "provider[\"registry.terraform.io/elastic/elasticstack\"]"
│ produced an unexpected new value: .groups: was
│ cty.SetValEmpty(cty.String), but now null.
│ 
│ This is a bug in the provider, which should be reported in the
[...]

PR #1544, opened 2025-12-11
Fix up handling null and empty types

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1543

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,8 @@ inputs = {
 ```
 
 ### Changes
+
+- Fix handling empty types in `elasticstack_elasticsearch_ml_anomaly_detection_job` ([#1544](https://github.com/elastic/terraform-provider-elasticstack/pull/1544))
 - Move the `input` block to an `inputs` map in `elasticstack_fleet_integration_policy` ([#1482](https://github.com/elastic/terraform-provider-elasticstack/pull/1482))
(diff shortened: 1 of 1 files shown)


internal/elasticsearch/ml/anomaly_detection_job/models_tf.go
```diff
@@ -348,14 +331,13 @@ func (tfModel *AnomalyDetectionJobTFModel) fromAPIModel(ctx context.Context, api
 
 // Helper functions for schema attribute types
 // Conversion helper methods
-func (tfModel *AnomalyDetectionJobTFModel) convertAnalysisConfigFromAPI(ctx context.Context, apiConfig *AnalysisConfigAPIModel, diags *fwdiags.Diagnostics) types.Object {
+func (tfModel *AnomalyDetectionJobTFModel) convertAnalysisConfigFromAPI(ctx context.Context, apiConfig *AnalysisConfigAPIModel, diags *diag.Diagnostics) AnalysisConfigTFModel {
 	if apiConfig == nil || apiConfig.BucketSpan == "" {
-		return types.ObjectNull(getAnalysisConfigAttrTypes())
+		return tfModel.AnalysisConfig
```
dimuon: Shall we return an empty structure instead? Is it safe to return the current state?

Issue #1553, opened 2025-12-15
[Bug] Version 0.12.2 Space_ids is not valid


V3t3rn:
**Describe the bug**
n version 0.12.2 space_ids is not recognized as an input. When this is set it goes to the default space and errors out on the space_ids.  

**To Reproduce**
1. TF configuration used 
```
resource "elasticstack_fleet_agent_policy" "customer_logging_policy" {
depends_on = | elasticstack_kibana_space.customer_space ]
name="$(upper(var.customer_code)}-Logging"
namespace = "default"
description = "Logging
agent policy for customer ${var.customer_code}"
space ids var.customer_code
sys_monitoring = true
monitor_logs
= true
monitor metrics
= false
global_data.
tags
= {
"organization.id"
={
string_value
= var.customer code
"customer. code"
string_value
```
2. TF operations to execute to get the error `terraform apply`
3. See the error in the output '...'

**Expected behavior**
With space_ids it is expected to deploy the policy under the space identified. However space_ids is supposed to be set to work in that sace in general.

**Screenshots**
<img width="1550" height="1116" alt="Image" src="https://github.com/user-attachments/assets/47f33a2c-2db6-4510-b1d7-bf5ef7d21b77" />

**Versions (please complete the following information):**
 - OS: [e.g. Linux]
 - Terraform Version [0.12.2]
 - Provider version [e.g. v0.1.0]
 - Elasticsearch Version [9.2.2]

**Additional context**
Add any other context about the problem here. Links to specific affected code files and paths here are also extremely useful (if known).
[Slack Thread](https://elastic.slack.com/archives/C02L2QX6ZHB/p1765551277698919)

V3t3rn:
Apologizes the customer never responded with the setup they had. I have to wait for them to respond.

PR #1555, opened 2025-12-16
Validate space ids meet the API requirements

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/1553

Without this we get a fairly unhelpful 400 API response, it's a nicer UX providing immediate, and specific feedback.

### DIFF
--- a/internal/kibana/space.go
+++ b/internal/kibana/space.go
@@ -26,7 +26,7 @@ func ResourceSpace() *schema.Resource {
 			Type:         schema.TypeString,
 			Required:     true,
 			ForceNew:     true,
-			ValidateFunc: validation.StringMatch(regexp.MustCompile("[a-z0-9_-]+"), "must only contain lowercase letters, numbers, hyphens, and underscores"),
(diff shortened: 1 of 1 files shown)


dimuon (approved):

internal/kibana/space.go
```diff
@@ -22,10 +22,11 @@ func ResourceSpace() *schema.Resource {
 			Computed:    true,
 		},
 		"space_id": {
-			Description: "The space ID that is part of the Kibana URL when inside the space.",
-			Type:        schema.TypeString,
-			Required:    true,
-			ForceNew:    true,
+			Description:  "The space ID that is part of the Kibana URL when inside the space.",
```
dimuon, line 25: While the validation makes sense, I wanted to mention - just in case - that the original issue was raised against fleet policy space_ids.

tobio: Yea, I really only tagged the issue because I found this when investigating that issue.

Issue #1415, opened 2025-11-03
[Bug] Provider produced inconsistent result after apply in elasticstack_fleet_integration_policy resource


bplies-ATX:
**Describe the bug**
Cannot successfully use the `elasticstack_fleet_integration_policy` resource with a serious `input` containing anything substantial.

**To Reproduce**
1. TF configuration used '...'

```
data "elasticstack_fleet_integration" "custom_logs_stream" {
  name = "filestream"
}

resource "elasticstack_fleet_integration_policy" "sentinel_app_logs" {
  name                = "Sentinel application logs"
  namespace           = "sentinel"
  description         = ""
  agent_policy_id     = elasticstack_fleet_agent_policy.qa_environment_agent_policy.id
  integration_name    = data.elasticstack_fleet_integration.custom_logs_stream.name
  integration_version = "1.2.0"
  enabled             = true

  input {
    enabled  = true
    input_id = "filestream-filestream"
    streams_json = jsonencode({
      "filestream.generic" : {
        "enabled" : true,
        "vars" : {
          "paths" : ["/logs/sentinel*.log"],
          "data_stream.dataset" : "custom.app",
          "processors" : file("${path.module}/../../common/integration_processors/sentinel.yml"),
          "exclude_files" : ["\\.gz$"],
          # These were observed from Kibana UI request and provided here to avoid drift
          "include_files" : [],
          "tags" : [],
          "recursive_glob" : true,
          "clean_inactive" : -1,
          "harvester_limit" : 0,
          "fingerprint" : true,
          "fingerprint_offset" : 0,
          "fingerprint_length" : 1024,
          "file_identity_native" : false,
          "exclude_lines" : [],
          "include_lines" : []
        }
      }
    })
    vars_json           = "{}"
  }

  lifecycle { # Attempt to smooth over any provider problems for now
    ignore_changes = [
      input,
      vars_json,
    ]
  }
}
```

2. TF operations to execute to get the error '...' [e.g `terraform plan`,`terraform apply`, `terraform destroy`]

```
terraform plan

  # elasticstack_fleet_integration_policy.sentinel_app_logs is tainted, so must be replaced
-/+ resource "elasticstack_fleet_integration_policy" "sentinel_app_logs" {
      ~ id                  = "bda3f7dc-9e5c-4945-a0b2-5ca99ed7d56b" -> (known after apply)
        name                = "Sentinel application logs"
[...]

Issue #889, opened 2024-10-31
produced an unexpected new value: .input[1].streams_json: inconsistent values for sensitive │ attribute.


kaykhan:
Hi im having some issues determining what inconsistent sensitive values it is referring to. 

I was only hoping to update `processors`. I have (i tihnk) replicated the values of the others `vars` into terraform

```
resource "elasticstack_fleet_integration_policy" "kubernetes_policy_integration_policy" {
  name                = "kubernetes-1"
  namespace           = "default"
  description         = "kubernetes-1"
  agent_policy_id     = elasticstack_fleet_agent_policy.eck_agent_policy.policy_id
  integration_name    = elasticstack_fleet_integration.kubernetes_integration.name
  integration_version = elasticstack_fleet_integration.kubernetes_integration.version


  input {
    enabled  = false
    input_id = "audit-logs-filestream"
  }
  input {
    enabled  = true
    input_id = "container-logs-filestream"

    streams_json = jsonencode({
      "kubernetes.container_logs" : {
        "vars" : {
          "paths" : ["/var/log/containers/*$${kubernetes.container.id}.log"],
          "symlinks" : true,
          "containerParserStream" : "all",
          "containerParserFormat" : "auto",
          "data_stream.dataset" : "kubernetes.container_logs",
          "additionalParsersConfig" : "#",
          "custom" : "",
          "processors" : <<YAML
- if:
    equals.kubernetes.labels.log-json-decode: "true"
  then:
    - decode_json_fields:
        fields: ["message"]
        process_array: false
        max_depth: 5
        target: "custom_json"
        overwrite_keys: true
        add_error_key: true
YAML
        }
      }
    })
  }

  input {
    enabled  = true
    input_id = "events-kubernetes/metrics"
  }

  input {
    enabled  = true
    input_id = "kube-apiserver-kubernetes/metrics"
  }

  input {
    enabled  = false
    input_id = "kube-controller-manager-kubernetes/metrics"
  }

  input {
    enabled  = true
    input_id = "kube-proxy-kubernetes/metrics"
  }

  input {
    enabled  = false
    input_id = "kube-scheduler-kubernetes/metrics"
  }

  input {
    enabled  = true
    input_id = "kube-state-metrics-kubernetes/metrics"
  }

  input {
    enabled  = true
    input_id = "kubelet-kubernetes/metrics"
  }
}


```

[...]

PR #1558, opened 2025-12-16
Only require any input parameters to be specified if they differ from the defaults

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/617
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1415
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/889

This PR aims to allow users to only provide streams/vars for integration policies where they differ from the defaults for a specific integration. In reality, users have been attempting to do just that, but hitting some form of inconsistent state error. These errors should no longer occur as we're taking into account the integration defaults when comparing the user config and the API derived state.


internal/fleet/integration_policy/models_defaults.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, nil, nil
+	}
+
+	var diags diag.Diagnostics
+
+	var policyTemplate apiPolicyTemplate
+	var dataStreams []apiDatastream
+
+	if pkg.PolicyTemplates != nil && len(*pkg.PolicyTemplates) > 0 {
+		policyTemplateIf := (*pkg.PolicyTemplates)[0]
```
dimuon: Why do we use the first element only? Does it make sense to comment the decision?

internal/fleet/integration_policy/input_value.go
```diff
@@ -0,0 +1,376 @@
...
+
+	// Compare each enabled stream
+	for streamID, oldStream := range enabledOldStreams {
+		newStream, exists := enabledNewStreams[streamID]
+		if !exists {
+			return false, diags
+		}
+
+		// Compare enabled flags
+		if !oldStream.Enabled.Equal(newStream.Enabled) {
```
dimuon: why do we compare `enable` attributes if we are considering only enabled streams?

internal/fleet/integration_policy/input_value.go
```diff
@@ -0,0 +1,376 @@
...
+func (v InputValue) Equal(o attr.Value) bool {
+	other, ok := o.(InputValue)
+	if !ok {
+		return false
+	}
+	return v.ObjectValue.Equal(other.ObjectValue)
+}
+
+func (v InputValue) MaybeEnabled(ctx context.Context) (bool, diag.Diagnostics) {
+	if v.IsNull() || v.IsUnknown() {
```
dimuon, line 38:
```suggestion
	if !utils.IsKnown(v) {
```

dimuon (approved):
Nice change 👍

internal/fleet/integration_policy/input_value.go
tobio: Good point, I've removed this.

internal/fleet/integration_policy/models_defaults.go
tobio: Good catch, updated this to check all templates.

Issue #1561, opened 2025-12-16
[Bug] Incorrect parsing of dates for anomaly detector datafeed


edsavage:
**Describe the bug**
When updating an anomaly detector datafeed state to be `started` with `start` and `end` times defined in valid RFC 3339 format, the times sent to the job are 
a) different from those configured in the terraform resource
b) result in the datafeed being unable to retrieve any data

**To Reproduce**
1. Create job in ES machine learning with configuration
```
{
  "analysis_config": {
    "bucket_span": "15m",
    "detectors": [
      {
        "function": "count",
        "detector_description": "count"
      },
      {
        "function": "mean",
        "field_name": "nginx.access.body_sent.bytes",
        "detector_description": "mean(\"nginx.access.body_sent.bytes\")"
      }
    ],
    "influencers": [
      "nginx.access.geoip.city_name",
      "nginx.access.user_agent.build"
    ],
    "model_prune_window": "30d"
  },
  "analysis_limits": {
    "model_memory_limit": "110MB",
    "categorization_examples_limit": 4
  },
  "data_description": {
    "time_field": "@timestamp",
    "time_format": "epoch_ms"
  },
  "model_snapshot_retention_days": 10,
  "daily_model_snapshot_retention_after_days": 1,
  "datafeed_config": {
    "datafeed_id": "datafeed-nginx",
    "query": {
      "bool": {
        "must": [
          {
            "match_all": {}
          }
        ]
      }
    },
    "indices": [
      "filebeat-nginx-elasticco-full"
    ]
  }
}
```
2. TF configuration used 
```
resource "elasticstack_elasticsearch_ml_anomaly_detection_job" "nginx" {
  job_id      = "nginx"
  description = "Anomaly detection for network traffic"
  analysis_config = {
    bucket_span = "15m"
    detectors = [
      {
        function             = "count"
        detector_description = "count"
      },
      {
        function             = "mean"
        field_name           = "nginx.access.body_sent.bytes"
        detector_description = "mean(\"nginx.access.body_sent.bytes\")"
      }
    ]
    influencers        = ["nginx.access.geoip.city_name", "nginx.access.user_agent.build"]
    model_prune_window = "30d"
  }
  analysis_limits = {
    model_memory_limit            = "110MB"
    categorization_examples_limit = 4
  }
  data_description = {
    time_field  = "@timestamp"
    time_format = "epoch_ms"
  }
  model_snapshot_retention_days             = 10
[...]

Issue #1557, opened 2025-12-16
[Bug] Changing anomaly detector datafeed state to "started" results in provider error message


edsavage:
**Describe the bug**
After importing an existing job and datafeed an error message

```
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to elasticstack_elasticsearch_ml_datafeed_state.nginx, provider "provider[\"registry.terraform.io/elastic/elasticstack\"]" produced an
│ unexpected new value: .start: was null, but now cty.StringVal("2017-02-02T10:58:43+13:00").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

is emitted after applying a change to start the datafeed.

**To Reproduce**
1. Job configuration in ES
```
{
  "analysis_config": {
    "bucket_span": "15m",
    "detectors": [
      {
        "function": "count",
        "detector_description": "count"
      },
      {
        "function": "mean",
        "field_name": "nginx.access.body_sent.bytes",
        "detector_description": "mean(\"nginx.access.body_sent.bytes\")"
      }
    ],
    "influencers": [
      "nginx.access.geoip.city_name",
      "nginx.access.user_agent.build"
    ],
    "model_prune_window": "30d"
  },
  "analysis_limits": {
    "model_memory_limit": "110MB",
    "categorization_examples_limit": 4
  },
  "data_description": {
    "time_field": "@timestamp",
    "time_format": "epoch_ms"
  },
  "model_snapshot_retention_days": 10,
  "daily_model_snapshot_retention_after_days": 1,
  "datafeed_config": {
    "datafeed_id": "datafeed-nginx",
    "query": {
      "bool": {
        "must": [
          {
            "match_all": {}
          }
        ]
      }
    },
    "indices": [
      "filebeat-nginx-elasticco-full"
    ]
  }
}

```
2. TF configuration used 
```
resource "elasticstack_elasticsearch_ml_anomaly_detection_job" "nginx" {
  job_id      = "nginx"
  description = "Anomaly detection for network traffic"
  analysis_config = {
    bucket_span = "15m"
    detectors = [
      {
        function             = "count"
        detector_description = "count"
      },
      {
        function             = "mean"
        field_name           = "nginx.access.body_sent.bytes"
        detector_description = "mean(\"nginx.access.body_sent.bytes\")"
      }
    ]
    influencers        = ["nginx.access.geoip.city_name", "nginx.access.user_agent.build"]
[...]

PR #1563, opened 2025-12-17
Fix state consistency issues with ML Datafeed start and end times.

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1557
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1561

This PR:
* Correctly sets the `start` time to unknown if the datafeed state if transitioning from `stopped` -> `started`
* Persists any user specified timezones in the `start` and `end` times to avoid a state consistency issue when the user supplied timezone does not match the ES server timezone. 
* Fixes a bug where the `start` and `end` times were being set as Unix second values when the API was expecting Unix milli-second values.

### DIFF
--- a/internal/elasticsearch/ml/datafeed_state/models.go
+++ b/internal/elasticsearch/ml/datafeed_state/models.go
@@ -50,12 +50,14 @@ func (d *MLDatafeedStateData) SetStartAndEndFromAPI(datafeedStats *models.Datafe
 		}
 
 		if datafeedStats.RunningState.SearchInterval != nil {
-			start, diags := timeInSameLocation(datafeedStats.RunningState.SearchInterval.StartMS, d.Start)
+			start, timeDiags := timeInSameLocation(datafeedStats.RunningState.SearchInterval.StartMS, d.Start)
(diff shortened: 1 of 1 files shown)


internal/elasticsearch/ml/datafeed_state/schema.go
```diff
@@ -74,6 +74,7 @@ func GetSchema() schema.Schema {
 				Computed:            true,
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.UseStateForUnknown(),
+					SetUnknownIfStateHasChanges(),
 				},
 			},
 			"end": schema.StringAttribute{
```
dimuon, line 80: Shall we treat it similarly to “start”? (e.g. use `SetUnknownIfStateHasChanges`)?

tobio: `end` isn't computed so setting it to unknown isn't valid.

dimuon: Yeah, I know, but I meant making it Computed as well :)

Issue #1093, opened 2025-04-07
[Bug] policy_id can not be set for elasticstack_fleet_integration_policy


mattias-wallenberg-stegra:
**Describe the bug**
When applying, terraform gave the following output:

> Error: Provider produced inconsistent result after apply
│
│ When applying changes to module.elastic.elasticstack_fleet_integration_policy.example, provider "provider[\"registry.terraform.io/elastic/elasticstack\"]" produced an unexpected new
│ value: .policy_id: was cty.StringVal("example-policy-id"), but now cty.StringVal("f222d2c6-1e1e-41d1-9bbe-a6afd2d5048c").

It appears that the policy id given is not respected - instead a new UUID is assigned.

**To Reproduce**

Tf configuration:

```
resource "elasticstack_fleet_integration_policy" "example" {
  name                = "Example"
  namespace           = "default"
  policy_id           = "example-policy-id"
  agent_policy_id     = elasticstack_fleet_agent_policy.example.policy_id
  integration_name    = elasticstack_fleet_integration.example.name
  integration_version = elasticstack_fleet_integration.example.version

  # vars_json and input redacted from this bug report
}
```

The error appeared when using `terraform apply`. If I remove the `policy_id` I can apply without error

**Expected behavior**
The integration policy should be created with the policy id `example-policy-id`

**Additional context**
I am using terraform v1.11.3 and provider version v0.11.14. Latest released Kibana and elastic version at the time of writing.

PR #1594, opened 2025-12-29
Set the ID in the integration policy create request if it's known

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1093


dimuon:
(needs changelog)

Issue #1595, opened 2025-12-30
[Bug] Fleet Agent Policy creation fails for Kibana space-scoped roles due to lack of space-aware API routing


patricwi:
**Describe the bug**
The elasticstack Terraform provider (`v0.13.1`) is unable to create Fleet Agent Policies when the executing user’s permissions are restricted to a specific Kibana space. While the resource configuration allows for the definition of space_ids, the provider's underlying API client does not utilize space-scoped endpoints, forcing a global API call that requires "All Spaces" (cluster-wide) privileges.

**To Reproduce**
1. Create a space called `test-space`.
2. Create a user `test-tf-space` with the role `test-fleet-space`, so it has only access to modify the fleet policies and integrations for a single space:
```
{
  "test-fleet-space": {
    "cluster": [
      "monitor"
    ],
    "indices": [],
    "applications": [
      {
        "application": "kibana-.kibana",
        "privileges": [
          "feature_fleetv2.all",
          "feature_fleet.all"
        ],
        "resources": [
          "space:test-space"
        ]
      }
    ],
    "run_as": [],
    "metadata": {},
    "transient_metadata": {
      "enabled": true
    }
  }
}
```
3. Attempt to run the TF script:
```
terraform {
  required_version = ">= 1.0.0"
  required_providers {
    elasticstack = {
      source  = "elastic/elasticstack"
      version = "0.13.1"
    }
  }
}

provider "elasticstack" {
  elasticsearch {
    username   = "test-tf-space"
    password   = "<password>"
    endpoints = ["<ES endpoint>"]
  }
  kibana {
    username   = "test-tf-space"
    password   = "<password>"
    endpoints = ["<KB endpoint>"]
  }
}

resource "elasticstack_fleet_agent_policy" "test_policy" {
  space_ids        = ["test-space"]
  name             = "ela-test-dev"
  namespace        = "test"
  description      = "ela test policy"
  sys_monitoring   = true
  monitor_logs     = true
  monitor_metrics  = true
  host_name_format = "fqdn"
}

data "elasticstack_fleet_enrollment_tokens" "test_policy" {
  policy_id = elasticstack_fleet_agent_policy.test_policy.id
  space_id  = "test-space"
}
```
--> it will fail with

```
│ Error: Unexpected status code from server: got HTTP 403
│ 
│  with elasticstack_fleet_agent_policy.test_policy,
[...]

PR #1597, opened 2025-12-31
Allow agent policies to be created by space restricted roles

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1595


internal/fleet/agent_policy/create.go
```diff
@@ -39,8 +39,18 @@ func (r *agentPolicyResource) Create(ctx context.Context, req resource.CreateReq
 		return
 	}
 
+	var spaceID string
+	if !planModel.SpaceIds.IsNull() && !planModel.SpaceIds.IsUnknown() {
```
dimuon:
```suggestion
	if utils.IsKnown(planModel) {
```

PR #1616, opened 2026-01-13
Fix integration policy vars_json bug

### LINKED ISSUE #1628: Fatal error: concurrent map writes in fleet/integration_policy.getPackageInfo when managing 40+ resources
## Description

The provider crashes with a fatal error due to concurrent map writes when managing a large number of Kibana spaces (40) with associated fleet integration policies.

## Error Message

```
fatal error: concurrent map writes

goroutine 10860 [running]:
internal/runtime/maps.fatal({0x104022968?, 0x4?})
github.com/elastic/terraform-provider-elasticstack/internal/fleet/integration_policy.getPackageInfo({0x10493d078, 0x140083e2930}, 0x14000690400, {0x1400779d100, 0x3}, {0x1400779d103, 0x5})
        /Users/jamesgarside/Github/Elastic/terraform-provider-elasticstack/internal/fleet/integration_policy/resource.go:93 +0x284
github.com/elastic/terraform-provider-elasticstack/internal/fleet/integration_policy.(*integrationPolicyResource).Read(0x14000093ea8, {0x10493d078, 0x140083e2930}, {{{{0x1049461f8, 0x14003cd6600}, {0x1045e39e0, 0x14006fd1b30}}, {0x10494aaa8, 0x14001973860}}, 0x0, ...}, ...)
        /Users/jamesgarside/Github/Elastic/terraform-provider-elasticstack/internal/fleet/integration_policy/read.go:62 +0x4b0
```

## Context

This occurs during `terraform apply` when:
- Managing 40 Kibana spaces (bt-01 through bt-40) + default space
- Each space has associated fleet integration policies
- High resource count triggers concurrent operations

Additional errors appear alongside the crash:
```
Error: failed to get prebuilt rules status: 429 Too Many Requests
Error: Plugin did not respond
Error: Request cancelled
```

The 429 rate limiting errors suggest the concurrent operations may be overwhelming the Kibana API, which then triggers the race condition in the map access.

## Environment

- **Provider Version**: `v0.13.1`
- **Terraform Version**: `>= 1.5`
- **Go Version**: `1.25.4`
- **Platform**: macOS (darwin-arm64)
- **Resource Count**: 40+ spaces with associated resources

## Affected Resources

- `elasticstack_kibana_install_prebuilt_rules`
- Fleet integration policies (likely all resources using `getPackageInfo`)

## Root Cause Analysis

The stack trace indicates:
1. `getPackageInfo` function at `internal/fleet/integration_policy/resource.go:93` is accessing a map
2. Multiple goroutines are simultaneously writing to this map without synchronization
3. This is a classic race condition - the map is not protected by a mutex or other synchronization mechanism

## Reproduction Steps

1. Create Terraform configuration with 40+ Kibana spaces
2. Add fleet integration policies or prebuilt rules for each space
3. Run `terraform apply` with default parallelism
4. Provider crashes with concurrent map writes

## Expected Behavior

The provider should safely handle concurrent resource operations, either by:
- Using `sync.Map` instead of regular maps for shared state
- Protecting map access with `sync.RWMutex`
- Serializing access to package info cache

### Temporary workarounds that help:

1. **Reduce parallelism**:
   ```bash
   terraform apply -parallelism=5
   ```

2. **Target specific resources**:
   ```bash
   terraform apply -target=resource.name
   ```

3. **Comment out problematic resources** temporarily

## Suggested Fix

In `internal/fleet/integration_policy/resource.go:93` and related code, the package info cache/map needs proper synchronization:

```go
// Instead of:
var packageInfoCache = make(map[string]*PackageInfo)

// Use:
var packageInfoCache = sync.Map{} // or protect with sync.RWMutex
```

## Additional Context

This is a critical bug that makes the provider unusable for larger deployments with many spaces. The race condition is deterministic enough that it occurs consistently with 40+ resources.

## References

- Related to concurrent access patterns in fleet integration policy management
- May affect other provider resources that cache package information

jamesagarside:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1742

---------------------------------------
## Overview
This PR contains multiple fixes and enhancements for Fleet resources:

1. **Integration Policy `vars_json` Bug Fix** - Fixes HTTP 400 error when reconciling policies
2. **Integration Policy Race Condition Fix** - Fixes concurrent map writes causing fatal crashes (#1628)
3. **Fleet Proxy Resource** - New resource for proxy configuration
4. **Fleet Output Sensitive Field Handling** - Fixes null value overwrites

### Problem
After manually upgrading Kibana versions, Terraform reconciliation would fail with HTTP 400 errors. The provider was sending internal `__tf_provider_context` metadata to the Kibana API.

### Solution
Use `.SanitizedValue()` instead of `.ValueString()` to strip internal metadata before API calls.

### Files Changed
- `internal/fleet/integration_policy/models.go`
- `internal/fleet/integration_policy/models_test.go`

### Problem
Fixes issue #1628 - Fatal concurrent map writes when managing 40+ fleet integration policies across multiple Kibana spaces. The `knownPackages` cache used an unsynchronized map causing race conditions during Terraform's parallel resource operations.

**Error Location:** `internal/fleet/integration_policy/resource.go:93` in `getPackageInfo()`

**Trigger:** 40 Kibana spaces × 12 integration policy types = 480+ resources with Terraform's default parallelism (10)

### Solution
Replace `map[string]kbapi.PackageInfo` with `sync.Map` for thread-safe concurrent access without explicit locking.

### Changes
- Replace unsynchronized map with `sync.Map` in `knownPackages`
- Update `getPackageInfo()` to use `Store()` method
- Update `getCachedPackageInfo()` to use `Load()` with type assertion
- Update `populateVarsJSONDefaults()` for `sync.Map` compatibility
- Fix test functions to use `Store()`/`Delete()` instead of map assignments

### Verification
✅ Build successful  
✅ All unit tests pass with `-race` flag  
✅ No data races detected

### Files Changed
- `internal/fleet/integration_policy/resource.go`
- `internal/fleet/integration_policy/vars_json_value.go`
- `internal/fleet/integration_policy/vars_json_value_test.go`

## 3. Fleet Proxy Resource (New)

Implements `elasticstack_fleet_proxy` resource for routing Fleet Server and Elasticsearch traffic through proxy servers.

### Features
- Full CRUD operations with import support
- TLS certificate configuration
- Space-aware multi-tenancy
- Integration with Fleet Output and Server Host resources

### Files Added
- `internal/fleet/proxy/` (complete resource implementation)
- Documentation and examples

### Problem
Pre-existing bug where sensitive fields were being overwritten with null values after API calls, causing "Provider produced inconsistent result" errors.

### Solution
Preserve sensitive fields during state reconciliation using selective field restoration.

### Files Changed
- `internal/fleet/output/models_sensitive.go`
- `internal/fleet/output/models_sensitive_test.go`

## Testing
- All existing tests pass
- New acceptance tests added for Fleet Proxy resource
- Race detector confirms no data races in integration policy cache

## Related Issues
- Fixes #1628 (Race condition in integration policy cache)

## Deployment Impact
This fix unblocks large-scale deployments managing 100+ integration policies across multiple Kibana spaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fix value conversion error in `elasticstack_elasticsearch_index_alias` when indices are unknown at plan time. ([#1755](https://github.com/elastic/terraform-provider-elasticstack/pull/1755))
 - Fix state consistency error in `elasticstack_kibana_security_exception_list` when `os_types` are used in Elastic Stack 9.2 ([#1740](https://github.com/elastic/terraform-provider-elasticstack/pull/1740))
 - Fix state consistency error in `elasticstack_kibana_security_detection_rule` when `threat_filter` is supplied ([#1758](https://github.com/elastic/terraform-provider-elasticstack/pull/1758))
+- Fix state consistency error in `elasticstack_fleet_integration_policy` when the policy was updated outside of the Terraform workflow ([#1616](https://github.com/elastic/terraform-provider-elasticstack/pull/1616))
 
 ## [0.14.2] - 2026-02-19
 


internal/fleet/integration_policy/acc_test.go
```diff
@@ -576,6 +576,77 @@ func TestAccResourceIntegrationPolicyGCPVertexAI(t *testing.T) {
...
+	policyName := sdkacctest.RandStringFromCharSet(22, sdkacctest.CharSetAlphaNum)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { acctest.PreCheck(t) },
+		CheckDestroy: checkResourceIntegrationPolicyDestroy,
+		Steps: []resource.TestStep{
+			{
+				ProtoV6ProviderFactories: acctest.Providers,
+				SkipFunc:                 versionutils.CheckIfVersionIsUnsupported(minVersionIntegrationPolicy),
+				Config: `
```
tobio: ✅ Already done in commit 8c2c1ae5 - configs moved to testdata/TestAccResourceIntegrationPolicy_VersionUpdate/{create,update}/main.tf

internal/fleet/integration_policy/update.go
```diff
@@ -26,6 +26,24 @@ func (r *integrationPolicyResource) Update(ctx context.Context, req resource.Upd
...
 
+	// Preserve computed fields from state before building the API request
+	// This ensures fields like agent_policy_id are included in the update request
+	// even when they're not explicitly changed by the user
+	if utils.IsKnown(stateModel.ID) {
+		planModel.ID = stateModel.ID
+	}
+	if utils.IsKnown(stateModel.PolicyID) {
+		planModel.PolicyID = stateModel.PolicyID
+	}
```
tobio: Both of these fields are set to `UseStateForUnknown` which should render these blocks redundant. Debugging the acceptance test in this PR confirms that, is there something I'm missing here?

internal/fleet/integration_policy/update.go
```diff
@@ -26,6 +26,24 @@ func (r *integrationPolicyResource) Update(ctx context.Context, req resource.Upd
...
+	// Only preserve optional fields when plan doesn't have a value (user didn't change them)
+	// This prevents overwriting user changes while still fixing the null bug
+	if !utils.IsKnown(planModel.AgentPolicyID) && utils.IsKnown(stateModel.AgentPolicyID) {
+		planModel.AgentPolicyID = stateModel.AgentPolicyID
+	}
+	if !utils.IsKnown(planModel.AgentPolicyIDs) && utils.IsKnown(stateModel.AgentPolicyIDs) {
+		planModel.AgentPolicyIDs = stateModel.AgentPolicyIDs
+	}
```
tobio: Neither of these fields are marked as computed, and the new acceptance test continues to pass after commenting these blocks out. It's not clear why we need this change, and since these fields are only ever user provider (i.e not computed) this seems like a bug.

internal/fleet/integration_policy/acc_test.go
```diff
@@ -576,6 +576,77 @@ func TestAccResourceIntegrationPolicyGCPVertexAI(t *testing.T) {
 	})
 }
 
+// TestAccResourceIntegrationPolicy_VersionUpdate tests that updating integration_version
+// preserves agent_policy_id without causing inconsistent state errors.
+// This is a regression test for https://github.com/elastic/terraform-provider-elasticstack/pull/1616
+func TestAccResourceIntegrationPolicy_VersionUpdate(t *testing.T) {
```
tobio, line 802: This test passes locally when I revert all changes in `update.go`, so I don't think we're testing the actual failure scenario here. 

I _can_ reproduce the error if one test step uses `agent_policy_id` and the other test step configures `agent_policy_ids`, I'm not sure if that's the bug you're hitting, but it seems relevant.

internal/fleet/integration_policy/update.go
```diff
@@ -90,24 +108,20 @@ func (r *integrationPolicyResource) Update(ctx context.Context, req resource.Upd
 
 	// Restore the agent policy field that was originally configured
 	// This prevents populateFromAPI from changing which field is used
+	// IMPORTANT: Use state values, not API response, to avoid null values causing inconsistent state
 	if stateUsedAgentPolicyID && !stateUsedAgentPolicyIDs {
```
tobio: So, assuming the acceptance test I have is the failure case you're hitting (`agent_policy_id` -> `agent_policy_ids`) then I think these conditions are the real bug here. 

Why do we care what's set in state here? I could understand checking if these values are configured in the plan since that's what we're using to build the update request. We should be using the value from the updated policy rather than just copying the configured value though.

tobio:
Merging in main will fix the failing acceptance tests.

internal/fleet/integration_policy/acc_test.go
```diff
@@ -576,6 +576,64 @@ func TestAccResourceIntegrationPolicyGCPVertexAI(t *testing.T) {
...
 }
 
+// TestAccResourceIntegrationPolicy_VersionUpdate tests that updating integration_version
+// works correctly without causing inconsistent state errors.
+//
+// Note: This test validates basic version update functionality. The actual fix in PR #1616
+// addresses vars_json sanitization (stripping internal __tf_provider_context metadata before
+// sending to Kibana API), which prevents HTTP 400 errors when reconciling after manual
+// integration upgrades in the Kibana UI. That specific scenario is difficult to test in
+// an automated acceptance test.
```
tobio: Can we utilise [`PreConfig`](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/index/data_stream_lifecycle/acc_test.go#L55-L79) to manipulate the integration policy outside of TF so we're actually testing the failure case here please?

internal/fleet/output/create.go
```diff
@@ -48,12 +48,32 @@ func (r *outputResource) Create(ctx context.Context, req resource.CreateRequest,
 		return
 	}
 
+	// Preserve sensitive field values from plan before populating from API
+	// The Fleet API does not return sensitive field values for security reasons
+	originalConfigYaml := planModel.ConfigYaml
```
tobio: We've duplicated this preservation logic across create/read/update. Can we instead move this into `populateFromAPI` and re-use it instead?

internal/fleet/proxy/schema.go
```diff
@@ -0,0 +1,80 @@
...
+				Description: "PEM-encoded certificate authorities for verifying the proxy server certificate.",
+				Optional:    true,
+			},
+			"certificate_key": schema.StringAttribute{
+				Description: "PEM-encoded private key for the client certificate.",
+				Optional:    true,
+				Sensitive:   true,
+			},
+			"is_preconfigured": schema.BoolAttribute{
+				Description: "Indicates if the proxy is preconfigured (managed outside Terraform). Note: This field cannot be updated after creation; changes will force resource recreation.",
```
tobio: > managed outside Terraform

Is this correct? The API docs are pretty sparse, but AFAICT `is_preconfigured` has nothing to do with being managed by TF?

internal/fleet/proxy/models.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (model *proxyModel) populateFromAPI(ctx context.Context, proxy *kbapi.GetFleetProxiesItemidResponse) (diags diag.Diagnostics) {
+	if proxy == nil || proxy.JSON200 == nil {
+		return
+	}
+
+	item := proxy.JSON200.Item
+
+	// Set computed fields
+	model.ID = types.StringValue(item.Id)
```
tobio: Should `ID` be a `CompositeID` like the other Kibana resources capturing `<space_id>/<proxy_id>`.

examples/resources/elasticstack_fleet_proxy/with_output.tf
tobio, line 1: I don't think these extra examples are providing huge value. Can we combine them into a single example file so we can just use the default docs template?

jamesagarside:
## 🚨 Critical Bug Fix Added: Race Condition in Integration Policy Cache

I've added a fix for issue #1628 - a **critical race condition** that causes fatal crashes when managing 40+ integration policies.

### The Problem
The `knownPackages` cache was using an unsynchronized `map[string]kbapi.PackageInfo`, causing concurrent map writes during Terraform's parallel resource operations:

```
fatal error: concurrent map writes
goroutine 123 [running]:
internal/fleet/integration_policy.getPackageInfo(...)
    resource.go:93
```

**Trigger scenario**: 40 Kibana spaces × 12 integration policies = 480+ resources with Terraform parallelism = 10

### The Solution
Replaced the unsynchronized map with `sync.Map` for thread-safe concurrent access:

```go
// Before
var knownPackages = map[string]kbapi.PackageInfo{}
knownPackages[key] = pkg
pkg, ok := knownPackages[key]

// After  
var knownPackages sync.Map
knownPackages.Store(key, pkg)
value, ok := knownPackages.Load(key)
pkg := value.(kbapi.PackageInfo)
```

### Verification
✅ Provider builds successfully  
✅ All unit tests pass with `-race` flag enabled  
✅ No data races detected  
✅ Cache functionality preserved

### Impact
This unblocks large-scale deployments managing 100+ integration policies across multiple Kibana spaces, which previously resulted in immediate crashes.

**Files changed:**
- `internal/fleet/integration_policy/resource.go`
- `internal/fleet/integration_policy/vars_json_value.go`
- `internal/fleet/integration_policy/vars_json_value_test.go`

bplies-ATX:
Referenced in Support Case # 02050657

tobio (approved):

Issue #1535, opened 2025-12-10
[Bug] Jira connector setup causes error on UI (action edit)


girtsLv:
**Describe the bug**
The `elasticstack_kibana_security_detection_rule` provider creates rules with Jira connector actions that appear to be configured correctly, but when the rule is opened in the Kibana UI edit mode, an error occurs: "Error: Cannot read properties of undefined (reading 'issueType')".

**Expected behavior**
The Jira connector action should be properly configured so that when the rule is opened in the Kibana UI edit mode, all fields are correctly populated and no errors occur.

**Debug output**
When a rule with a Jira connector action is created via Terraform and then opened in the Kibana UI edit mode, the following error is displayed:

Error Message:
```
Error: Cannot read properties of undefined (reading 'issueType')
```

**Screenshots**
<img width="875" height="723" alt="Image" src="https://github.com/user-attachments/assets/337458ac-2b8c-4e58-a9dd-901377df64a1" />

**Versions (please complete the following information):**
- Terraform Version: 1.5.7
- Elasticstack Provider Version: 0.13.0
- Provider Source: `elastic/elasticstack`

**Additional context**
Add any other context about the problem here. Links to specific affected code files and paths here are also extremely useful (if known).The following minimal Terraform configuration demonstrates the issue:

```terraform
locals {
  jira_description = <<-EOT
    Rule {{context.rule.name}} generated {{state.signals_count}} alerts

    {{{context.results_link}}}

    {{#context.alerts}}
    - **Host Name**: {{host.name}}
    - **Customer**: {{data_stream.namespace}}
    - **User Name**: {{user.name}}
    - **Threat Name**: {{file.Ext.malware_classification.identifier}} {{winlog.event_data.Threat_Name}} {{winlog.event_data.Threat Name}}
    - **Path**: {{file.path}} {{winlog.event_data.Path}}
    - **Hash**: {{file.hash.sha1}}
    - **Quarantine**: {{file.Ext.quarantine_path}}
    - **Process**: {{process.command_line}}
    -------------------------------
    {{/context.alerts}}
  EOT
}

resource "elasticstack_kibana_security_detection_rule" "minimal_rule" {
  name        = "Minimal Detection Rule"
  description = "A minimal security detection rule"
  rule_id     = "minimal-rule-002"
  type        = "query"
  index       = ["logs-*"]
  query       = "event.code: 1116 and event.code: 1117"

  actions = [
    {
[...]

PR #1634, opened 2026-01-21
fix: detection rule alert params formatting issue (ref:#1535)

fabiogermann:
This PR fixes the issue where for example the JIRA action requires nested params.
https://github.com/elastic/terraform-provider-elasticstack/issues/1535

### DIFF
--- a/internal/kibana/security_detection_rule/schema.go
+++ b/internal/kibana/security_detection_rule/schema.go
@@ -369,7 +369,7 @@ func GetSchema() schema.Schema {
 						},
 						"params": schema.MapAttribute{
 							ElementType:         types.StringType,
-							MarkdownDescription: "Object containing the allowed connector fields, which varies according to the connector type.",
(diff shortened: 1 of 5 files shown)


internal/kibana/security_detection_rule/schema.go
```diff
@@ -369,7 +369,7 @@ func GetSchema() schema.Schema {
 						},
 						"params": schema.MapAttribute{
```
tobio, line 370: I think we should be changing this attribute to a `StringAttribute` with `CustomType: jsontypes.NormalizedType{}`. `Params` then holds the entire JSON encoded params instead of magically translating the string values into another type.

fabiogermann: I was considering that, but did not want to add a breaking change for existing usage.

fabiogermann: @tobio would you want to introduce a breaking change?

tobio: Sorry, I'd totally missed this notification somehow. Yep, ultimately this change is more consistent with the rest of the provider, and less prone to bugs moving forward. This is a pretty simple breaking change to adapt to, on a recently released resource as well which helps.

tobio:
Fixed in https://github.com/elastic/terraform-provider-elasticstack/issues/2339

Issue #352, opened 2023-06-05
[Feature] support attaching ILM policies to managed index templates


yohancourbe:
**Is your feature request related to a problem? Please describe.**
ILM policy is attached to an index template by setting the `index.lifecycle.name` on the resource `elasticstack_elasticsearch_index_template` which works as expected. The problem comes with managed index templates that are managed by the cluster itself (e.g. installed through Fleet integration) to which we would to attach our ILM policy, but that we wouldn't want to have managed by Terraform.

**Describe the resource you would like to have implemented.**
We should have a resource like `elasticstack_elasticsearch_index_lifecycle_index_template_attachment` or something alike.

**Describe the solution you'd like**
For managed index templates, the ILM policy should be possible to attach without managing the index template from Terraform

**Describe alternatives you've considered**
* Importing the managed index template and replicate its configuration - heavy workaround, not scalable
* Create the ILM with Terraform and set up the link manually - not scalable
* Create a higher priority index template that I manage - might break with integration upgrade

**Additional context**
n/a

PR #1641, opened 2026-01-26
Index template attachment

dimuon:
## Overview
This PR implements a new resource that attaches ILM policies to Fleet-managed or externally-managed index templates by creating/updating the `<template>@custom` component template.
Addresses: https://github.com/elastic/terraform-provider-elasticstack/issues/352

## Use Case
Fleet-managed index templates (e.g., logs-system.syslog) cannot be directly managed by Terraform without taking full ownership. This resource allows users to attach ILM policies to these templates by leveraging the `@custom` component template pattern that Fleet includes in composed_of.

## Key Features
* Non-destructive: Preserves existing settings in `@custom` component templates
* Clean deletion: Only removes the ILM setting
* Merge behavior: Reads existing template, merges ILM setting, writes back
* Warnings: Logs warning on create if ILM setting already exists (potential conflict)

### DIFF
--- a/internal/elasticsearch/index/templateilmattachment/acc_test.go
+++ b/internal/elasticsearch/index/templateilmattachment/acc_test.go
@@ -111,6 +111,13 @@ func TestAccResourceIndexTemplateIlmAttachment_preservesTemplateOnDestroy(t *tes
 	resource.Test(t, resource.TestCase{
 		PreCheck: func() {
 			acctest.PreCheck(t)
+			notSupported, err := versionutils.CheckIfVersionIsUnsupported(templateilmattachment.MinVersion)()
+			if err != nil {
+				t.Fatalf("checking version: %v", err)
+			}
+			if notSupported {
+				t.Skip("Elasticsearch version does not support this test")
+			}
 			createPreservesTestComponentTemplate(t)
 		},
 		CheckDestroy: checkPreservesTemplateDestroy,


internal/elasticsearch/index/index_template_ilm_attachment/create.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	// Merge the ILM setting
+	componentTemplate.Template.Settings = mergeILMSetting(
+		componentTemplate.Template.Settings,
+		plan.LifecycleName.ValueString(),
+	)
+
+	// Write the component template
+	if sdkDiags := elasticsearch.PutComponentTemplate(ctx, r.client, &componentTemplate); sdkDiags.HasError() {
```
tobio, line 102: This resource exclusively operates on component templates, but the original issue asks to manage the ILM of the final template. 

Should we be managing the final template here instead? 
Do we need to support both component templates, and templates?
Is that one resource or two if we do need to support both types?

internal/elasticsearch/index/index_template_ilm_attachment/read.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients/elasticsearch"
+	"github.com/elastic/terraform-provider-elasticstack/internal/diagutil"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+)
+
+const customSuffix = "@custom"
+
+func (r *Resource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
```
tobio: Can this re-use `readILMAttachment` instead of duplicating much of the logic?

internal/elasticsearch/index/index_template_ilm_attachment/resource.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Ensure provider defined types fully satisfy framework interfaces
+var _ resource.Resource = &Resource{}
+var _ resource.ResourceWithConfigure = &Resource{}
+var _ resource.ResourceWithImportState = &Resource{}
+
+var (
+	// MinVersion is the minimum Elasticsearch version required for this resource.
+	// Minimum version requirement - will be validated during CI testing.
+	MinVersion = version.Must(version.NewVersion("8.0.0"))
```
tobio: Templates and ILM have existed well before this, are we being overly prescriptive here?

internal/elasticsearch/index/index_template_ilm_attachment/schema.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+)
+
+func (r *Resource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = getSchema()
+}
+
+func getSchema() schema.Schema {
+	return schema.Schema{
+		Description: "Attaches an ILM policy to a Fleet-managed or externally-managed index template " +
```
tobio: We could extract this multi-line description into a markdown file and embed it into the Go codebase to have less nasty long strings hanging about.

templates/resources/elasticsearch_index_template_ilm_attachment.md.tmpl
```diff
@@ -0,0 +1,73 @@
...
+subcategory: "Index"
+description: |-
+{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
+---
+
+# {{.Name}} ({{.Type}})
+
+{{ .Description | trimspace }}
+
+See the [official Elastic documentation](https://elastic.co/guide/en/fleet/current/data-streams-scenario3.html) for more information about customizing Fleet-managed data streams.
```
tobio: Can we pull all the extra content out of this template and into the resource description itself so that we can delete this template and just use the default?

internal/elasticsearch/index/index_template_ilm_attachment/create.go
dimuon: How can we manage the final template instead? Shall we just overwrite `settings.index.lifecycle.name` in it?

We don’t need to support both types - we just use the `@custom` component template to attach an ILM policy to a Fleet-managed template, which appears to be the [recommended](https://www.elastic.co/docs/reference/fleet/data-streams#data-streams-index-templates-edit) way to customize Fleet-managed templates. 

While the current approach is still quite niche, what alternatives do we have to address the original issue?

internal/elasticsearch/index/index_template_ilm_attachment/resource.go
dimuon: Hmm... acc tests on CI [failed](https://github.com/elastic/terraform-provider-elasticstack/actions/runs/21365876224/job/61497465700) with 404 for 7.17.13.

dimuon: Probably because it fails to fetch `@custom` component template.

internal/elasticsearch/index/index_template_ilm_attachment/create.go
tobio: Yea, I think you're right, and we're not locking ourselves into anything if we do need to broaden the scope here later on.

> While the current approach is still quite niche

This is my main concern, however the original issue is also quite niche (managing ILM on system/Fleet managed templates), which like you say recommend doing exactly what you're doing here. 

Is there a world where someone wants to manage the final index template independently to the ILM setting but not use a component template for ILM? Probably, but we can add support for that down the track if it does prove to be a common use-case (I'd be surprised).

internal/elasticsearch/index/index_template_ilm_attachment/resource.go
tobio: I think this is fine if we're restricting this to component templates.

dimuon:
@tobio , I’ve addressed the comments and added more tests. Would you mind giving it another look?

tobio (approved):

Issue #1465, opened 2025-11-20
[Bug] Cannot add/update Metric filter in elasticstack_kibana_slo resource


jmoon-elastic:
**Describe the bug**
Metric filter is not added or updated when applying the change to [elasticstack_kibana_slo](https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/kibana_slo) resource.

**To Reproduce**
1. TF configuration used 
```
resource "elasticstack_kibana_slo" "custom_kql_slo_elasticsearch_cluster_health" {
  count = var.deploy_slo_elasticsearch_cluster_health ? 1 : 0

  name        = "Elasticsearch Cluster Health"
  description = "Elasticsearch data availability based on cluster health status"

  kql_custom_indicator {
    index           = ".monitoring-*"
    good            = "cluster_state.status: \"green\""
    total           = "cluster_state.status: *"
    filter          = ""
    timestamp_field = "@timestamp"
  }

  time_window {
    duration = "30d"
    type     = "rolling"
  }

  budgeting_method = "occurrences"

  objective {
    target           = 0.99
  }

  group_by = [ "elasticsearch.cluster.name" ]

  settings {
    sync_delay = "5m"
    frequency  = "5m"
  }

}

resource "elasticstack_kibana_slo" "timeslice_slo_search_latency_on_boomi_index" {
  count = var.deploy_slo_boomi_timeslice_search_latency ? 1 : 0

  name        = "Search Latency on Boomi Index"
  description = "This SLO tracks search latency on a given index pattern below a certain threshold. Please adjust the index pattern (boomi*) and search threshold (2 seconds) as required."

  timeslice_metric_indicator {
    index           = ".monitoring-*"
    timestamp_field = "@timestamp"
    filter          = "elasticsearch.index.name: boomi_*"
    metric {
      metrics {
        name        = "A"
        aggregation = "max"
        field       = "index_stats.total.search.query_time_in_millis"
        filter      = ""
      }
      metrics {
        name        = "B"
        aggregation = "min"
        field       = "index_stats.total.search.query_time_in_millis"
        filter      = "index_stats.total.search.query_total > 100"
      }
      metrics {
        name        = "C"
        aggregation = "max"
        field       = "index_stats.total.search.query_total"
#       filter      = ""
      }
      metrics {
        name        = "D"
        aggregation = "min"
        field       = "index_stats.total.search.query_total"
#        filter      = ""
      }
[...]

PR #1647, opened 2026-01-28
Migrate SLO resource to the plugin framework

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/1465


internal/kibana/slo/acc_test.go
dimuon: Do we still need it? Also, can we remove `ResourceSlo()` (and maybe the whole internal/kibana/slo.go)

internal/kibana/slo/constants.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/hashicorp/go-version"
+)
+
+var (
+	SLOSupportsGroupByMinVersion                = version.Must(version.NewVersion("8.10.0"))
+	SLOSupportsMultipleGroupByMinVersion        = version.Must(version.NewVersion("8.14.0"))
+	SLOSupportsPreventInitialBackfillMinVersion = version.Must(version.NewVersion("8.15.0"))
+	SLOSupportsDataViewIDMinVersion             = version.Must(version.NewVersion("8.15.0"))
+)
+
```
dimuon, line 13: Are these vars replace the ones from slo.go? If so, can we remove the old ones?

PR #1652, opened 2026-01-29
Validate panel config blocks match the configured panel type.

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/650

### DIFF
--- a/internal/kibana/dashboard/schema.go
+++ b/internal/kibana/dashboard/schema.go
@@ -197,7 +197,7 @@ func getPanelSchema() schema.NestedAttributeObject {
 	return schema.NestedAttributeObject{
 		Attributes: map[string]schema.Attribute{
 			"type": schema.StringAttribute{
-				MarkdownDescription: "The type of the panel (e.g. 'visualization', 'search', 'map', 'lens').",
+				MarkdownDescription: "The type of the panel (e.g. 'DASHBOARD_MARKDOWN', 'lens').",
 				Required:            true,
 			},
 			"grid": schema.SingleNestedAttribute{
--- a/internal/utils/validators/conditional.go
+++ b/internal/utils/validators/conditional.go
@@ -39,46 +39,58 @@ func (v condition) MarkdownDescription(ctx context.Context) string {
 	return v.description()
 }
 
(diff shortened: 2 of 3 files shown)


internal/utils/validators/conditional.go
```diff
@@ -49,9 +52,32 @@ func (v condition) MarkdownDescription(ctx context.Context) string {
...
+	if v.dependentPathExpression != nil {
+		// Merge the path expression with the current path to resolve relative references
+		merged := currentPath.Expression().Merge(*v.dependentPathExpression)
+		matchedPaths, matchDiags := config.PathMatches(ctx, merged)
+		diags.Append(matchDiags...)
+		if diags.HasError() {
+			return false, "", diags
+		}
+
+		// For validation purposes, we expect exactly one match
```
dimuon: It doesn't sound totally correct - we just use the first matched path.

internal/kibana/dashboard/schema.go
```diff
@@ -252,6 +253,7 @@ func getPanelSchema() schema.NestedAttributeObject {
 				},
 				Validators: []validator.Object{
 					objectvalidator.ConflictsWith(path.MatchRelative().AtParent().AtName("config_json")),
+					validators.AllowedIfDependentPathExpressionOneOf(path.MatchRelative().AtParent().AtName("type"), []string{"DASHBOARD_MARKDOWN"}),
```
dimuon, line 256: Does it make sense to add `DASHBOARD_MARKDOWN` to `type` field markdown description?

internal/utils/validators/conditional_test.go
```diff
@@ -620,3 +616,469 @@ func TestDependantPathOneOf_Description(t *testing.T) {
...
+
+			config := tfsdk.Config{
+				Raw:    rawConfig,
+				Schema: testSchema,
+			}
+
+			// Create validator using path expression
+			v := ForbiddenIfDependentPathExpressionOneOf(
+				path.MatchRelative().AtParent().AtName("protocol"),
+				[]string{"https"},
```
dimuon, line 969: Not a big deal, but would `http` make more sense than `https` here?

PR #1656, opened 2026-01-29
WIP restructuring client generation

tobio:
Related to https://github.com/elastic/kibana/pull/249978
Related to https://github.com/elastic/kibana-team/issues/2787

### DIFF
--- a/generated/kbapi/Makefile
+++ b/generated/kbapi/Makefile
@@ -2,7 +2,7 @@
 SHELL := /bin/bash
 ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
-github_ref ?= 55be4b139f0e0076c896c1d8a6a8011aae76afc2
+github_ref ?= refs/pull/249978/head
 oas_url    := https://raw.githubusercontent.com/elastic/kibana/$(github_ref)/oas_docs/output/kibana.yaml
 
 .PHONY: all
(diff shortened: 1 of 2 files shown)


TinaHeiligers:
@tobio I've started work on the client generator and need to know what the long term vision is for `transform_schema.go`, given that there's a component name mismatch between what Kibana names components to what the provider expects component names to be. (related to the mapping function we discussed ~ 2 months ago)

Our options are:

- Path A: Rename transform (short-term bridge)
    Map `component` names to expected names so existing transforms keep working. It's incremental but means we have to keep maintaining `transform_schema.go`.

- Path B: Align Kibana `component` naming to what the provider expects (long-term)
    Modify the Kibana code to produce names that match what `oapi-codegen` and the provider already expect (e.g. `agent_policy` instead of `ApiFleetAgentPolicies_Get_Response_200_Item`). This would let the provider gradually delete transform logic rather than adding more.

- Path C: Accept Kibana component names, update the provider to use new `Go` types
    Let Kibana name components its way, update the provider code to use the new generated type names. This is a bigger bang but gets rid of transforms fastest.

If the goal is to eliminate transform_schema.go, I'd suggest:

- Short-term: Use a minimal rename map just to unblock the pipeline and prove it works end-to-end (what I'm doing now)
- Medium-term: Progressively delete transform sections as Kibana handles more ref-to-component conversions, rather than adding rename entries
- Long-term: The transforms that do structural changes (discriminators, simplifying vars, removing deprecated `anyOf` branches, deleting properties) are the ones that
need to move upstream into Kibana's OAS definitions or overlays — renaming alone won't eliminate those

The question for you: are you OK with the `Go` type names changing (e.g. `AgentPolicy` → `ApiFleetAgentPoliciesGetResponse200Item`)? If yes, Path C is
    fastest. If not, we need either Kibana-side naming control or a thin rename layer that shrinks over time.

Issue #1677, opened 2026-02-04
[Bug] Empty sets are not handled gracefully in the elasticstack_elasticsearch_security_role resource


wandergeek:
**Describe the bug**
The provider does crashes if some default fields are used in the `elasticstack_elasticsearch_security_role` resource. Empty sets are not handled gracefully.

**To Reproduce**

```hcl
resource "elasticstack_elasticsearch_security_role" "slo_readwrite" {
  name    = "slo_readwrite"
  cluster = [
    "manage_index_templates",
    "manage_ilm",
    "manage_pipeline",
    "manage_transform"
  ]

  indices {
    names      = ["slo-*", ".slo-*"]
    privileges = ["all"]
    field_security {
      grant  = ["*"]
      except = []
    }
    allow_restricted_indices = false
  }

  applications {
    application = "kibana-.kibana"
    privileges  = ["feature_slo.all"]
    resources   = ["*"]
  }

  run_as = []

  metadata = jsonencode({})
}
```

Init with terraform >= `v0.13.0`:

```hcl
terraform {
  required_providers {
    elasticstack = {
      source  = "elastic/elasticstack"
       version = "0.13.0" 
    }
  }
}
```

Run `terraform init && terraform apply`
```
Terraform will perform the following actions:

  # elasticstack_elasticsearch_security_role.slo_readwrite will be created
  + resource "elasticstack_elasticsearch_security_role" "slo_readwrite" {
      + cluster  = [
          + "manage_ilm",
          + "manage_index_templates",
          + "manage_pipeline",
          + "manage_transform",
        ]
      + id       = (known after apply)
      + metadata = jsonencode({})
      + name     = "slo_readwrite"
      + run_as   = []

      + applications {
          + application = "kibana-.kibana"
          + privileges  = [
              + "feature_slo.all",
            ]
          + resources   = [
              + "*",
            ]
        }

      + indices {
          + allow_restricted_indices = false
          + names                    = [
              + ".slo-*",
              + "slo-*",
            ]
          + privileges               = [
              + "all",
            ]

          + field_security {
              + except = []
              + grant  = [
                  + "*",
                ]
            }
        }
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

elasticstack_elasticsearch_security_role.slo_readwrite: Creating...
╷
[...]

PR #1679, opened 2026-02-04
Fix inconsistency with empty sets in security role resource

wandergeek:
Fixes: https://github.com/elastic/terraform-provider-elasticstack/issues/1677


internal/elasticsearch/security/role/acc_test.go
```diff
@@ -152,6 +152,36 @@ func TestAccResourceSecurityRoleEmptySets(t *testing.T) {
 	})
 }
 
+// When the API returns nil for optional fields with defaults (due to omitempty),
+// the provider should return the actual default values (false for allow_restricted_indices,
+// empty set for except), not null, to prevent "planned set element does not correlate
+// with any element in actual" errors.
+func TestAccResourceSecurityRoleDefaultValues(t *testing.T) {
```
tobio, line 159: This test still passes after reverting the changes in `models.go`, is there a specific Elastic Stack version you're seeing this failure with? 

From the issue, it looks like you're seeing a crash in 0.13.0, https://github.com/elastic/terraform-provider-elasticstack/pull/1581 seems to address the same issue, has this already been fixed in main?

PR #1701, opened 2026-02-10
Validate alerting rule params and reject unknown keys

wandergeek:
## Summary
- preserve generated alerting params schema variants while keeping runtime request compatibility as `map[string]interface{}`
- add provider-side `rule_type_id`-aware params validation with required-key and unexpected-key checks, including union schema support and compatibility allowances for known Kibana payload variants
- add focused unit tests covering valid/invalid known schemas, unknown rule-type pass-through, union matching, and extra-key rejection behavior

## Reference
- Plan gist: https://gist.github.com/wandergeek/b72e618055f91ffa8da7ecfd47bb81cd
- Kibana issue to fix spec: https://github.com/elastic/kibana/issues/252451

## Test plan
- [x] `make lint`
- [x] `make docs-generate`
- [x] `make test`
- [x] `make docker-testacc`

### DIFF
--- a/internal/kibana/alerting_rule/models.go
+++ b/internal/kibana/alerting_rule/models.go
@@ -3,7 +3,6 @@ package alerting_rule
 import (
 	"context"
 	"encoding/json"
-	"strings"
 
 	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
 	"github.com/elastic/terraform-provider-elasticstack/internal/models"
@@ -253,7 +252,7 @@ func (m alertingRuleModel) toAPIModel(ctx context.Context, serverVersion *versio
 		if utils.IsKnown(m.RuleTypeID) {
 			errs := validateRuleParams(m.RuleTypeID.ValueString(), params)
 			if len(errs) > 0 {
-				diags.AddAttributeError(path.Root("params"), "Invalid params for rule_type_id", strings.Join(errs, "; "))
+				diags.AddAttributeError(path.Root("params"), "Invalid params for rule_type_id", formatParamsValidationErrors(errs))
 				return models.AlertingRule{}, diags
 			}
 		}
--- a/internal/kibana/alerting_rule/validate.go
+++ b/internal/kibana/alerting_rule/validate.go
@@ -14,6 +14,25 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 )
 
+// paramsSchemaSpec contains precomputed key metadata and decode factory for
+// one generated params variant. This avoids reflection in runtime validation.
+type paramsSchemaSpec struct {
+	name         string
+	newTarget    func() interface{}
+	requiredKeys map[string]struct{}
+	allowedKeys  map[string]struct{}
+}
+
+func mustNewParamsSchemaSpec(name string, newTarget func() interface{}) paramsSchemaSpec {
+	requiredKeys, allowedKeys := paramsSchemaKeys(newTarget())
+	return paramsSchemaSpec{
+		name:         name,
+		newTarget:    newTarget,
+		requiredKeys: requiredKeys,
+		allowedKeys:  allowedKeys,
+	}
+}
+
 func (r *Resource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
 	var data alertingRuleModel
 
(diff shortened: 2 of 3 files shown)


internal/kibana/alerting_rule/models.go
```diff
@@ -248,6 +249,15 @@ func (m alertingRuleModel) toAPIModel(ctx context.Context, serverVersion *versio
 			diags.AddError("Failed to unmarshal params", err.Error())
 			return models.AlertingRule{}, diags
 		}
+
+		if utils.IsKnown(m.RuleTypeID) {
+			errs := validateRuleParams(m.RuleTypeID.ValueString(), params)
+			if len(errs) > 0 {
+				diags.AddAttributeError(path.Root("params"), "Invalid params for rule_type_id", strings.Join(errs, "; "))
```
tobio:
```suggestion
				diags.AddAttributeError(path.Root("params"), "Invalid params for rule_type_id", strings.Join(errs, "\n"))
```

The diagnostic works with multiline strings pretty well. Does breaking each error onto it's own line make things easier to consume?

internal/kibana/alerting_rule/validate.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+var ruleTypeAdditionalAllowedParamsKeys = map[string][]string{
+	// The generated type currently models legacy single-window fields, while
+	// Kibana accepts modern multi-window payloads under `windows`.
+	"slo.rules.burnRate": {"windows", "dependencies"},
+	// Kibana supports passing selected hit fields to actions, but that key is
+	// currently missing from generated `.es-query` params models.
+	".es-query": {"sourceFields"},
+}
```
tobio, line 137: It sounds like these need to be added to the Kibana spec? Can we open a Kibana issue for doing that?

internal/kibana/alerting_rule/validate.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+		if best == nil || len(candidateErrs) < len(best) {
+			best = candidateErrs
+		}
+	}
+
+	return best
+}
+
+func requiredKeysMissing(target interface{}, params map[string]interface{}) []string {
```
tobio: Couple of points here:
* We don't support nested structs within the params types. I think that's probably fine, but is it? Have you checked? Hmm, it's not even that params would have nested objects, but if the openapi spec ever had shared params types this code would break:

```
components:
  commonParams:
    properties:
      sharedParam:...
  params1:
    allOf:
      - $ref: commonParams
      - properties:
        specificParam:..
  params2:
    allOf:
      - $ref: commonParams
      - properties:
        otherSpecificParam:..
```

* I think this code would be much simpler if we created some intermediate representation of the params once and used it in both places. Something like:
```map[string]struct{isRequired bool}``` where the key is the param key. We'd isolate the reflection mess and just be doing map operations in the actual validation. 
* All this reflection stuff is pretty tightly coupled to the current oapi-codegen implementation. It's actually pretty simple to parse the actual API spec with https://pkg.go.dev/github.com/getkin/kin-openapi/openapi3 and use the actual source of truth (we'd have to embed the spec into the provider binary, but worse things have happened). 

I think we can (and should) implement this without reflection:
1. Use the API spec via openapi3. It wouldn't be 'difficult' and would be nice to use the actual spec, but there's a bunch of complexity there still.
2. Use [json2](https://pkg.go.dev/encoding/json/v2#RejectUnknownMembers):
  * Setting `RejectUnknownMembers` will give us an error if the user provided params specifies something not in the type
  * Marshalling the zero-value of the API params struct to JSON, and then to map[string]interface{} will give us the required (non-pointer, non-omit-zero) fields.

Clear as mud? Happy to talk about it. I'm concerned this is a bit of a maintenance den moving forward.

File: generated/kbapi/transform_schema.go [hunk 4]
```diff
@@ -987,19 +987,14 @@ func fixDashboardPanelItemRefs(schema *Schema) {
...
+// fixAlertingRuleParams keeps the POST alerting rule params anyOf refs so
+// concrete params_* component models are generated, but overrides the request
+// field type to map[string]interface{} for runtime compatibility.
 func fixAlertingRuleParams(schema *Schema) {
 	postEndpoint := schema.MustGetPath("/api/alerting/rule/{id}").MustGetEndpoint("post")
 	paramsSchema := postEndpoint.MustGetMap("requestBody.content.application/json.schema.properties.params")
 
-	paramsSchema.Delete("anyOf")
-	paramsSchema.Set("type", "object")
+	paramsSchema.Set("x-go-type", "map[string]interface{}")
```
tobio: Without this, the generated type would have a `UnmarshalJSON` function which would be pretty simple to switch over to but avoid adding another spec mutation. 

```suggestion
```

internal/kibana/alerting_rule/validate.go
```diff
@@ -0,0 +1,228 @@
...
+		if len(missingKeys) > 0 {
+			candidateErrs = append(candidateErrs, fmt.Sprintf("missing required params keys: %s", strings.Join(missingKeys, ", ")))
+		}
+		if len(unexpectedKeys) > 0 {
+			candidateErrs = append(candidateErrs, fmt.Sprintf("unexpected params keys: %s", strings.Join(unexpectedKeys, ", ")))
+		}
+		if best == nil || len(candidateErrs) < len(best) {
+			best = candidateErrs
+		}
+	}
```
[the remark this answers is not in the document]
wandergeek: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/elastic/terraform-provider-elasticstack/pull/1701#discussion_r2785757479)

PR #1710, opened 2026-02-11
Enable security detection rules by tags

tehbooom:
## Overview

This PR creates a new resource to enable security detection rules. This extends https://github.com/elastic/terraform-provider-elasticstack/pull/1296 when we enabled prebuilt rules. Very simple implementation as it matches rules based on tags which uses key value pairs.

Below is an example of the resource

```hcl
resource "elasticstack_kibana_security_enable_rule" "windows" {
  key                            = "OS"
  value                         = "windows"
  disable_on_destroy = false
}
```

Tested with prebuilt rules installed on an already deployed cluster

## Features

- Enable rules by tag
- Allow rules to be enabled even after destroy with `disable_on_destroy = false` (default is true)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Changes
 * Add `elasticstack_elasticsearch_index_template_ilm_attachment` resource to attach ILM policies to Fleet-managed or externally-managed index templates via the `@custom` component template. ([#1641](https://github.com/elastic/terraform-provider-elasticstack/pull/1641))
+- Add `elasticstack_kibana_security_enable_rule` resource ([1710](https://github.com/elastic/terraform-provider-elasticstack/pull/1710))
 
 ## [0.14.2] - 2026-02-19
 
@@ -113,7 +114,6 @@ inputs = {
 
 ### Changes
 
-- Add `elasticstack_kibana_security_enable_rule` resource ([1710](https://github.com/elastic/terraform-provider-elasticstack/pull/1710))
 - Add import support for `elasticstack_elasticsearch_script` resource ([#1637](https://github.com/elastic/terraform-provider-elasticstack/pull/1637))
 - Migrate `elasticstack_kibana_alerting_rule` to use plugin framework ([#1664](https://github.com/elastic/terraform-provider-elasticstack/pull/1664))
 - Migrate `elasticstack_kibana_slo` resource to the Terraform plugin framework ([#1647](https://github.com/elastic/terraform-provider-elasticstack/pull/1647))


internal/kibana/security_enable_rule/acc_test.go
```diff
@@ -0,0 +1,359 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("elasticstack_kibana_security_enable_rule.test", "space_id", spaceID),
+					resource.TestCheckResourceAttr("elasticstack_kibana_security_enable_rule.test", "key", tagKey),
+					resource.TestCheckResourceAttr("elasticstack_kibana_security_enable_rule.test", "value", tagValue),
+					resource.TestCheckResourceAttr("elasticstack_kibana_security_enable_rule.test", "all_rules_enabled", "true"),
+					checkRulesEnabled(spaceID, tagKey, tagValue, true),
+				),
+			},
+			{
+				// Manually disable one rule outside of Terraform to test drift detection
```
tobio, line 97: ❤️

internal/kibana/security_enable_rule/update.go
```diff
@@ -0,0 +1,65 @@
...
+		return diags
+	}
+
+	serverVersion, sdkDiags := r.client.ServerVersion(ctx)
+	diags.Append(diagutil.FrameworkDiagsFromSDK(sdkDiags)...)
+	if diags.HasError() {
+		return diags
+	}
+
+	minVersion := version.Must(version.NewVersion("8.11.0"))
```
tobio: nit: Can we pull this out to a package level variable? Leaving it here makes it possible for the provider to panic only once this function is called if the version was somehow invalid. Moving it to a package level variable will force a panic as the binary is loaded.

internal/kibana/security_enable_rule/schema.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+)
+
+func (r *EnableRuleResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Enables Elastic Security detection rules based on tags. This resource uses the Kibana bulk action API to enable all rules matching the specified tag key-value pair. Requires Elastic Stack version 8.11.0 or higher. See https://www.elastic.co/guide/en/security/current/rules-ui-management.html",
```
tobio:
```suggestion
		Description: "Enables Elastic Security detection rules based on tags. This resource uses the Kibana bulk action API to enable all rules matching the specified tag key-value pair. Requires Elastic Stack version 8.11.0 or higher. See the [documentation](https://www.elastic.co/guide/en/security/current/rules-ui-management.html) for more details. ",
```

internal/kibana/security_enable_rule/read.go
```diff
@@ -0,0 +1,73 @@
...
+		return
+	}
+
+	serverVersion, sdkDiags := r.client.ServerVersion(ctx)
+	resp.Diagnostics.Append(diagutil.FrameworkDiagsFromSDK(sdkDiags)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	minVersion := version.Must(version.NewVersion("8.11.0"))
```
tobio: Can we pull this out to a package level variable so it can be re-used rather than repeated in delete/read/update?

internal/kibana/security_enable_rule/read.go
```diff
@@ -0,0 +1,73 @@
...
+
+		model.AllRulesEnabled = types.BoolValue(true)
+	} else {
+		tflog.Debug(ctx, "All rules matching the tag are enabled", map[string]interface{}{
+			"space_id": spaceID,
+			"key":      key,
+			"value":    value,
+		})
+		model.AllRulesEnabled = types.BoolValue(true)
+	}
```
tobio:
```suggestion
	model.AllRulesEnabled = types.BoolValue(allEnabled)
```

Let's change the implementation a little. Having the `Read` path update the underlying resource isn't ideal here.

Instead, if the read path updates `AllRulesEnabled` to false, and we default `AllRulesEnabled` to true then it'll get marked as drift and trigger an update on the next apply.

internal/kibana/security_enable_rule/schema.go
```diff
@@ -0,0 +1,60 @@
...
+			},
+			"disable_on_destroy": schema.BoolAttribute{
+				Description: "Whether to disable the rules when this resource is destroyed. Defaults to true.",
+				Optional:    true,
+				Computed:    true,
+				Default:     booldefault.StaticBool(true),
+			},
+			"all_rules_enabled": schema.BoolAttribute{
+				Description: "Computed attribute indicating whether all rules matching the tag are currently enabled. Used for drift detection.",
+				Computed:    true,
```
tobio, line 73:
```suggestion
				Computed:    true,
				Default:     booldefault.StaticBool(true),
```

PR #1731, opened 2026-02-18
SLO: avoid KQL indicator drift when good/total omitted

wandergeek:
### Summary
- Prevent Terraform state drift for SLO `kql_custom_indicator` when `good`/`total` are omitted by treating empty-string API values as unset.
- Add unit + acceptance coverage for the omitted `good`/`total` case.

### Changes
- Set `good`/`total` to explicit empty strings in API payloads when not configured (`internal/kibana/slo/models_kql_custom_indicator.go`).
- Ignore empty-string `good`/`total` coming back from the API when populating state.
- Add regression tests:
  - `internal/kibana/slo/models_kql_custom_indicator_test.go`
  - `internal/kibana/slo/acc_test.go` + `internal/kibana/slo/testdata/TestAccResourceSlo_kql_custom_indicator_without_good_total/`

### Testing
- `make test`

_This PR contains changes made with assistance by Cursor_


internal/kibana/slo/models_kql_custom_indicator.go
tobio, line 1: Lets revert the changes to this file (or at least the `populateFromKql...` changes) and just add "" as the default in the schema for both of these attributes, it maps better to the old SDK behaviour of this resource.

PR #1762, opened 2026-02-26
Implement fleet_output data source

daemitus:
I've found a need to reference outputs. Mainly just to pull their output-id for use in agent/integration policies (I've got a terraform_data running curl currently). This adds a `elasticstack_fleet_output` data source, searchable by name and space optionally. It does not implement the various union specific fields, such as kafka/logstash. If someone makes a ticket for it however, it wouldn't be too difficult to pull it in from the resource logic.


internal/fleet/outputds/acc_test.go
```diff
@@ -0,0 +1,94 @@
...
+			{
+				SkipFunc:    versionutils.CheckIfVersionIsUnsupported(minVersionOutput),
+				Config:      testAccDataSourceOutputMissing,
+				ExpectError: regexp.MustCompile("Output not found"),
+			},
+		},
+	})
+}
+
+const testAccDataSourceOutput = `
```
tobio: Can you pull the Terraform config out into external directories ([example](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/fleet/output/acc_test.go#L104-L107)).

tobio (commented):
Thanks for getting back involved :) 

Given we're already retrieving all the outputs and filtering client side, it would make sense to include all the data sources in the data source output (like the [spaces](https://github.com/elastic/terraform-provider-elasticstack/tree/main/internal/kibana/spaces) data source), and allow arbitrary filtering in a for expression instead.

tobio (approved):
Wicked, thanks!

Issue #1668, opened 2026-02-03
[Bug] `elasticstack_elasticsearch_security_role` Inconsistent State Bug


hamdan-elastic:
# Provider Bug: Inconsistent State After Apply for Security Role with Query Fields in Indices

**Describe the bug**

When creating or updating an `elasticstack_elasticsearch_security_role` resource that contains `indices` or `remote_indices` blocks with the `query` field populated, the provider throws "Provider produced inconsistent result after apply" errors. The error occurs even when the role is successfully created in Elasticsearch - this is purely a state comparison issue where the provider cannot properly match set elements containing JSON query strings.

The provider reports that planned set elements "do not correlate with any element in actual", indicating that the provider's hash/comparison logic for set-based resources is unable to handle JSON string values correctly, particularly when comparing the planned state with the actual state returned from Elasticsearch.

**To Reproduce**

1. **TF configuration used** - Create a file `test_security_role_bug.tf` with the following content:

```hcl
terraform {
  required_providers {
    elasticstack = {
      source  = "elastic/elasticstack"
      version = "~> 0.13.1"
    }
  }
}

provider "elasticstack" {
  elasticsearch {
    endpoints = [var.elasticsearch_endpoint]
    api_key   = var.elasticsearch_api_key
  }
}

variable "elasticsearch_endpoint" {
  description = "Elasticsearch endpoint URL"
  type        = string
}

variable "elasticsearch_api_key" {
  description = "Elasticsearch API key"
  type        = string
  sensitive   = true
}

# Static security role definition for bug reproduction
resource "elasticstack_elasticsearch_security_role" "test_role" {
  name        = "test_bug_reproduction"
  description = "Test role to reproduce provider bug with indices query field"
  cluster     = []
  run_as      = []
  metadata    = jsonencode({})

  # First indices block with query field - THIS CAUSES THE BUG
  indices {
    names                    = [".alerts-*", ".internal-alerts-*"]
    privileges               = ["read", "view_index_metadata"]
    allow_restricted_indices = false
    
    query = jsonencode({
      bool = {
        must = [
          {
            term = {
              "labels.datacenter" = {
                value = "BCP"
              }
            }
          },
          {
            term = {
              "kibana.space_ids" = {
                value = "os_windows"
              }
            }
          }
        ]
        filter   = []
        should   = []
        must_not = []
      }
    })

    field_security {
      grant  = ["*"]
      except = []
    }
  }

[...]

PR #1780, opened 2026-03-02
Support empty (`""`) descriptions in Elasticsearch roles 

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/1668

### DIFF
--- a/.notice_rules.json
+++ b/.notice_rules.json
@@ -1,5 +1,6 @@
 {
   "allowlist": [
+    "0BSD",
     "Apache-2.0",
     "BSD-2-Clause",
     "BSD-3-Clause",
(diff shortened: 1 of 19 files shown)


internal/elasticsearch/security/role/models.go
```diff
@@ -274,11 +274,25 @@ func (data *Data) fromAPIModel(ctx context.Context, role *models.Role) diag.Diag
...
 	// Description
-	data.Description = types.StringPointerValue(role.Description)
+	if role.Description != nil {
+		data.Description = types.StringValue(*role.Description)
+	} else {
+		// If the API omits/returns null for description, preserve a configured empty string ("")
+		// to avoid post-apply state consistency issues.
+		switch {
+		case originalDescription.IsNull():
+			data.Description = types.StringNull()
```
dimuon: The first case is reduntant with the `default`

```suggestion
```

dev-docs/reqs/elasticsearch/security/role.md
```diff
@@ -0,0 +1,108 @@
+# `elasticstack_elasticsearch_security_role` — Schema and Functional Requirements
+
+Resource implementation: `internal/elasticsearch/security/role`
+
+## Schema
+
+```hcl
+resource "elasticstack_elasticsearch_security_role" "example" {
```
dimuon, line 8: Does it make sense to refer to the generated schema instead? 
E.g. replace the HCL block with a link:
```
   ## Schema
   See [generated schema documentation](../../../docs/resources/elasticsearch_security_role.md#schema)
   for the full attribute reference (generated from `schema.go`).
```

dev-docs/reqs/template.md
```diff
@@ -0,0 +1,33 @@
+# `<RESOURCE_NAME>` — Schema and Functional Requirements
```
dimuon, line 1: Would it make sense to also add a skill along with some guidelines re using the template and EARS specs?

PR #1841, opened 2026-03-06
Fix index template alias routing

tobio:
From https://github.com/elastic/terraform-provider-elasticstack/pull/1839

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,6 +55,9 @@ jobs:
   test:
     name: Matrix Acceptance Test
     needs: build
+    permissions:
+      contents: read
+      issues: write
     runs-on: ${{ matrix.runner == 'ubuntu-22.04' && 'ubuntu-22.04' || 'ubuntu-latest' }}
     env:
       ELASTIC_PASSWORD: password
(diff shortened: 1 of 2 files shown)


dev-docs/requirements/elasticsearch/index/component_template.md
```diff
@@ -0,0 +1,76 @@
+# `elasticstack_elasticsearch_component_template` — Schema and Functional Requirements
+
+Resource implementation: `internal/elasticsearch/index/component_template.go`
+
+## Schema
+
+```hcl
+resource "elasticstack_elasticsearch_component_template" "example" {
```
dimuon, line 8: Is my understanding correct that this schema definition is the primary source of truth, and that we expect agents to generate the actual schema from it?

dev-docs/requirements/elasticsearch/index/component_template.md
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  template { # required, list(block), max 1
+    mappings = <optional, json object string>
+    settings = <optional, json object string>
+
+    alias { # optional, set(block)
+      name           = <required, string>
+      filter         = <optional, json string>
+      index_routing  = <optional, string>
```
dimuon, line 40: Is it intentional that `component_template` does not mark both `index_routing` and `search_routing` as `Computed`, and also doesn’t define a spec similar to `[REQ-031]` used for the template?

dev-docs/requirements/elasticsearch/index/component_template.md
tobio: Correct

dev-docs/requirements/elasticsearch/index/component_template.md
tobio: Yea, the behaviour with component templates is different (they don't come back set to the `routing` value).

PR #1895, opened 2026-03-13
Support sub-minute schedule values (10s, 30s) for synthetics monitors

shahzad31:
## Support sub-minute schedule values (10s, 30s) for synthetics monitors

Fixes https://github.com/elastic/kibana/issues/232446

### Summary

- Adds support for sub-minute monitor schedules (`0.1` for 10s, `0.5` for 30s) in the `elasticstack_kibana_synthetics_monitor` resource, addressing [elastic/kibana#232446](https://github.com/elastic/kibana/issues/232446)
- Changes the `schedule` attribute from `Int64` to `Float64` to accept fractional values while maintaining backward compatibility with existing minute-based values (1, 3, 5, 10, 15, 30, 60, 120, 240)
- Adds custom JSON marshaling on `MonitorSchedule` so that `0.1` serializes as `"10s"` and `0.5` as `"30s"` when sent to the Kibana API

### Details

The Kibana Synthetics API recently added support for running monitors every 10 or 30 seconds. The API accepts these as string values (`"10s"`, `"30s"`) alongside the existing numeric minute values. Previously, the Terraform provider's `schedule` attribute was an integer, making it impossible to specify sub-minute intervals.

This change:
- Updates `MonitorSchedule` type from `int` to `float64` in the kbapi client library
- Converts the TF schema `schedule` field from `Int64Attribute` to `Float64Attribute`
- Maps `0.1` to/from `"10s"` and `0.5` to/from `"30s"` when communicating with the Kibana API
- Adds unit tests for both sub-minute schedule values

#### Usage

```hcl
resource "elasticstack_kibana_synthetics_monitor" "fast_check" {
  name     = "fast-check"
  schedule = 0.1  # every 10 seconds
  # ...
}

resource "elasticstack_kibana_synthetics_monitor" "medium_check" {
  name     = "medium-check"
  schedule = 0.5  # every 30 seconds
  # ...
}

### CODE REVIEW

File: docs/resources/elasticsearch_security_user.md [hunk 1]
```diff
@@ -63,16 +63,14 @@ resource "elasticstack_elasticsearch_security_user" "dev" {
...
-- `password_wo` (String, Sensitive, [Write-only](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments)) Write-only password attribute for use with ephemeral resources. Passwords must be at least 6 characters long. This attribute is designed for use with ephemeral resources like `vault_kv_secret_v2` to prevent secrets from being stored in the Terraform state. Must be used with `password_wo_version`.
+- `password_wo` (String, Sensitive) Write-only password attribute for use with ephemeral resources. Passwords must be at least 6 characters long. This attribute is designed for use with ephemeral resources like `vault_kv_secret_v2` to prevent secrets from being stored in the Terraform state. Must be used with `password_wo_version`.
 - `password_wo_version` (String) Version identifier for the write-only password. This field is used to trigger updates when the password changes. Required when `password_wo` is set. Typically, you would use a hash of the password or a version identifier from your secret management system.
```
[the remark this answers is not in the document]
shahzad31: @copilot apply changes based on [this feedback](https://github.com/elastic/terraform-provider-elasticstack/pull/1895#discussion_r2931821897)

internal/kibana/synthetics/monitor/schema.go
```diff
@@ -177,14 +178,14 @@ func monitorConfigSchema() schema.Schema {
 					),
 				},
 			},
-			"schedule": schema.Int64Attribute{
+			"schedule": schema.Float64Attribute{
```
tobio, line 181: I'm not a big fan of the change here:
* `0.1m` != `10s`, but also `schedule = 0.1666666666` is obviously terrible. 
* We're just perpetuating a poor mapping between TF and API schemas. 

I think it would be better to recognise the schema doesn't meet the API capabilities here and map things better. This gives us flexibility to adapt to other potential API changes as well. 

IMO that means mimicking the API format more closely:

```
schedule = {
    unit = "s"
    number = 10
}
```

I think this gives us the best chance to maintain support if other scheduling options are added in the future. 

Alternatively, we could turn `schedule` into a proper duration (e.g `schedule = "10s"`) (via `customtypes.DurationType`) and keep the explicit `OneOf` validation on the allowed values. This would be a simpler breaking change, would avoid the weird fractional minutes, but wouldn't necessarily easily support other future scheduling options.

shahzad31: @tobio i guess we could add a new field may be to make it bit more frictionless , but i agree , i guess it's likely not possible to suport number | string in schema

tobio: Yea, so there's a [dynamic attribute](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/dynamic) which could work here. 

In the past we've generally steered away from it because of the preference of [Hashicorp](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/dynamic-data). This case is very simple, so it's definitely not the worst approach and avoids a breaking change. 

I'm happy if you want to explore that path. The other options involve a breaking schema change anyway, so the cost of a future change is similar if there's significant negative sentiment around dynamic.

PR #1902, opened 2026-03-15
Update dashboard spec

tobio:


internal/kibana/dashboard/models_datatable_panel.go
```diff
@@ -105,55 +76,46 @@ func (c datatablePanelConfigConverter) mapPanelToAPI(pm panelModel, apiConfig *k
...
-	}
+	return attrs, diags
+}
 
-	var config1 kbapi.DashboardPanelItemConfig7
-	if err := config1.FromDashboardPanelItemConfig70(config10); err != nil {
-		diags.AddError("Failed to create config1", err.Error())
-		return diags
-	}
+func (c datatablePanelConfigConverter) populateFromAPIPanel(ctx context.Context, pm *panelModel, attrs kbapi.KbnDashboardPanelLens_Config_0_Attributes) diag.Diagnostics {
```
dimuon: Method like this (populateFromAPIPanel, mapPanelToAPI, handlesTFPanelConfig, and handlesAPIPanelConfig; in this converter and others) are used only in tests now. 
Shall we change tests too? Do we still need these methods?

internal/kibana/dashboard/models_metric_panel.go
```diff
@@ -32,110 +32,56 @@ import (
...
-	}
-
-	attrsMap, ok := attrs.(map[string]any)
-	if !ok {
-		return nil
-	}
-
-	// Marshal and unmarshal to get the MetricChart
-	attrsJSON, err := json.Marshal(attrsMap)
+func (c metricChartPanelConfigConverter) populateFromAttributes(ctx context.Context, pm *panelModel, attrs kbapi.KbnDashboardPanelLens_Config_0_Attributes) diag.Diagnostics {
```
dimuon, line 46: `metricChartConfigModel.fromAPI` is called only from tests now while this funcs isn't covered by tests (at least directly).

internal/kibana/dashboard/acc_gauge_panels_test.go
```diff
@@ -85,6 +85,10 @@ func TestAccResourceDashboardGauge(t *testing.T) {
 				ResourceName:      "elasticstack_kibana_dashboard.test",
 				ImportState:       true,
 				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{
```
dimuon, line 88: Why do we ignore this in many tests now?

internal/kibana/dashboard/models_datatable_panel.go
tobio: Removed them

internal/kibana/dashboard/acc_gauge_panels_test.go
tobio: On my list to see if we can tighten these up. I'm kind of ignoring it for now so we can make progress on keeping things up to date but it's tracked in the follow up task.

Issue #1619, opened 2026-01-13
[Feature] Export/Import Agent builder agents, tools, and workflows


tehbooom:
**Is your feature request related to a problem? Please describe.**
Exporting and transferring agents, tools, and workflows between clusters is somewhat of a headache. If you want export an agent you created you need to export all related tools and or workflows then import them. With an export and import feature this should would make it easier to share between clusters.

**Describe the resource you would like to have implemented.**
## Export

Export an agent with its tools and or workflows [get agent by id API](https://www.elastic.co/docs/api/doc/kibana/operation/operation-get-agent-builder-agents-id)

> This would export each tool and if the tool is a workflow it would also export that as well. So later we can reference its read-only attributes to import/create the workflows first, then tools, and finally the agent
```hcl
data "elasticstack_kibana_export_ab_agent" "example" {
  id = "agent.id"
  include_dependencies = true
}
```

Export an agent only (default) [get agent by id API](https://www.elastic.co/docs/api/doc/kibana/operation/operation-get-agent-builder-agents-id)
```hcl
data "elasticstack_kibana_export_ab_agent" "example" {
  id = "agent.id"
}
```

Export tools with [get tool by id API](https://www.elastic.co/docs/api/doc/kibana/operation/operation-get-agent-builder-tools-toolid)
```hcl
data "elasticstack_kibana_export_ab_tool" "example" {
  id = "tool.id"
}
```

Export workflow with `GET /api/workflows/{workflow_id}`
```hcl
data "elasticstack_kibana_export_ab_workflow" "example" {
  id = "workflow-id123"
}
```

## Import

Create an agent with [create an agent](https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-agent-builder-agents)
```hcl
resource "elasticstack_kibana_ab_agent" "example" {
  id = "agent.id"
  name =  "Search Index Helper"
  labels =  ["custom-indices", "department-search"]
  description =  "Hi! I can help you search the data within the indices starting with \"content-\" prefix."
  avatar_color = "#BFDBFF"
  avatar_symbol =  "SI"
  tools = [
    "platform.core.search",
    "platform.core.list_indices",
    "platform.core.get_index_mapping",
    "platform.core.get_document_by_id"
   ]
  instructions = "You are a custom agent that wants to help searching data using all indices starting with prefix \"content-\"."
}
```

[...]

PR #1923, opened 2026-03-17
Add workflow resource and data source

tehbooom:
Related #1619 

Adds workflow resource and data source

> [!NOTE]
> ### Add `elasticstack_kibana_agentbuilder_workflow` resource and `elasticstack_kibana_agentbuilder_export_workflow` data source
> - Adds a new `elasticstack_kibana_agentbuilder_workflow` resource supporting full CRUD for Kibana Agent Builder workflows, including composite `space_id/workflow_id` identity, auto-generated IDs, and validity diagnostics on create/update.
> - Adds a new `elasticstack_kibana_agentbuilder_export_workflow` data source that reads a workflow definition by ID (supporting composite IDs) and returns its YAML configuration.
> - Introduces a `NormalizedYamlValue` custom Terraform type in [`normalized_yaml_value.go`](https://github.com/elastic/terraform-provider-elasticstack/pull/1923/files#diff-55922c980d27604048fef8e9734bc552cf368a5e126f76193d2a35c528a426ce) that validates YAML syntax and uses semantic equality (normalizing map key order, ignoring whitespace) to suppress spurious plan diffs.
> - Both resource and data source are gated to Kibana >= 9.4.0-SNAPSHOT and registered in the provider via [`plugin_framework.go`](https://github.com/elastic/terraform-provider-elasticstack/pull/1923/files#diff-5a66d72c3748bd201add43c9271e7b8dc29f28fd06eee792c837aa87e70a063b).
> - Risk: the update API returns a partial response, so a follow-up GET is always issued to populate state; any latency or failure in that second call will surface as an error.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bb9eb7e.</sup>
> 
>


internal/kibana/agentbuilderworkflow/create.go
```diff
@@ -0,0 +1,74 @@
...
+	resp.Diagnostics.Append(diagutil.FrameworkDiagsFromSDK(sdkDiags)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if serverVersion.LessThan(minKibanaAgentBuilderAPIVersion) {
+		resp.Diagnostics.AddError("Unsupported server version",
+			fmt.Sprintf("Agent Builder workflows require Elastic Stack v%s or later.", minKibanaAgentBuilderAPIVersion))
+		return
+	}
```
tobio:
```suggestion
	supported, sdkDiags := r.client.EnforceMinVersion(ctx, minKibanaAgentBuilderAPIVersion)
	resp.Diagnostics.Append(diagutil.FrameworkDiagsFromSDK(sdkDiags)...)
	if resp.Diagnostics.HasError() {
		return
	}

	if !supported {
		resp.Diagnostics.AddError("Unsupported server version",
			fmt.Sprintf("Agent Builder workflows require Elastic Stack v%s or later.", minKibanaAgentBuilderAPIVersion))
		return
	}
```

`EnforceMinVersion` will work better here, it take serverless into account properly.

internal/kibana/agentbuilderworkflow/schema.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+)
+
+func (r *WorkflowResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = getSchema()
+}
+
+func getSchema() schema.Schema {
```
tobio, line 35: Are workflows space specific?

internal/kibana/agentbuilderworkflow/schema.go
```diff
@@ -0,0 +1,68 @@
...
+			"id": schema.StringAttribute{
+				Computed:            true,
+				Optional:            true,
+				MarkdownDescription: "The workflow ID. If not provided, it will be auto-generated. IDs are `workflow-<UUIDv4>`",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"configuration": schema.StringAttribute{
```
tobio:
```suggestion
			"configuration_yaml": schema.StringAttribute{
```

Renaming so we have some flexibility to add a richly typed `configuration` attribute in the future.


We're going to get issues here caused by whitespace/ordering differences. Can you add a normalised custom type here. We use `jsontypes.Normalized` elsewhere and Golang's json serialiser handles yaml just fine, so maybe that works here as well. Otherwise, it wouldn't be a huge effort (like I'm assuming it's roughly 3m watching Cursor do the work) to reimplement the json type as a normalised yaml type within the provider codebase.

internal/kibana/exportagentbuilder/common.go
tobio, line 1: Feels like this could go into `internal/clients/kibanaoapi` or just get added to the `workflows.go` file being added into that package in this PR.

internal/kibana/agentbuilderworkflow/acc_test.go
```diff
@@ -0,0 +1,220 @@
...
+	workflowID := "workflow-" + workflowUUID.String()
+	resourceID := "elasticstack_kibana_agentbuilder_workflow.test"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { preCheckWithWorkflowsEnabled(t) },
+		ProtoV6ProviderFactories: acctest.Providers,
+		Steps: []resource.TestStep{
+			{
+				SkipFunc: versionutils.CheckIfVersionIsUnsupported(minKibanaAgentBuilderAPIVersion),
+				Config: providerConfig + `
```
tobio: Pull this config out into an external `.tf` file. Take a look at the [system user](https://github.com/elastic/terraform-provider-elasticstack/blob/main/internal/elasticsearch/security/user/acc_test.go#L47-L50) tests as an example. 

Aside from making the code a little simpler to track it means these acceptance test configs can be pretty easily used directly with the Terraform CLI if there's ever any manual investigations to do.

internal/kibana/exportagentbuilder/workflow/acc_test.go
tobio, line 1: `data_source_test` -> `acc_test`? 

Maybe you're planning on having a resource here in a follow up in which case ignore me.

internal/kibana/exportagentbuilder/workflow/data_source_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+// TestAccDataSourceKibanaExportABWorkflow tests the export_agentbuilder_workflow data source
+func TestAccDataSourceKibanaExportAgentBuilderWorkflow(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { preCheckWithWorkflowsEnabled(t) },
+		ProtoV6ProviderFactories: acctest.Providers,
+		Steps: []resource.TestStep{
+			{
+				SkipFunc: versionutils.CheckIfVersionIsUnsupported(minKibanaAgentBuilderAPIVersion),
+				Config: `
```
tobio: Same thing with the config.

internal/kibana/exportagentbuilder/workflow/schema.go
```diff
@@ -0,0 +1,50 @@
...
+// Schema defines the schema for the data source.
+func (d *DataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Export an Agent Builder workflow by ID. See https://www.elastic.co/guide/en/kibana/current/agent-builder-api.html",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description: "The workflow ID to export.",
+				Required:    true,
+			},
+			"space_id": schema.StringAttribute{
```
tobio, line 37: We have a `space_id` here, but not in the workflow resource. It sounds like we need one there too.

internal/kibana/agentbuilderworkflow/schema.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (r *WorkflowResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = getSchema()
+}
+
+func getSchema() schema.Schema {
+	return schema.Schema{
+		MarkdownDescription: "Manages Kibana Agent Builder workflows. See the [Workflows documentation](https://www.elastic.co/docs/explore-analyze/workflows) for more information.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
```
tobio, line 39: If we're adding a `space_id` to this resource let's make sure the import path correctly handle spaces. 

Usually that means we have `id` and `workflow_id` where `id` is a composite `<workflow_id>/<space_id>`.

internal/kibana/exportagentbuilder/workflow/data_source.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// DataSource is the data source implementation.
+type DataSource struct {
+	client *clients.APIClient
+}
+
+// Metadata returns the data source type name.
+func (d *DataSource) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_kibana_export_agentbuilder_workflow"
```
tobio:
```suggestion
	resp.TypeName = req.ProviderTypeName + "_kibana_agentbuilder_export_workflow"
```

Does this make more sense? We'd be keeping the `agentbuilder` entities together. This data source would be exporting a workflow within the agentbuilder namespace, rather than exporting an agentbuilder workflow.

internal/kibana/agentbuilderworkflow/schema.go
tehbooom: yes

internal/kibana/agentbuilderworkflow/schema.go
tehbooom: `types.String` should be fine here and here is why.

As a user if I add whitespace, newline, or reorder my current yaml that should directly reflect in Kibana. 

Even though reordering/newlines/whitespaces doesn't matter to yaml it does matter to a user looking at the workflow.

If I do

```yaml
name: Test Workflow
description: A test workflow for acceptance testing
enabled: true
triggers:
  - type: manual
inputs:
  - name: message
    type: string
    default: "hello world"
steps:
  - name: hello_world_step
    type: console
    with:
      message: "{{ inputs.message }}"
  - name: hello_world_step_2
    type: console
    with:
      message: "{{ inputs.message }}"
```

But if I change it to

```yaml
name: Test Workflow
description: A test workflow for acceptance testing
enabled: true
triggers:
  - type: manual
 
inputs:
  - name: message
    type: string
    default: "hello world"

steps:
  - name: hello_world_step
    type: console
    with:
      message: "{{ inputs.message }}"

  - name: hello_world_step_2
    type: console
    with:
      message: "{{ inputs.message }}"
```

This should be a drift and terraform should update the workflow.  The current implementation is fine and does this as expected. I've implemented more tests to validate this behavior

internal/clients/kibanaoapi/workflows.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+// addInternalOriginHeader adds the required x-elastic-internal-origin header for workflow APIs
+func addInternalOriginHeader(_ context.Context, req *http.Request) error {
+	req.Header.Set("x-elastic-internal-origin", "Kibana")
+	return nil
+}
+
+// GetWorkflow reads a specific workflow from the API.
+func GetWorkflow(ctx context.Context, client *Client, workflowID string) (*kbapi.WorkflowDetailDto, diag.Diagnostics) {
```
tobio: The functions in the file don't accept a space parameter, so I don't think we're properly supporting spaces at the moment?

internal/kibana/agentbuilderworkflow/schema.go
tobio: Strongly disagree here. We regularly have issues with Kibana returning subtly different, though semantically equal JSON. We've not had a single issue where a user has issues with those same differences. 

Whilst the raw YAML is more visible in this case, I don't expect to see wildly different outcomes. Yhat is Kibana doesn't always return the _exact_ YAML that was provided, and users don't really care about how it's represented outside of TF.

Issue #1163, opened 2025-06-10
Support for ML Inference API


italodirenzo93:
**Is your feature request related to a problem? Please describe.**
I'm attempting to implement an ingest pipeline for my index that generated `dense_vector`s using the `inference` pipeline processor. This is currently not possible as both the [Inference API](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-put-1) and the [Inference processor](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ingest-put-pipeline#operation-ingest-put-pipeline-body-application-json-processors-inference) are not available in the Elasticsearch Terraform provider.

**Describe the resource you would like to have implemented.**
Broad support for the Inference API but for my use case specifically, the following endpoints as resources:
- [Create an inference endpoint](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-put-1)
- [Update an inference endpoint](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-update)

As well as the following as data objects:
- [Inference pipeline processor](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-ingest-put-pipeline#operation-ingest-put-pipeline-body-application-json-processors-inference)

**Pseudo-code example**
```tf
resource "elasticstack_elasticsearch_inference_endpoint" "my_endpoint" {
  service = "openai"
  service_settings = {
    api_key = "..."
    model_id = "text-embedding-3-small"
  }
}

data "elasticstack_elasticsearch_ingest_processor_inference" "my_proc" {
  model_id = "my_endpoint"
  input_output = {
    input_field = "foo"
    output_field = "bar"
  }
}

resource "elasticstack_elasticsearch_ingest_pipeline" "my_pipeline" {
  name = "..."
  processors= [data.elasticstack_elasticsearch_ingest_processor_append.my_proc.json]
}

resource "elasticstack_elasticsearch_index" "my_index" {
  name = "..."
  default_pipeline = elasticstack_elasticsearch_ingest_pipeline.my_pipeline.name
}
```

PR #1955, opened 2026-03-20
Add Inference endpoint resource

tehbooom:
Relates #1163

Creates an inference endpoint in Elasticsearch. 

Changes to `inference_id`, `task_type`, or `service` force resource replacement, while secrets within `service_settings` and `task_settings` can be updated in place.

Tested with `azureaistudio` on v9.3.1

> [!NOTE]
> ### Add `elasticstack_elasticsearch_inference_endpoint` Terraform resource
> - Adds a new Terraform resource for managing Elasticsearch inference endpoints, supporting create, read, update, delete, and import operations via the Elasticsearch Inference API.
> - Implements drift prevention: `task_settings` state is intersected with user-configured keys to avoid surfacing server-applied defaults, and `service_settings` are preserved to avoid losing sensitive fields on refresh.
> - Adds `chunking_settings` normalization in [chunking_defaults.go](https://github.com/elastic/terraform-provider-elasticstack/pull/1955/files#diff-d3cb8a13feb8f06d73e578f0cbeb07526d57bfe19867f3bda4cf11515a58ef24) that applies strategy-appropriate defaults (e.g. `sentence` or `word`) to align state with API responses.
> - Requires Elasticsearch ≥ 8.18.0; changing `task_type` or `service` forces resource replacement.
> - Acceptance tests cover create/update, no-op plan, import, forced replace on `task_type` change, and task_settings drift scenarios.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b11a4cb.</sup>
> 
>

### DIFF
--- a/internal/elasticsearch/inference/inferenceendpoint/schema.go
+++ b/internal/elasticsearch/inference/inferenceendpoint/schema.go
@@ -20,6 +20,7 @@ package inferenceendpoint
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	providerschema "github.com/elastic/terraform-provider-elasticstack/internal/schema"
 	"github.com/elastic/terraform-provider-elasticstack/internal/utils/customtypes"
@@ -39,6 +40,7 @@ const inferenceEndpointDescription = "Creates or updates an inference endpoint."
 var (
 	MinSupportedVersion = version.Must(version.NewVersion("8.18.0"))
 	validTaskTypes      = []string{"sparse_embedding", "text_embedding", "rerank", "completion", "chat_completion", "embedding"}
+	taskTypesMarkdown   = fmt.Sprintf("[`%s`]", strings.Join(validTaskTypes, "`, `"))
 )
 
 func (r *inferenceEndpointResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
@@ -65,7 +67,7 @@ func (r *inferenceEndpointResource) Schema(_ context.Context, _ resource.SchemaR
 			"task_type": schema.StringAttribute{
 				Optional:            true,
 				Computed:            true,
-				MarkdownDescription: fmt.Sprintf("must be one of [%v]", validTaskTypes),
+				MarkdownDescription: fmt.Sprintf("must be one of %s", taskTypesMarkdown),
 
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplaceIfConfigured(),
--- a/.github/aw/actions-lock.json
+++ b/.github/aw/actions-lock.json
@@ -80,6 +80,11 @@
       "version": "v0.65.5",
       "sha": "15b2fa31e9a1b771c9773c162273924d8f5ea516"
     },
+    "github/gh-aw-actions/setup@v0.67.0": {
+      "repo": "github/gh-aw-actions/setup",
+      "version": "v0.67.0",
+      "sha": "cde65c546c2b0f6d3f3a9492a04e6687887c4fe8"
+    },
     "github/gh-aw/actions/setup@v0.53.2": {
       "repo": "github/gh-aw/actions/setup",
       "version": "v0.53.2",
--- a/docs/resources/elasticsearch_inference_endpoint.md
+++ b/docs/resources/elasticsearch_inference_endpoint.md
@@ -44,7 +44,7 @@ resource "elasticstack_elasticsearch_inference_endpoint" "example" {
 - `chunking_settings` (String) Configuration for chunking input text, as a JSON object. Applicable only for embedding task types.
 - `elasticsearch_connection` (Block List) Elasticsearch connection configuration block. (see [below for nested schema](#nestedblock--elasticsearch_connection))
 - `task_settings` (String) Task-specific settings, as a JSON object. Optional and service-dependent. Only keys explicitly set here are tracked; server-applied defaults returned by the API are ignored to avoid perpetual drift.
-- `task_type` (String) must be one of [[sparse_embedding text_embedding rerank completion chat_completion embedding]]
+- `task_type` (String) must be one of [`sparse_embedding`, `text_embedding`, `rerank`, `completion`, `chat_completion`, `embedding`]
 
 ### Read-Only
 


internal/elasticsearch/inference/inferenceendpoint/schema.go
```diff
@@ -0,0 +1,88 @@
...
+				},
+			},
+			"task_type": schema.StringAttribute{
+				MarkdownDescription: taskTypeDescription,
+				Optional:            true,
+				Computed:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplaceIfConfigured(),
+					stringplanmodifier.UseStateForUnknown(),
+				},
```
tobio, line 75: Can we add `stringvalidators.OneOf(...)` to this to enforce that this value is one of the items listed in the description. Whilst we're there, can we make sure that the description and validation share the same list of values, e.g 

```
description = fmt.Sprintf("must be one of [%v]", validTaskTypes)
...
stringvalidator.OneOf(validTaskTypes)
```

internal/elasticsearch/inference/inferenceendpoint/models.go
```diff
@@ -0,0 +1,159 @@
...
+				diags.AddError("JSON Marshal Error", fmt.Sprintf("Error marshaling service_settings: %s", err))
+				return diags
+			}
+			data.ServiceSettings = jsontypes.NewNormalizedValue(string(b))
+		}
+	}
+
+	// task_settings and chunking_settings: only populate from the API if the
+	// user explicitly configured them. ES returns defaults for these fields
+	// even when not set by the user, which would cause a persistent diff.
```
tobio: What happens when the user configures `task_settings`/`chunking_settings` but doesn't define fields with defaults? IIUC we'd still have a perpetual diff. 

It sounds like we want to use `customtypes.JSONWithDefaultsType` for these values.

internal/elasticsearch/inference/inferenceendpoint/models.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (data *Data) fromAPIModel(_ context.Context, endpoint *elasticsearch.InferenceEndpoint) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	data.InferenceID = types.StringValue(endpoint.InferenceID)
+	data.TaskType = types.StringValue(endpoint.TaskType)
+	data.Service = types.StringValue(endpoint.Service)
+
+	// service_settings: preserve plan value since the API may omit sensitive fields (e.g. api_key)
+	// We only update if the field was previously null/unknown (first read after import).
```
tobio, line 105: Just curious, the [API docs](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-inference-put) use this as an example:

```json
{
 "service": "cohere",
 "service_settings": {
   "model_id": "rerank-english-v3.0",
   "api_key": "{{COHERE_API_KEY}}"
 }
}
```

Does ES interpolate the `COHERE_API_KEY` value from the secret store or is this purely an API example?

internal/elasticsearch/inference/inferenceendpoint/update.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/hashicorp/go-version"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+var MinSupportedVersion = version.Must(version.NewVersion("8.18.0"))
+
+func (r *inferenceEndpointResource) upsert(ctx context.Context, plan tfsdk.Plan, state *tfsdk.State) diag.Diagnostics {
```
tobio: Update doesn't use upsert, can we either move this to `Create` or make it reusable between the two paths, though looking at the client these are two distinct operations.

internal/elasticsearch/inference/inferenceendpoint/models.go
```diff
@@ -0,0 +1,159 @@
...
+			diags.AddError("Invalid chunking_settings JSON", fmt.Sprintf("Error parsing chunking_settings: %s", err))
+			return nil, diags
+		}
+		endpoint.ChunkingSettings = cs
+	}
+
+	return endpoint, diags
+}
+
+func (data *Data) toUpdateModel(_ context.Context) (*elasticsearch.InferenceEndpointUpdate, diag.Diagnostics) {
```
tobio: nit: Ignore me, but `toUpdateModel` could just call `toAPIModel` and then clear ServiceType and ChunkingSettings.

internal/elasticsearch/inference/inferenceendpoint/schema.go
```diff
@@ -0,0 +1,88 @@
...
+				Required:            true,
+				CustomType:          jsontypes.NormalizedType{},
+				Sensitive:           true,
+			},
+			"task_settings": schema.StringAttribute{
+				MarkdownDescription: "Task-specific settings, as a JSON object. Optional and service-dependent.",
+				Optional:            true,
+				CustomType:          jsontypes.NormalizedType{},
+			},
+			"chunking_settings": schema.StringAttribute{
```
tobio, line 99: This isn't included in the update request, should this be marked as `RequiresReplace` as well? Looks like chunking settings is supported in the update endpoint, so maybe this was just missed in toUpdate

internal/elasticsearch/inference/inferenceendpoint/models.go
tehbooom: Purely an example

internal/elasticsearch/inference/inferenceendpoint/schema.go
```diff
@@ -0,0 +1,108 @@
...
+				Required:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"task_type": schema.StringAttribute{
+				MarkdownDescription: taskTypeDescription,
+				Optional:            true,
+				Computed:            true,
+				Description:         fmt.Sprintf("must be one of [%v]", validTaskTypes),
```
tobio:
```suggestion
				Optional:            true,
				Computed:            true,
				MarkdownDescription:         fmt.Sprintf("must be one of [%v]", validTaskTypes),
```

internal/elasticsearch/inference/inferenceendpoint/create.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients/elasticsearch"
+	"github.com/elastic/terraform-provider-elasticstack/internal/diagutil"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func (r *inferenceEndpointResource) upsert(ctx context.Context, plan tfsdk.Plan, state *tfsdk.State) diag.Diagnostics {
```
tobio: `Update` doesn't call `upsert`, we should either actually use `upsert` in both `Create`/`Update` or move it to `Create`. At the moment it looks like they use different API endpoints, so maybe moving this to create is the right solution?

tehbooom: Moved to `Create`

tobio (approved):
🚢 thanks for seeing this through :)

Issue #1425, opened 2025-11-06
[Feature] Kibana Alerting - Allow configuration of Alerting flapping detection


emrcbrn:
**Describe the resource you would like to have implemented.**

Allow the configuration of [flapping detection in the rule body](https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-alerting-rule-id#operation-post-alerting-rule-id-body-application-json-flapping) via Terraform, [which currently cannot be configured](https://registry.terraform.io/providers/elastic/elasticstack/latest/docs/resources/kibana_alerting_rule)

**Describe alternatives you've considered**

Post-installation configuration of such alerts, which is cumbersome to do so rather than implement it as code.

PR #1966, opened 2026-03-22
Alert flapping

tobio:
Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/1425

> [!NOTE]
> ### Add flapping detection support to `elasticstack_kibana_alerting_rule`
> - Adds a new optional `flapping` nested block to the alerting rule resource with `look_back_window`, `status_change_threshold`, and `enabled` attributes, mapped through the Kibana API on create and update.
> - Version-gates the feature: `flapping` requires Kibana ≥8.16.0; `flapping.enabled` requires ≥9.3.0. Attempts on older versions produce plan-time errors.
> - Adds plan-time validation that rejects combining rule-level `notify_when` or `throttle` with per-action `frequency` blocks, and a plan modifier that coerces unknown `notify_when` to null when per-action frequency is present.
> - Omitting `flapping` on update retains the previous value rather than clearing it.
> - Risk: existing configurations that set both rule-level `notify_when`/`throttle` and per-action `frequency` will now fail at plan time.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d2e3c1f.</sup>
> 
>


tobio:
@macroscope-app review

internal/kibana/alertingrule/models.go
```diff
@@ -434,6 +489,22 @@ func (m alertingRuleModel) toAPIModel(ctx context.Context, serverVersion *versio
...
+			return models.AlertingRule{}, diags
+		}
+		rule.Flapping = &models.AlertingRuleFlapping{
+			LookBackWindow:        fm.LookBackWindow.ValueInt64(),
+			StatusChangeThreshold: fm.StatusChangeThreshold.ValueInt64(),
+		}
+		if typeutils.IsKnown(fm.Enabled) && !fm.Enabled.IsNull() {
+			rule.Flapping.Enabled = fm.Enabled.ValueBoolPointer()
+		}
+	}
```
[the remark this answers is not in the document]
tobio: https://github.com/elastic/terraform-provider-elasticstack/pull/1966/changes#diff-bc4ae2eac63cdb0e1efd286eddcf64fc0efb7be0f1b253082a2201b124cdddadR170-R174

Both these attributes are validated as present at plan time

internal/kibana/alertingrule/testdata/TestAccResourceAlertingRuleFlapping/create/rule.tf
```diff
@@ -0,0 +1,62 @@
...
+  connector_type_id = ".index"
+  config = jsonencode({
+    index              = "my-index"
+    executionTimeField = "alert_date"
+  })
+}
+
+resource "elasticstack_kibana_alerting_rule" "test_rule" {
+  name     = var.name
+  rule_id  = "cf33ce2d-9fc4-5131-a350-b5bd6482736c"
```
dimuon: This `rule_id` (`cf33ce2d-9fc4-5131-a350-b5bd6482736c`) is the same UUID used by the `alert_delay` and `alerts_filter` steps in `TestAccResourceAlertingRule`. If these tests run in parallel against the same Kibana instance they'll collide on the same rule. The `FlappingEnabled` test already uses a unique ID (`e133ce2d...`), so this one just needs a fresh UUID.

Also affects `testdata/TestAccResourceAlertingRuleFlapping/update/rule.tf:21`.

internal/clients/kibanaoapi/alerting_rule.go
```diff
@@ -398,6 +413,18 @@ func buildCreateRequestBody(rule models.AlertingRule) kbapi.PostAlertingRuleIdJS
...
+	if rule.Flapping != nil {
+		body.Flapping = &struct {
+			Enabled               *bool   `json:"enabled,omitempty"`
+			LookBackWindow        float32 `json:"look_back_window"`
+			StatusChangeThreshold float32 `json:"status_change_threshold"`
+		}{
+			LookBackWindow:        float32(rule.Flapping.LookBackWindow),
+			StatusChangeThreshold: float32(rule.Flapping.StatusChangeThreshold),
+			Enabled:               rule.Flapping.Enabled,
+		}
```
dimuon: `buildCreateRequestBody` and `buildUpdateRequestBody` contain identical flapping struct construction. Consider extracting a small helper to avoid the duplication drifting over time, e.g.:

```go
func flappingBodyField(f *models.AlertingRuleFlapping) *struct {
	Enabled               *bool   `json:"enabled,omitempty"`
	LookBackWindow        float32 `json:"look_back_window"`
	StatusChangeThreshold float32 `json:"status_change_threshold"`
} {
	if f == nil {
		return nil
	}
	return &struct {
		Enabled               *bool   `json:"enabled,omitempty"`
		LookBackWindow        float32 `json:"look_back_window"`
		StatusChangeThreshold float32 `json:"status_change_threshold"`
	}{
		LookBackWindow:        float32(f.LookBackWindow),
		StatusChangeThreshold: float32(f.StatusChangeThreshold),
		Enabled:               f.Enabled,
	}
}
```

Then both builders just do `body.Flapping = flappingBodyField(rule.Flapping)`.

internal/kibana/alertingrule/acc_test.go
```diff
@@ -591,6 +591,92 @@ func TestAccResourceAlertingRuleAlertDelay(t *testing.T) {
...
+					"name": config.StringVariable(ruleName),
+				},
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("elasticstack_kibana_alerting_rule.test_rule", "name", ruleName),
+					resource.TestCheckResourceAttr("elasticstack_kibana_alerting_rule.test_rule", "flapping.look_back_window", "20"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_alerting_rule.test_rule", "flapping.status_change_threshold", "5"),
+				),
+			},
+		},
+	})
```
dimuon, line 694: There's no test step that removes the `flapping` block after it was set. This is the scenario most likely to surprise users — the design doc acknowledges that omitting `flapping` on update leaves server state unchanged, which can cause a perpetual plan diff after refresh. A third step here that drops the `flapping` block and verifies the resulting state would document this behavior and catch regressions if the omit-on-update semantics ever change.

internal/kibana/alertingrule/acc_test.go
```diff
@@ -591,6 +591,92 @@ func TestAccResourceAlertingRuleAlertDelay(t *testing.T) {
...
+				},
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("elasticstack_kibana_alerting_rule.test_rule", "name", ruleName),
+					resource.TestCheckResourceAttr("elasticstack_kibana_alerting_rule.test_rule", "flapping.look_back_window", "20"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_alerting_rule.test_rule", "flapping.status_change_threshold", "5"),
+					resource.TestCheckResourceAttr("elasticstack_kibana_alerting_rule.test_rule", "flapping.enabled", "false"),
+				),
+			},
+		},
+	})
```
dimuon, line 752: Does it make sense to test import?

PR #1969, opened 2026-03-23
feat: add ML calendar and calendar event resources

edsavage:
## Summary

- Add `elasticstack_elasticsearch_ml_calendar` resource for managing ML calendars and job associations
- Add `elasticstack_elasticsearch_ml_calendar_event` resource for managing individual scheduled events with RFC3339 times
- Both resources support import, follow Plugin Framework patterns, and include full test suites

## Changelog

Customer impact: enhancement
Summary: Add `elasticstack_elasticsearch_ml_calendar` and `elasticstack_elasticsearch_ml_calendar_event` resources for managing Elasticsearch ML calendars, scheduled events, and job associations in Terraform.

## Details

**Calendar resource**: Creates/deletes via PUT/DELETE calendar API. Updates job associations in-place by diffing `job_ids` and calling individual PutCalendarJob/DeleteCalendarJob endpoints. `description` requires replacement (PUT is create-only).

**Calendar event resource**: Creates via POST calendar events API, deletes via DELETE. No update API exists — all attributes use RequiresReplace. Server-generated `event_id` is discovered by diffing events before/after creation.

## Test plan

- [x] Unit tests: 29 non-acceptance `Test*` functions in `internal/elasticsearch/ml/calendar` and `internal/elasticsearch/ml/calendar_event` (models, read/create, wire conversions, validation, paging, composite IDs)
- [x] Acceptance tests: 12 `TestAcc*` functions in those packages
  - Calendar: `TestAccResourceMLCalendar`, `TestAccResourceMLCalendarNoJobs`, `TestAccResourceMLCalendarImport`, `TestAccResourceMLCalendar_validation_invalidCalendarIDRegex`, `TestAccResourceMLCalendar_validation_calendarIDTooLong`, `TestAccResourceMLCalendar_importWrongIDFormat`
  - Calendar event: `TestAccResourceMLCalendarEvent`, `TestAccResourceMLCalendarEvent_optionalSchedulingFields`, `TestAccResourceMLCalendarEventImport`, `TestAccResourceMLCalendarEvent_validation_endBeforeStart`, `TestAccResourceMLCalendarEvent_validation_invalidCalendarIDRegex`, `TestAccResourceMLCalendarEvent_importWrongIDFormat`
- [x] `make build` passes (including doc generation)

## Reviewer notes

- **Composite IDs:** Global `CompositeIDFromStr` / `CompositeIDFromStrFw` matches prior releases. Nested segments after the first `/` apply only to `CompositeIDFromStrForElasticsearch` / `…Fw` on the Elasticsearch Plugin Framework envelope `Read`/`Delete` (see `internal/clients/api_client.go` and the unreleased CHANGELOG entry).
- **`resource_envelope_test.go`:** Separate create/update stub functions exist only to assert distinct nil-callback paths (see file header comment).
- **`calendar_event` resource:** `newCalendarEventResource` documents why the envelope uses placeholder create callbacks with a custom `Create` override.
- **Elasticsearch 8.16:** Acceptance test `TestAccResourceMLCalendarEvent_optionalSchedulingFields` uses `versionutils.SkipIfUnsupported`; create-time guards in `create.go` reject optional POST fields on older clusters.

> [!NOTE]
> ### Add `elasticstack_elasticsearch_ml_calendar` and `elasticstack_elasticsearch_ml_calendar_event` Terraform resources
> - Adds two new Terraform resources for managing Elasticsearch ML anomaly detection calendars and their associated scheduled events.
> - The calendar resource supports create, read, update (job association reconciliation), delete, and import; the calendar event resource supports create, read, delete, and import (update requires replacement).
> - Calendar events are identified by a composite ID of `<cluster_uuid>/<calendar_id>/<event_id>`; event IDs are assigned by Elasticsearch and discovered by diffing pre/post-create event lists.
> - Time fields (`start_time`, `end_time`) use RFC3339 strings in Terraform state and are converted to/from epoch milliseconds for the API.
> - Both resources are registered in [provider/plugin_framework.go](https://github.com/elastic/terraform-provider-elasticstack/pull/1969/files#diff-5a66d72c3748bd201add43c9271e7b8dc29f28fd06eee792c837aa87e70a063b) and include acceptance tests and generated docs.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 5a294b1. 29 files reviewed, 5 issues evaluated, 0 issues filtered, 1 comment posted</summary> (Automatic summaries will resume when PR exits draft mode or review begins).
>
> ### 🗂️ Filtered Issues
>
> </details>
>
>


internal/elasticsearch/ml/calendar/update.go
```diff
@@ -0,0 +1,167 @@
...
+			if err != nil {
+				resp.Diagnostics.AddError("Failed to add job to calendar", fmt.Sprintf("Failed to add job %s to calendar %s: %s", jobID, calendarID, err.Error()))
+				return
+			}
+			defer addRes.Body.Close()
+
+			diags = diagutil.CheckErrorFromFW(addRes, fmt.Sprintf("Unable to add job %s to calendar %s", jobID, calendarID))
+			resp.Diagnostics.Append(diags...)
+			if resp.Diagnostics.HasError() {
+				return
```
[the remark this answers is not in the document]
edsavage: fix it for me

internal/elasticsearch/ml/calendar/resource.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (r *calendarResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	compID, diags := clients.CompositeIDFromStrFw(req.ID)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), req.ID)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("calendar_id"), compID.ResourceID)...)
```
tobio:
```suggestion
	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
```

`readCalendar` is only using the `id` attribute.

internal/elasticsearch/ml/calendar_event/schema.go
```diff
@@ -0,0 +1,85 @@
...
+			"calendar_id": schema.StringAttribute{
+				MarkdownDescription: "The identifier for the calendar that owns the event.",
+				Required:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+				Validators: []validator.String{
+					stringvalidator.LengthBetween(1, 64),
+				},
+			},
```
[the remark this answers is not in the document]
edsavage: Addressed: `calendar_id` now uses length + regex validators aligned with `elasticstack_elasticsearch_ml_calendar`.

internal/elasticsearch/ml/calendar_event/schema.go
```diff
@@ -0,0 +1,85 @@
...
+			"end_time": schema.StringAttribute{
+				MarkdownDescription: "The end time of the scheduled event in RFC 3339 format.",
+				Required:            true,
+				CustomType:          timetypes.RFC3339Type{},
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
```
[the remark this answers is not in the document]
edsavage: Addressed: `calendarEventResource.ValidateConfig` in `validate.go` calls `validateEventEndAfterStart`, which rejects plans where `end_time` is not strictly after `start_time` (RFC3339 parse + `time.After`).

internal/elasticsearch/ml/calendar_event/create.go
```diff
@@ -0,0 +1,111 @@
...
+
+	existingIDs := make(map[string]struct{})
+	preRes, err := typedClient.Ml.GetCalendarEvents(calendarID).Size(10000).Do(ctx)
+	if err == nil && preRes != nil {
+		for _, e := range preRes.Events {
+			if e.EventId != nil {
+				existingIDs[*e.EventId] = struct{}{}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
edsavage: Addressed: the pre-create listing uses `walkMLCalendarEventPages`, which returns diagnostics on non-404 list failures (no silent ignore). Event IDs are taken from `PostCalendarEvents` response when present; the snapshot diff + `calendarEventMatchesPlan` disambiguates multiple new events.

internal/elasticsearch/ml/calendar_event/models.go
```diff
@@ -0,0 +1,128 @@
...
+
+type PostCalendarEventsResponse struct {
+	Events []CalendarEventAPIModel `json:"events"`
+}
+
+type GetCalendarEventsResponse struct {
+	Count  int                     `json:"count"`
+	Events []CalendarEventAPIModel `json:"events"`
+}
+
```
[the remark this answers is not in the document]
edsavage: Addressed: unused `PostCalendarEventsRequest` / `GetCalendarEventsResponse` JSON wrapper types were removed; create/read use typed API structs (`postcalendarevents.NewRequest`, `types.CalendarEvent`, etc.).

internal/elasticsearch/ml/calendar_event/models.go
```diff
@@ -0,0 +1,115 @@
...
+	// The API returns epoch milliseconds as float64 (JSON number decoded via any).
+	startMillis, ok := apiModel.StartTime.(float64)
+	if !ok {
+		diags.AddError("Invalid start_time format", "Expected epoch milliseconds as a number from the API")
+		return diags
+	}
+	endMillis, ok := apiModel.EndTime.(float64)
+	if !ok {
+		diags.AddError("Invalid end_time format", "Expected epoch milliseconds as a number from the API")
+		return diags
```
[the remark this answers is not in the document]
edsavage: Addressed: `fromAPIModel` maps API times through `calendarEventAnyTimeToUnixMilli`, which handles float64/int/json.Number/RFC3339 strings and `types.DateTime` (delegating to the typed-client path where needed).

internal/elasticsearch/ml/calendar_event/read.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+		diags.Append(state.fromAPIModel(ctx, calendarEventTypedToAPI(&event))...)
+		if diags.HasError() {
+			return state, false, diags
+		}
+		tflog.Debug(ctx, fmt.Sprintf("Successfully read ML calendar event %s from calendar: %s", eventID, calendarID))
+		return state, true, diags
+	}
+
+	return state, false, nil
```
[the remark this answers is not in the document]
edsavage: Addressed: `readCalendarEvent` walks pages via `walkMLCalendarEventPages` (1000 events per page) until the requested `event_id` is found or the calendar is exhausted—no single 10k cap.

internal/elasticsearch/ml/calendar_event/create.go
```diff
@@ -0,0 +1,184 @@
...
+
+	existingIDs := make(map[string]struct{})
+	preRes, err := typedClient.Ml.GetCalendarEvents(calendarID).Size(10000).Do(ctx)
+	if err != nil {
+		diags.AddError(
+			"Failed to list calendar events before create",
+			fmt.Sprintf("Cannot snapshot existing event IDs for calendar %s — %s", calendarID, err.Error()),
+		)
+		return plan, diags
+	}
```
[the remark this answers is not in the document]
edsavage: Addressed: the same paged walk is used for the pre-create ID snapshot and the post-create candidate scan, so calendars with >10k events are handled without truncating to a single page.

internal/elasticsearch/ml/calendar/models.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (m *TFModel) fromAPIModel(ctx context.Context, apiModel *APIModel) fwdiags.Diagnostics {
+	var diags fwdiags.Diagnostics
+
+	m.CalendarID = types.StringValue(apiModel.CalendarID)
+
+	if apiModel.Description != "" {
+		m.Description = types.StringValue(apiModel.Description)
+	} else {
```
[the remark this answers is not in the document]
edsavage: Addressed: `TFModel.fromAPIModel` preserves empty descriptions with `types.StringValue(apiModel.Description)`, and the calendar schema marks `description` optional+computed with `stringdefault.StaticString("")` so `description = ""` does not perpetual-drift against the API.

internal/elasticsearch/ml/calendar_event/read.go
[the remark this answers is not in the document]
edsavage: In read.go, now walkMLCalendarEventPages until match or exhaustion.

internal/elasticsearch/ml/calendar_event/create.go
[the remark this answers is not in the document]
edsavage: Use same pager for pre-create IDs and post-create candidates.

internal/elasticsearch/ml/calendar/models.go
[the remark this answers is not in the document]
edsavage: calendar/models.go: types.StringValue(apiModel.Description); calendar/schema.go — optional+computed + stringdefault.StaticString("").

internal/clients/api_client.go
```diff
@@ -59,11 +59,37 @@ func CompositeIDFromStr(id string) (*CompositeID, diag.Diagnostics) {
 		diags
 }
 
+// CompositeIDFromStrForElasticsearch parses an ID as <cluster_uuid>/<resource_identifier> by
+// splitting only on the first slash, so resource_identifier may contain additional slashes.
+// Some Elasticsearch resources (for example ML calendar events) use a resource segment of the
+// form "<calendar_id>/<event_id>".
+func CompositeIDFromStrForElasticsearch(id string) (*CompositeID, diag.Diagnostics) {
```
tobio: This is backwards compatible with the current format. IMO we should just add this change to `CompositeIDFromStr`.

internal/elasticsearch/ml/calendar/create.go
```diff
@@ -0,0 +1,74 @@
...
+		return plan, diags
+	}
+
+	body, err := json.Marshal(apiModel)
+	if err != nil {
+		diags.AddError("Failed to marshal calendar configuration", err.Error())
+		return plan, diags
+	}
+
+	_, err = typedClient.Ml.PutCalendar(calendarID).Raw(bytes.NewReader(body)).Do(ctx)
```
tobio: Can we have `toAPICreateModel` return `putcalendar.Request` directly and just do: 

```suggestion
	_, err = typedClient.Ml.PutCalendar(calendarID).Request(apiModel).Do(ctx)
```

internal/elasticsearch/ml/calendar/models.go
```diff
@@ -0,0 +1,95 @@
...
+		var jobIDs []string
+		d := m.JobIDs.ElementsAs(ctx, &jobIDs, false)
+		diags.Append(d...)
+		apiModel.JobIDs = jobIDs
+	}
+
+	return apiModel, diags
+}
+
+func (m *TFModel) fromAPIModel(ctx context.Context, apiModel *APIModel) fwdiags.Diagnostics {
```
tobio: Lets just use the client model directly. 

```suggestion
func (m *TFModel) fromAPIModel(ctx context.Context, apiModel types.Calendar) fwdiags.Diagnostics {
```

internal/elasticsearch/ml/calendar/resource.go
tobio:
```suggestion
	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
```

internal/elasticsearch/ml/calendar/schema.go
```diff
@@ -0,0 +1,77 @@
...
+			"description": schema.StringAttribute{
+				MarkdownDescription: "A description of the calendar.",
+				Optional:            true,
+				Computed:            true,
+				Default:             stringdefault.StaticString(""),
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"job_ids": schema.SetAttribute{
```
tobio: https://github.com/elastic/terraform-provider-elasticstack/pull/2933 allows attaching a job to a calendar, but this resource manages the _entire_ job list unconditionally. 

Combining these two resources will cause state consistency issues. 

We can allow both paths, with clear documentation that users can use only one at a time. That is either manage all jobs attached to the calendar via this resource, or manage individual jobs via the other one, not both. 

I think we also want to change the read logic to only populate `job_ids` if the user has explicitly configured it.

internal/elasticsearch/ml/calendar/update.go
```diff
@@ -0,0 +1,137 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	var state TFModel
+	diags = req.State.Get(ctx, &state)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
tobio: We're not actually using the state value, we're reading the current calendar from the API. If we're not using state we should convert this to an updateFunc used by the envelope.

internal/elasticsearch/ml/calendar/update.go
```diff
@@ -0,0 +1,137 @@
...
+
+	for _, jobID := range currentCalendar.JobIDs {
+		if _, exists := planJobIDSet[jobID]; !exists {
+			_, err := typedClient.Ml.DeleteCalendarJob(calendarID, jobID).Do(ctx)
+			if err != nil {
+				resp.Diagnostics.AddError("Failed to remove job from calendar", fmt.Sprintf("Failed to remove job %s from calendar %s: %s", jobID, calendarID, err.Error()))
+				return
+			}
+		}
+	}
```
tobio: This makes me wonder if supporting `job_ids` on a calendar is worth it at all.

internal/elasticsearch/ml/calendar_event/create.go
```diff
@@ -0,0 +1,176 @@
...
+	tflog.Debug(ctx, fmt.Sprintf("Creating ML calendar event for calendar: %s", calendarID))
+
+	typedClient, err := client.GetESClient()
+	if err != nil {
+		diags.AddError("Failed to get Elasticsearch client", err.Error())
+		return plan, diags
+	}
+
+	existingIDs := make(map[string]struct{})
+	diags.Append(walkMLCalendarEventPages(ctx, typedClient, calendarID, func(events []calendarEventWire) bool {
```
tobio: This is potentially quite costly. Under what conditions does ES not return an event ID on creation?

internal/elasticsearch/ml/calendar_event/create.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+		return false
+	})...)
+	if diags.HasError() {
+		return plan, diags
+	}
+
+	postBody := struct {
+		Events []calendarEventWire `json:"events"`
+	}{Events: []calendarEventWire{planWire}}
```
tobio: Just use `postcalendarevents.Request`

internal/elasticsearch/ml/calendar_event/create.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	if postHTTP.StatusCode >= 300 {
+		diags.AddError("Failed to create ML calendar event", fmt.Sprintf("Unable to create ML calendar event for calendar %s — status %d: %s", calendarID, postHTTP.StatusCode, string(postBodyBytes)))
+		return plan, diags
+	}
+
+	var eventID string
+	var postWire struct {
+		Events []calendarEventWire `json:"events"`
+	}
```
tobio:
```suggestion
	res, err := typedClient.Ml.PostCalendarEvents(calendarID).Raw(bytes.NewReader(body)).Do(ctx)
	if err != nil {
		diags.AddError("Failed to create ML calendar event", fmt.Sprintf("Unable to create ML calendar event for calendar %s — %s", calendarID, err.Error()))
		return plan, diags
	}
```

`postcalendarevents.Response` is identical to what we're parsing here. Any need for handling it manually?

internal/elasticsearch/ml/calendar_event/read.go
```diff
@@ -0,0 +1,94 @@
...
+			diags.Append(state.fromAPIModel(ctx, apiModel)...)
+			if diags.HasError() {
+				return true
+			}
+			found = true
+			tflog.Debug(ctx, fmt.Sprintf("Successfully read ML calendar event %s from calendar: %s", eventID, calendarID))
+			return true
+		}
+		return false
+	})
```
tobio: In all cases when we're reading calendar events we're interested in a very specific time frame, and . Can we reduce the search space by adding `from/to` query params around the timeframe for this specific event? 

I think that might be an optimistic optimisation and we fall back to a full scan if we fail to find the target event (it may have been updated out of band and no longer be in the right timespan).

internal/elasticsearch/ml/calendar_event/resource.go
```diff
@@ -0,0 +1,124 @@
...
+			phUpdate,
+		),
+	}
+}
+
+func NewCalendarEventResource() resource.Resource {
+	return newCalendarEventResource()
+}
+
+func (r *calendarEventResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
```
tobio, line 62: I don't think this is doing anything fancy, can we just pass `createCalendarEvent` to the `NewElasticsearchResource`?

internal/elasticsearch/ml/calendar_event/resource.go
```diff
@@ -0,0 +1,124 @@
...
+
+	calendarID, eventID, splitDiags := splitCalendarEventResourcePath(compID.ResourceID)
+	resp.Diagnostics.Append(splitDiags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), req.ID)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("calendar_id"), calendarID)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("event_id"), eventID)...)
```
tobio:
```suggestion
	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
```

internal/elasticsearch/ml/calendar_event/wire_events.go
```diff
@@ -0,0 +1,290 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+
+	fwdiags "github.com/hashicorp/terraform-plugin-framework/diag"
+)
+
+// calendarEventWire is the JSON shape for ML calendar events (POST body and GET list responses).
+// The go-elasticsearch typed CalendarEvent struct does not yet include skip_result, skip_model_update,
+// or force_time_shift, so we decode list/post payloads with this type.
```
tobio: 👍

internal/elasticsearch/ml/calendar_event/create.go
edsavage: We’ve switched the happy path to `PostCalendarEvents(...).Request(req).Do(ctx)`, so we use `postcalendarevents.Response` there instead of `Perform` + reading/decoding the body by hand.

We still convert each returned `types.CalendarEvent` into our existing `calendarEventWire` shape so we can reuse the same matching / `event-id` logic as the raw-JSON path (and the list fallback), without duplicating that code for two types.

When we have to POST `skip_result` / `skip_model_update` / `force_time_shift`, `types.CalendarEvent` still doesn’t carry those fields in go-elasticsearch v8.19.5, so that branch keeps Raw + manual read/decode of the body. Once the typed `CalendarEvent` (and `Request`) include those fields, we can drop the raw path and rely on `Do()` / `Response` end-to-end.

internal/elasticsearch/ml/calendar_event/resource.go
edsavage: Not yet: the envelope requires a known GetResourceID() before create and passes that same id into post-create read. Here event_id is unknown until after POST, and read needs calendar_id/event_id. Custom Create avoids that prelude and reads using the ids returned from createCalendarEvent.

internal/elasticsearch/ml/calendar_event/create.go
```diff
@@ -0,0 +1,245 @@
...
+		if diags.HasError() {
+			return plan, diags
+		}
+
+		req := postcalendarevents.NewRequest()
+		req.Events = []estypes.CalendarEvent{ev}
+
+		resp, err := typedClient.Ml.PostCalendarEvents(calendarID).Request(req).Do(ctx)
+		if err != nil {
+			if esErr, ok := errors.AsType[*estypes.ElasticsearchError](err); ok {
```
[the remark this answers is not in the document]
edsavage: The provider targets Go 1.26 and errors.AsType is valid there, ignoring this suggestion.

internal/elasticsearch/ml/calendar_event/delete.go
```diff
@@ -0,0 +1,60 @@
...
+
+	typedClient, err := client.GetESClient()
+	if err != nil {
+		diags.AddError("Failed to get Elasticsearch client", err.Error())
+		return diags
+	}
+
+	_, err = typedClient.Ml.DeleteCalendarEvent(calendarID, eventID).Do(ctx)
+	if err != nil {
+		if esErr, ok := errors.AsType[*types.ElasticsearchError](err); ok && esErr.Status == 404 {
```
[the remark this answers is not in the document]
edsavage: The provider targets Go 1.26 and errors.AsType is valid there, ignoring this suggestion.

tobio:
This is still unresolved https://github.com/elastic/terraform-provider-elasticstack/pull/1969#discussion_r3238362484

Exactly what steps lead Elasticsearch to not return event IDs for newly created calendar entries? I imagine it's deterministic. We're adding a lot of complexity into this resource to handle missing IDs, that needs to be justified.

edsavage:
> This is still unresolved [#1969 (comment)](https://github.com/elastic/terraform-provider-elasticstack/pull/1969#discussion_r3238362484)
> 
> Exactly what steps lead Elasticsearch to not return event IDs for newly created calendar entries? I imagine it's deterministic. We're adding a lot of complexity into this resource to handle missing IDs, that needs to be justified.

Yeah, it’s deterministic. For normal creates (where we never send event_id), POST always omits it — not intermittently. POST indexes the event, then returns the request payload unchanged. The POST parser doesn’t accept event_id on the request at all, so the in-memory objects never have event_id set before the response is built. The id is assigned at 
index time and only shows up on GET, from the .ml-meta document _id.

I've confirmed this live on 8.0.1, 8.16.6, and 9.0.8: The POST response has no `event_id`, a windowed GET returns it. You also cannot _set_ `event_id` on POST - all three versions return `unknown field [event_id]` (it's not in `ScheduledEvent`'s strict request parser).

Terraform needs `event_id` (and composite id) immediately after create; POST won't provide it. The fallback is a targeted GET/list to match the event we just created, not a workaround for flaky API behavior.

The pre-create full-calendar snapshot is gone. We are not listing the whole calendar before create anymore. Listing is post-create, and only as a fallback.

internal/elasticsearch/ml/calendar_event/create.go
```diff
@@ -0,0 +1,230 @@
...
+			}
+		}
+	}
+	if eventID == "" && len(postWire) == 1 {
+		eventID = calendarEventWireEventID(&postWire[0])
+	}
+
+	if eventID == "" {
+		var matches []calendarEventWire
+		diags.Append(walkMLCalendarEventPages(ctx, typedClient, calendarID, func(events []calendarEventWire) bool {
```
tobio:
```suggestion
		diags.Append(walkMLCalendarEventPagesWithWindow(ctx, typedClient, calendarID, func(events []calendarEventWire) bool {
```

We should be able to just look at the event time window here right? Is there any case where we _wouldn't_ have a specific time window to search? It feels like we should be able to remove `walkMLCalendarEventPages` and only support searching the event window.

internal/elasticsearch/ml/calendar_event/create.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+		if sv.LessThan(mlCalendarEventOptionalAPIFieldsMinElasticsearch) {
+			diags.AddError(
+				"ML calendar event optional scheduling fields not supported",
+				fmt.Sprintf("skip_result, skip_model_update, and force_time_shift require Elasticsearch %s or newer; "+
+					"this cluster is %s. Omit these arguments or upgrade Elasticsearch.",
+					mlCalendarEventOptionalAPIFieldsMinElasticsearch.String(), sv.String()),
+			)
+			return plan, diags
+		}
```
tobio: This won't work correctly in serverless (it unhelpfully returns 8.11), use `EnforceMinVersion` instead to check both flavor and version.

internal/clients/api_client.go
edsavage: Addressed in 633ce677: `CompositeIDFromStr` now splits only on the first `/` (with legacy empty-cluster support). `CompositeIDFromStrForElasticsearch` is an alias. Added `api_client_composite_id_test.go`.

internal/elasticsearch/ml/calendar/create.go
edsavage: Addressed in 633ce677: create uses `typedClient.Ml.PutCalendar(calendarID).Request(newPutCalendarRequestFromTFModel(plan)).Do(ctx)` (returns `putcalendar.Request`).

internal/elasticsearch/ml/calendar/models.go
edsavage: Addressed in 633ce677: read maps `types.Calendar` via `applyTypedCalendarToTFModel` (no separate APIModel wrapper).

internal/elasticsearch/ml/calendar/resource.go
edsavage: Addressed in 633ce677: `ImportStatePassthroughID` for `id`, then set `calendar_id` from the parsed resource segment.

internal/elasticsearch/ml/calendar/schema.go
edsavage: Addressed on this branch: `job_ids` was removed from the calendar resource; job attachment is `elasticstack_elasticsearch_ml_calendar_job` (see schema/docs). That avoids conflicting with PR #2933.

internal/elasticsearch/ml/calendar/update.go
edsavage: Addressed: calendar `updateCalendar` is a no-op; job reconciliation lives on `calendar_job`.

internal/elasticsearch/ml/calendar_event/create.go
edsavage: ES usually returns `event_id` on POST; the list walk is only when the response omits it or matching is ambiguous. Create now uses `walkMLCalendarEventPagesWithWindow` scoped to the planned start/end (633ce677), not a full-calendar scan.

internal/elasticsearch/ml/calendar_event/create.go
edsavage: Happy path uses `postcalendarevents.NewRequest` + `.Request(req).Do(ctx)`. Raw POST remains only when optional scheduling fields are set (not yet on `types.CalendarEvent` in go-elasticsearch v8.19.5).

internal/elasticsearch/ml/calendar_event/create.go
edsavage: Same as above: typed `.Do(ctx)` on the happy path; manual body decode only for the raw optional-fields branch.

internal/elasticsearch/ml/calendar_event/read.go
edsavage: Read already uses start/end window first, then full scan fallback (out-of-band time changes). Create fallback listing now uses the plan window only (633ce677).

internal/elasticsearch/ml/calendar_event/resource.go
edsavage: Still a custom `Create`: `event_id` is unknown until after POST, so the envelope cannot pre-resolve the read id (see comment on the resource).

internal/elasticsearch/ml/calendar_event/resource.go
edsavage: Addressed in 633ce677: `ImportStatePassthroughID` + set `calendar_id` / `event_id` from parsed composite resource segment.

internal/elasticsearch/ml/calendar_event/create.go
edsavage: Addressed in 633ce677: post-create ID discovery uses `walkMLCalendarEventPagesWithWindow` with `calendarEventWireWindowRFC3339(planWire)`.

internal/elasticsearch/ml/calendar_event/create.go
edsavage: Addressed in 633ce677: optional-field gate uses `client.EnforceMinVersion` instead of `ServerVersion` (serverless-safe).

tobio (approved):

PR #2062, opened 2026-03-30
Add spec proposal for elastic defend

tobio:
Related to https://github.com/elastic/terraform-provider-elasticstack/issues/601

> [!NOTE]
> ### Add spec proposal for a dedicated Elastic Defend integration policy resource
> - Adds design, proposal, requirements, and task documents under [openspec/changes/add-fleet-elastic-defend-integration-policy/](https://github.com/elastic/terraform-provider-elasticstack/pull/2062/files#diff-0e8be7ef8e4dcd61e477ce9c8640bbdd57549b7d138d19649be8f9faaeef701f) to define a new `fleet-elastic-defend-integration-policy` Terraform resource.
> - The resource fixes the package name to `endpoint`, uses typed-only input encoding, and constrains its schema to Defend-owned configuration fields.
> - Defines a two-phase create flow (bootstrap then finalize) and treats server-managed fields (`artifact_manifest`, `version`) as opaque internal state.
> - Outlines required changes to the generated Fleet client (`kbapi`) to support typed inputs, version handling, and format selection in Fleet client helpers.
>
> 
> #### Changes since #2062 opened
>
> - Replaced generic OS policy schema with distinct, explicitly enumerated schemas for `windows`, `mac`, and `linux` operating systems, each containing typed nested attributes for sub-blocks including `events`, `malware`, `ransomware`, `memory_protection`, `behavior_protection`, `popup`, `logging`, `antivirus_registration`, and `attack_surface_reduction` [f89fa5e]
> - Updated attribute reference path for `preset` from `integration_config.value.endpointConfig.preset` to `config.integration_config.value.endpointConfig.preset` in documentation comments [f89fa5e]
> - Fixed header formatting by removing leading space from top-level title line in tasks documentation [f89fa5e]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3ccde08.</sup>
> 
>
>

### DIFF
--- a/openspec/changes/add-fleet-elastic-defend-integration-policy/design.md
+++ b/openspec/changes/add-fleet-elastic-defend-integration-policy/design.md
@@ -51,29 +51,157 @@ resource "elasticstack_fleet_elastic_defend_integration_policy" "example" {
   integration_version = <required, string>
   space_ids           = <optional+computed, set(string)>
 
-  preset = <optional, string> # maps to integration_config.value.endpointConfig.preset
+  preset = <optional, string> # maps to config.integration_config.value.endpointConfig.preset
 
   policy = <required, single nested attribute> {
-    windows = <optional, single nested attribute>
-    mac     = <optional, single nested attribute>
-    linux   = <optional, single nested attribute>
+    windows = <optional, single nested attribute> {
+      events = <optional, single nested attribute> {
+        # Windows-specific event collection flags
+        process             = <optional, bool>
+        network             = <optional, bool>
+        file                = <optional, bool>
+        dll_and_driver_load = <optional, bool>
+        dns                 = <optional, bool>
+        registry            = <optional, bool>
+        security            = <optional, bool>
+        authentication      = <optional, bool>
+      }
+      malware = <optional, single nested attribute> {
+        mode          = <optional, string>  # "off" | "detect" | "prevent"
+        blocklist     = <optional, bool>
+        on_write_scan = <optional, bool>
+        notify_user   = <optional, bool>
+      }
+      ransomware = <optional, single nested attribute> {
+        mode      = <optional, string>  # "off" | "detect" | "prevent"
+        supported = <optional, bool>
+      }
+      memory_protection = <optional, single nested attribute> {
+        mode      = <optional, string>  # "off" | "detect" | "prevent"
+        supported = <optional, bool>
+      }
+      behavior_protection = <optional, single nested attribute> {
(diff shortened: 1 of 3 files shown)


openspec/changes/add-fleet-elastic-defend-integration-policy/design.md
```diff
@@ -0,0 +1,162 @@
...
+
+- `events`
+- `malware`
+- `ransomware` (Windows only)
+- `memory_protection`
+- `behavior_protection`
+- `popup`
+- `logging`
+- `antivirus_registration` (Windows only)
+- `attack_surface_reduction` (Windows only)
```
dimuon: The design lists OS-specific attributes (`ransomware`, `antivirus_registration`, `attack_surface_reduction` for Windows; `session_data`, `tty_io` for Linux) but does not specify whether each OS block (`windows`, `mac`, `linux`) uses a distinct nested attribute schema containing only its applicable fields, or a shared schema across all three. This matters because:

- **Distinct schemas per OS** make invalid combinations (e.g., `policy.linux.ransomware`) structurally impossible at plan time
- **A shared schema** would require either custom validation to reject inapplicable attributes, or silently ignoring them
- The choice affects documentation (three different attribute reference tables vs. one with footnotes)

The spec should state which approach is intended before implementation begins.

dimuon:
### integration_version change behavior is unspecified (REQ-010 gap)

The spec documents the two-phase create flow (REQ-008/REQ-009) and the normal settings update (REQ-010), but doesn't address what happens when a user changes `integration_version` on an existing resource.

Fleet exposes a **dedicated upgrade endpoint** (`POST /api/fleet/package_policies/upgrade`) that is separate from the normal `PUT /api/fleet/package_policies/{id}`. The [Elastic Defend API docs](https://www.elastic.co/docs/solutions/security/configure-elastic-defend/create-an-elastic-defend-policy-using-api) show the Step 3 `PUT` call accepts `package.version` in the body, which suggests the normal update path *might* work — but for Elastic Defend specifically, the `artifact_manifest` complicates things: does it get regenerated by the server on a version change, making the stored opaque state stale?

Could we add a clarification to REQ-010 (or a new requirement) stating whether:
- Version changes use the normal update path (PUT with new `package.version`), or
- Version changes require calling the dedicated upgrade endpoint first, then applying settings

Even a one-liner like "version changes SHALL be handled via the normal update path" would close the gap for an implementer.

PR #2086, opened 2026-03-30
add tamper protection option to agent policy ressource

### LINKED ISSUE #2085: [Feature] Add tamper protection option to fleet policy ressource
So far, when a fleet policy is managed through the provider the flag is_protected /tamper protection can't me managed by the provider as the option is missing

SimonKoetting:
closes #2085 

## Changelog
Customer impact: enhancement
Summary: add tamper protection option to agent policy ressource

> [!NOTE]
> ### Add `is_protected` tamper protection option to `elasticstack_fleet_agent_policy` resource
> - Adds an optional `is_protected` boolean attribute to the Fleet agent policy resource, mapping to the Fleet API `is_protected` field. Requires Elastic Stack 8.10.0 or later.
> - Because Fleet requires Elastic Defend to be attached before tamper protection can be enabled, `ModifyPlan` rejects plans that attempt to create a new policy with `is_protected = true`, instructing users to create with `false` first, attach Elastic Defend, then re-apply.
> - After create or update requests with `is_protected = true`, the provider polls the Fleet API until the flag reflects `true`; if it stays `false`, the operation fails with a descriptive error about prerequisites.
> - Risk: Setting `is_protected = true` on an unsupported stack version or without Elastic Defend attached will cause apply to fail, not plan-time.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c83a99a.</sup>
> 
>
>


internal/fleet/agentpolicy/create.go
```diff
@@ -102,6 +123,20 @@ func (r *agentPolicyResource) Create(ctx context.Context, req resource.CreateReq
 		return
 	}
 
+	if typeutils.IsKnown(planWantsTamperProtection) && planWantsTamperProtection.ValueBool() &&
+		typeutils.IsKnown(planModel.IsProtected) && !planModel.IsProtected.ValueBool() {
+		resp.Diagnostics.AddError(
+			"Fleet API did not enable tamper protection",
+			"The agent policy was saved but is_protected is still false. "+
```
SimonKoetting, line 168: Not too sure if this is the best approach here, but as tamper protection can only be set as soon as Defend is added to the policy, this is the only way I imagine handling these scenarios and informing the user.

tobio: Wouldn't this *always* fail on creation since it's not possible for Defend to be applied to the agent policy yet? 

Should we just error if `is_protected = true` on creation instead rather than apply a plan that we know will fail?

internal/fleet/agentpolicy/acc_test.go
```diff
@@ -1001,3 +1004,35 @@ func TestAccResourceAgentPolicyWithRestrictedUser(t *testing.T) {
...
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { acctest.PreCheck(t) },
+		CheckDestroy: checkResourceAgentPolicyDestroy,
+		Steps: []resource.TestStep{
+			{
+				ProtoV6ProviderFactories: acctest.Providers,
+				SkipFunc: func() (bool, error) {
+					// Tamper protection requires specific licensing/capabilities and can be ignored by Fleet
+					// even on supported versions. Default to skipping unless explicitly enabled.
```
tobio, line 1019: What does this mean specifically? How do we have this test actually verify that tamper protection works within CI?

tobio (approved):

PR #2111, opened 2026-03-31
Add Agent Builder tool resource and data source

tehbooom:
Related to #1619 

> [!NOTE]
> ### Add `elasticstack_kibana_agentbuilder_tool` resource and data source
> - Adds a new `elasticstack_kibana_agentbuilder_tool` Terraform resource supporting full CRUD for Kibana Agent Builder tools (types: `esql`, `index_search`, `workflow`, `mcp`), with a composite `<tool_id>/<space_id>` import ID and version gating at Kibana 9.3.0+.
> - Adds a matching `elasticstack_kibana_agentbuilder_tool` data source that reads a tool by ID, with an optional `include_workflow` flag to fetch linked workflow details (requires Kibana 9.4.0+).
> - Renames the existing `elasticstack_kibana_agentbuilder_export_workflow` data source to `elasticstack_kibana_agentbuilder_workflow` and moves its implementation from `internal/kibana/exportagentbuilder/workflow` to `internal/kibana/agentbuilderworkflow`; the old name is no longer registered in the provider.
> - Extracts a shared `acctest.PreCheckWithWorkflowsEnabled` helper that version-gates tests and enables the Kibana workflows UI setting before test execution.
> - Behavioral Change: `elasticstack_kibana_agentbuilder_export_workflow` is removed from the provider and replaced by `elasticstack_kibana_agentbuilder_workflow`; any existing configurations using the old name will break.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0f7322d.</sup>
> 
>


internal/kibana/agentbuildertool/acc_test.go
```diff
@@ -0,0 +1,235 @@
...
+		t.Skipf("Skipping test: server version %s is below minimum %s", serverVersion, minKibanaAgentBuilderAPIVersion)
+	}
+
+	kibanaClient, err := client.GetKibanaOapiClient()
+	if err != nil {
+		t.Fatalf("Failed to get Kibana client: %v", err)
+	}
+
+	// Try the internal settings API endpoint
+	settingsURL := fmt.Sprintf("%s/internal/kibana/settings/workflows:ui:enabled", kibanaClient.URL)
```
tobio: We're duplicating this setup step across all the agent build tests. Can we extract this to a shared helper somewhere.

internal/kibana/agentbuildertool/schema.go
```diff
@@ -0,0 +1,97 @@
...
+				},
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"description": schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "The tool description.",
+			},
+			"tags": schema.ListAttribute{
```
tobio:
```suggestion
			"tags": schema.SetAttribute{
```

?

internal/kibana/agentbuildertool/schema.go
```diff
@@ -0,0 +1,97 @@
...
+				ElementType:         types.StringType,
+				Optional:            true,
+				MarkdownDescription: "List of tags for the tool.",
+			},
+			"configuration": schema.StringAttribute{
+				Required:            true,
```
tobio, line 93:
```suggestion
				Required:            true,
				CustomType: jsontypes.Normalized{},
```

I imagine this should use a normalised type.

internal/kibana/exportagentbuilder/tool/schema.go
```diff
@@ -0,0 +1,82 @@
...
+			},
+			"workflow_id": schema.StringAttribute{
+				Description: "The ID of the referenced workflow. Only populated when `include_workflow` is true.",
+				Computed:    true,
+			},
+			"workflow_configuration_yaml": schema.StringAttribute{
+				Description: "The YAML configuration of the referenced workflow. Only populated when `include_workflow` is true.",
+				Computed:    true,
+				CustomType:  customtypes.NormalizedYamlType{},
+			},
```
tobio: I feel like we're adding additional complexity here with essentially no customer benefit. We already provide a workflow export data source, which customers could tie into this and get essentially the same result. 

What's the benefit in adding these fields here rather than relying on the existing workflow export?

internal/kibana/exportagentbuilder/tool/data_source.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// DataSource is the data source implementation.
+type DataSource struct {
+	client *clients.APIClient
+}
+
+// Metadata returns the data source type name.
+func (d *DataSource) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_kibana_agentbuilder_export_tool"
```
tobio: So, I know this is consistent with `export_workflow`, but there's nothing really that makes these an `export` rather than just a standard TF data source (i.e they're just supplying the entity values as outputs, no special export format). 

WDYT about these just being `_kibana_agentbuilder_tool` and `_kibana_agentbuilder_workflow`?

internal/kibana/exportagentbuilder/tool/models.go
```diff
@@ -0,0 +1,38 @@
...
+
+package tool
+
+import (
+	"github.com/elastic/terraform-provider-elasticstack/internal/utils/customtypes"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// dataSourceModel maps the data source schema data.
+type dataSourceModel struct {
```
tobio: Similarly to the export comment above, this type and the read code path are essentially duplicated from the resource. I wonder if it makes sense to move this data source code alongside the resource and re-use models/read paths more.

internal/kibana/exportagentbuilder/tool/schema.go
tehbooom: IMO, as someone who has been writing tools, agents, and workflows a lot lately. It would be very annoying to hit the API to grab the workflow ID or go to the workflow and grab just the ID from the URL.

Tools and agents have very easy IDs for humans to read and remember while workflows do not since they use UUIDv4. If i wanted to export 50 tools, 5 agents, and 20 something workflows (almost all workflows have tools), then this export makes it so much faster. This example is exactly what I've done in a real scenario and without this functionality I would have to export all tools and then all workflows where if we did have this i would export all tools, then the 5 other workflows I know are not tied to a tool.

internal/kibana/exportagentbuilder/tool/data_source.go
tehbooom: I agree. Ill change the the tool resource and data source. Ill create another PR for workflow to follow the same pattern.

internal/kibana/agentbuildertool/acc_test.go
tehbooom: Added a shared helper function to `internal/acctest/acctest.go`

internal/kibana/exportagentbuilder/tool/data_source.go
tehbooom: Actually does it make more sense to update both in this PR?

internal/kibana/exportagentbuilder/tool/schema.go
tehbooom: This will also follow the same flow as when we retrieve an Agent. Users will want to export an Agent and have all its dependencies instead of having to grab all of them individually. So when I export an Agent I wants its tool and its workflows to then import elsewhere

tobio: > If i wanted to export 50 tools, 5 agents, and 20 something workflows (almost all workflows have tools), then this export makes it so much faster.

I don't quite follow, why including the _workflow_ on the _tool_ helps in this scenario. I'd have 50 tools, but only 20 workflows so there's some duplication there I need to deal with. 

> without this functionality I would have to export all tools and then all workflows where if we did have this i would export all tools, then the 5 other workflows I know are not tied to a tool.

So, naively I would expect that I'd do something like this

```hcl
data "tool" "abc" {
  id = "blah"
}

data "workflow" "abc" {
  id = data.tool.abc.configuration["workflow_id"]
}

# Or, actually de-dupe the workflows 
locals {
  workflows = distinct([
    data.tool.abc.configuration.workflow_id, 
    data.tool.def.configuration.workflow_id,
    "workflow_without_tool_1",
  ]) 
}

data workflow "workflows" {
  count = len(locals.workflows)
  id = locals.workflows[count.value]
}
```

To me, this feels a little backwards (tool -> workflow) and I wonder if we should either (or do some combination of):
1. Change these data sources to include all tools/workflows rather than a specific tool/workflow and let the user filter within HCL
```hcl
data "tools" "tools" {}
locals {
  workflows_with_tools = distinct([for t in data.tools.tools.values : t.configuration["workflow_id]])
}
```
2. Include tools in workflows, rather than workflows in tools. Reversing this would mean that I have a data source for the agents (or workflows) I care about that then include the dependencies (i.e workflows/tools) that I need to make them work elsewhere. 

Sorry for the essay, the current structure doesn't make much sense to me, but clearly does for you so I feel like I'm missing some crucial aspect :)

tehbooom: > I don't quite follow, why including the workflow on the tool helps in this scenario. I'd have 50 tools, but only 20 workflows so there's some duplication there I need to deal with.

There is no duplication here. My example is out of the 50 tools 15 are workflows. So i go ahead and export all 50 tools + the 15 workflows. Now when I go to export more workflows I only need to get the UUID of 5 workflows instead of clicking through the UI or hitting the API and doing a copy paste.

> Change these data sources to include all tools/workflows rather than a specific tool/workflow and let the user filter within HCL

One can create many tools but never create a workflow. Although they do work together they are not always combined. 

> Include tools in workflows, rather than workflows in tools.

You cannot create or export a workflow with a tool. The workflow has no idea about the tool but the tool has the workflow ID. When creating something with agent builder this is the flow Agent -> tool.id -> workflow.uuid. You can never go backwards.

Think of it like this. Agents = data views, tools = dashboards, and workflows = lens visualizations. All can be independent of one another but if you were to export a dashboard it would be great to export the lens visual as well instead of having to click through saved objects to find that specific UUID to export right? I hope that clears things up.

There is also [this conversation](https://elastic.slack.com/archives/C08LX7YSHU2/p1767992044336379) internally.

PR #2129, opened 2026-04-02
Update dataview without recreate via space API

Onder7994:
This PR adds the ability to update the namespaces field of a Kibana data view in-place, without forcing resource recreation.

We use Kibana alerting rules in our infrastructure, and some of those rules reference data views by ID. Previously, changing the namespaces field on a data view would cause Terraform to destroy and recreate it with a new ID, which broke any rules that depended on the original data view ID.

In the Kibana UI, it is possible to add or remove spaces from a data view without recreating it. This PR brings that same behavior to the Terraform provider.

Instead of triggering a replacement, the provider now calls the Kibana Spaces API (POST /api/spaces/_update_objects_spaces) during updates to diff the old and new namespace lists and apply only the delta — adding new spaces and removing ones that are no longer present.
Verification

Tested locally using a custom-built provider binary configured via .terraformrc dev overrides. After applying a namespaces change, verified via GET kbn:/api/data_views/data_view/<dataview_id> that:
1. The data view ID remains unchanged
2. The namespaces field reflects the updated value

> [!NOTE]
> ### Update `elasticstack_kibana_data_view` namespaces in-place via Kibana Spaces API
> - Previously, changes to `data_view.namespaces` forced resource replacement. The `RequiresReplace` plan modifier is removed from the `namespaces` attribute in [schema.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2129/files#diff-5ee27c86734848be031c141b124701b5ec7a03a803745b94f806d98281fc2ae0).
> - A new `UpdateDataViewNamespaces` function in [data_views_spaces.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2129/files#diff-58ff3d740a93c8f80961d683b08033e09656826103140fdb3f96f9e7e8467cb1) diffs old and new namespace lists and calls the Kibana Spaces `POST /api/spaces/_update_objects_spaces` API to add/remove spaces in-place.
> - The `Update` method in [update.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2129/files#diff-c2dc1b42c203fb31ca2e97507fd739c6dc34523822a74de11fc0b66a01788b7d) now extracts prior and planned namespaces and calls the new function after a successful data view update.
> - An acceptance test in [acc_test.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2129/files#diff-f7a5f0faa6837399f33bebeafd65eb86f2e96e16240042dbb1642bbdc405a1fd) validates that the data view ID remains stable across add, remove, and add+remove namespace operations.
> - Behavioral Change: namespace changes no longer destroy and recreate the data view resource.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5f4ac8f.</sup>
> 
>

### DIFF
--- a/openspec/specs/kibana-data-view/spec.md
+++ b/openspec/specs/kibana-data-view/spec.md
@@ -65,7 +65,7 @@ resource "elasticstack_kibana_data_view" "example" {
     }))>
 
     allow_no_index = <optional, computed, bool> # default false; RequiresReplace
-    namespaces     = <optional, list(string)>   # RequiresReplace
+    namespaces     = <optional, list(string)>
   }
 }
 ```
@@ -77,15 +77,21 @@ Notes:
 
 ## Requirements
 
-### Requirement: Kibana Data Views APIs (REQ-001)
+### Requirement: Kibana Data Views and Spaces APIs (REQ-001)
 
-The resource SHALL manage data views through Kibana's Data Views HTTP APIs: create, get, update, and delete ([Kibana data views API docs](https://www.elastic.co/guide/en/kibana/current/data-views-api.html)).
+The resource SHALL manage data views through Kibana's Data Views HTTP APIs for create, get, update, and delete, and SHALL use Kibana's Spaces object-sharing API when reconciling `data_view.namespaces`.
 
 #### Scenario: CRUD uses Data Views APIs
 
 - GIVEN a managed Kibana data view
 - WHEN create, read, update, or delete runs
 - THEN the provider SHALL use the corresponding Kibana Data Views API operation
+
+#### Scenario: Namespace reconciliation uses Spaces API
+
+- GIVEN a managed Kibana data view whose `data_view.namespaces` membership changes
+- WHEN update runs
+- THEN the provider SHALL reconcile the namespace delta through Kibana's Spaces object-sharing API
 
 ### Requirement: API and client error surfacing (REQ-002)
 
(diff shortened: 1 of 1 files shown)


internal/clients/kibanaoapi/data_views_spaces.go
```diff
@@ -0,0 +1,111 @@
...
+		return diags
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := client.HTTP.Do(req)
+	if err != nil {
+		diags.AddError("Failed to update data view namespaces", err.Error())
+		return diags
+	}
+	defer resp.Body.Close()
```
tobio: This should all be converted to use `client.API.PostSpacesUpdateObjectsSpaces` with the generated request body.

internal/kibana/dataview/update.go
```diff
@@ -52,6 +56,31 @@ func (r *Resource) Update(ctx context.Context, req resource.UpdateRequest, resp
 		return
 	}
 
+	// Update namespaces via spaces API
```
tobio, line 59: Can you add an acceptance test validating this new behaviour. At a minimum it should:

* Create a dataview with an initial set of spaces
* Add a new space
* Remove a space
* Both add and remove spaces in a single change

tobio (commented):
Thanks for adding this, it's a much nicer UX. There's a couple of changes required here, are you able to add an entry to `CHANGELOG.md` as well.

internal/kibana/dataview/update.go
Onder7994: I add acc test.
Run with this command:

`make testacc-vs-docker TESTARGS='-run ^TestAccResourceDataViewNamespaces$'`

Output:
`PASS internal/kibana/dataview.TestAccResourceDataViewNamespaces (10.05s)
PASS internal/kibana/dataview`

internal/clients/kibanaoapi/data_views_spaces.go
Onder7994: Converted to `client.API.PostSpacesUpdateObjectsSpaces`. Check please

Issue #1501, opened 2025-12-02
[Feature] Add remote_elasticsearch output support for fleet


SimonKoetting:
would be nice to have a remote_elasticsearch output available to be able to terraform also setups where you have some agents/data sent to a different cluster (e.g. monitoring cluster)

PR #2131, opened 2026-04-02
[Feature] Add remote_elasticsearch output support for fleet

SimonKoetting:
relates to #1501 

> [!NOTE]
> ### Add `remote_elasticsearch` output type support to `elasticstack_fleet_output` resource
> - Extends the Fleet output resource schema to accept `remote_elasticsearch` as a valid type, adding new attributes: `service_token` (sensitive), `sync_integrations`, `sync_uninstalled_integrations`, and `write_to_logs_streams`.
> - Adds model conversion logic in [`models_remote_elasticsearch.go`](https://github.com/elastic/terraform-provider-elasticstack/pull/2131/files#diff-f6d423286ebca1753c34cb0e738d15b63e745245902b6233d39c917aa3eee7e3) for create, update, and read API operations, preserving `service_token` in state when the API redacts it.
> - Clears remote-only fields on non-remote output types (`elasticsearch`, `logstash`, `kafka`) to prevent spurious plan diffs.
> - Makes SSL `certificate` and `key` optional (previously required) in the `ssl` block to accommodate remote Elasticsearch outputs that only need CA authorities.
> - Extends `validators.Condition` with a `ValidateBool` method to support conditional validation on boolean attributes.
> - Adds acceptance tests, unit tests, and a Terraform example for the new output type.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6ef3630.</sup>
> 
>


tobio:
Sorry for polluting this PR a bit. It's a good test case for some improvements to the verify workflow to allow it to run on forks and PR which both add and implement a change proposal.

internal/fleet/output/acc_test.go
```diff
@@ -355,6 +356,52 @@ func TestAccResourceOutputKafkaComplex(t *testing.T) {
 	})
 }
 
+func TestAccResourceOutputRemoteElasticsearch(t *testing.T) {
+	serviceToken := os.Getenv("TF_ACC_FLEET_REMOTE_SERVICE_TOKEN")
+	if serviceToken == "" {
+		t.Skip("TF_ACC_FLEET_REMOTE_SERVICE_TOKEN must be set for remote_elasticsearch acceptance coverage")
+	}
```
tobio: At the moment, this will mean this test is essentially never executed. Any reason we can't just do something like:

```suggestion
	client, err := clients.NewAcceptanceTestingClient()
	require.NoError(t, err)
	kibanaOapiClient, err := client.GetKibanaOapiClient()
	require.NoError(t, err)
	resp, err := kibanaOapiClient.API.PostFleetServiceTokensWithResponse(t.Context(), kbapi.PostFleetServiceTokensJSONRequestBody{
		Remote: new(true),
	})
	require.NoError(t, err)
	serviceToken := resp.JSON200.Value
```

PR #2340, opened 2026-04-17
fix(kibana_security_detection_rule)!: use jsontypes.Normalized for ac…

### LINKED ISSUE #2339: kibana_security_detection_rule: action `params` schema should be JSON-normalized, not `map(string)`
## Problem

The `actions[*].params` attribute is currently typed as `map(string)` in the provider schema. The Kibana API defines `params` as a free-form object where values can be strings, objects, or arrays depending on the connector type (e.g. `.slack_api` uses a nested `subActionParams` object, `.email` uses arrays for `to`/`cc`/`bcc`, `.index` uses an array of documents).

This mismatch causes `subActionParams` (and any other non-string param value) to be silently wiped on every apply:

- **API → TF**: Non-string values are serialized with `fmt.Sprintf("%v", v)`, producing Go's `map[key:value]` notation — invalid JSON that corrupts state.
- **TF → API**: The corrupted string is sent back to the API, which discards it, so `subActionParams` is lost.

## Fix

Change `params` from `MapAttribute{ElementType: types.StringType}` to `StringAttribute{CustomType: jsontypes.NormalizedType{}}`, consistent with how `filters` is already modelled in the same resource.

Users set the full params object as a single `jsonencode()` call:
```hcl
params = jsonencode({
  subAction       = "postMessage"
  subActionParams = {
    channelIds = ["C123456"]
    text       = "Alert fired"
  }
})
```

This correctly handles all connector types without special-casing individual keys, and round-trips through the API losslessly.

## Breaking change

Existing configurations that use the map syntax must migrate to `jsonencode({...})`. A migration note will be added to the changelog.

mrohland:
…tion params to fix nested values being wiped on apply

Fixes elastic/terraform-provider-elasticstack#2339

> [!NOTE]
> ### Change `kibana_security_detection_rule` actions `params` from map to JSON-encoded string
> - Changes `actions[*].params` in the `kibana_security_detection_rule` resource from `map(string)` to a JSON-encoded string (`jsontypes.Normalized`), supporting nested objects and arrays that the map type could not represent.
> - Bumps the resource schema to version 1 and adds a state upgrader in [state_upgrade.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2340/files#diff-2956ef51b466b5c922fd48571f9cbed840365c94aa87615a4003a744166c99d7) that automatically migrates existing `params` map objects to JSON strings on provider upgrade.
> - Updates API conversion utilities to marshal/unmarshal the JSON string when reading from or writing to the Kibana API.
> - Adds an acceptance test `TestAccResourceSecurityDetectionRule_StateUpgradeParamsFromMap` that verifies migration from provider v0.14.4 state to the new representation.
> - Behavioral Change: existing configs must replace `params = { message = "..." }` with `params = jsonencode({ message = "..." })`; state is migrated automatically but plan files referencing the old syntax will break.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c83ede6.</sup>
> 
>

## Changelog

Customer impact: breaking
Summary: Change `elasticstack_kibana_security_detection_rule.actions[].params` to a JSON string rather than a map of string values. This allows setting arbitrary, nested param values 

#### `elasticstack_kibana_security_detection_rule` action `params` format change

Previously `elasticstack_kibana_security_detection_rule` used a map of strings for action parameters. This caused issues with actions requiring non-string based parameters (see https://github.com/elastic/terraform-provider-elasticstack/issues/2339 for an example). This has been changed to a single JSON string value which supports arbitrary param values. 

Previously

```hcl
resource "elasticstack_kibana_security_detection_rule" "test" {
...

  actions = [
    {
...
      params = {
        message = "Test state upgrade alert"
      }
...
  ]
}
```

becomes 

```hcl
resource "elasticstack_kibana_security_detection_rule" "test" {
...

  actions = [
    {
...
      params = jsonencode({
        message = "Test state upgrade alert"
      })
...
  ]
}
```


internal/kibana/security_detection_rule/models.go
```diff
@@ -175,13 +175,13 @@ type CardinalityModel struct {
...
-	ActionTypeID types.String `tfsdk:"action_type_id"`
-	ID           types.String `tfsdk:"id"`
-	Params       types.Map    `tfsdk:"params"`
-	Group        types.String `tfsdk:"group"`
-	UUID         types.String `tfsdk:"uuid"`
-	AlertsFilter types.Map    `tfsdk:"alerts_filter"`
-	Frequency    types.Object `tfsdk:"frequency"`
+	ActionTypeID types.String       `tfsdk:"action_type_id"`
+	ID           types.String       `tfsdk:"id"`
+	Params       jsontypes.Normalized `tfsdk:"params"`
```
tobio, line 180: @mrohland this change is failing gofmt. You should just be able to run `make lint` locally and commit the changes.

tobio (commented):
Looks good aside from the formatting, thanks for making this change. Are you able to fix the acceptance test config as well?

internal/kibana/security_detection_rule/models.go
mrohland: done

mrohland:
@tobio fixed acceptance tests, added state upgrade path for old states and updated the doc to include an example on the jsonencode for params.

docs/resources/elasticsearch_security_user.md
```diff
@@ -63,16 +63,14 @@ resource "elasticstack_elasticsearch_security_user" "dev" {
 
 ### Optional
 
-> **NOTE**: [Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments) are supported in Terraform 1.11 and later.
```
tobio, line 66: This change is caused by an older local Terraform CLI version. We need to control this better in the Makefile, but in the interim we need to revert this change.

internal/kibana/security_detection_rule/state_upgrade.go
tobio, line 1: Are you able to add an acceptance test verifying the state migration.

tobio (commented):
So I wanted to just fix up these last bits and get this merged, but I can't push to your `main` branch. Are you able to either cherry pick https://github.com/elastic/terraform-provider-elasticstack/commit/c909f03a68f4250c44e41e8f2973269174047b4c into this PR, or make similar changes yourself?

mrohland:
@tobio cherry picked your acceptance test into the branch. I don't fully understand your comment about the terraform cli though, so just let me know if there is more I can do to get this issue fixed as cleanly as possible.

tobio:
> I don't fully understand your comment about the terraform cli though

The commit you pulled in fixed this as well. I've also fixed this problem in main, so you don't really need to worry about it if you ever make other contributions. 

For completeness though :). 

Previously `make docs-generate` would use the `terraform` binary in your local PATH to build the provider schema that's used to generate the docs, older versions that didn't support write-only attribute wouldn't include [this](https://github.com/elastic/terraform-provider-elasticstack/pull/2340/changes/c83ede6a7484bc15266c0eae75bc364c0ac46725#diff-04b0b4ff5594e18c5ff49400ba094e8c40e2f915ee8c3caf29f1b6abb39fcaeeR66) note in the docs. CI would the fail because it's using the latest `terraform` CLI and the committed generated docs would look different. 

I've [updated](https://github.com/elastic/terraform-provider-elasticstack/pull/2392) the local workflow to use a specified CLI version so this Just Works™️.

tobio (approved):
LGTM, thanks for adding this in and seeing it through!

PR #2446, opened 2026-04-24
fix(security_role): write Read result to resp.State so drift is detected

jfrantz-cw:
# fix(security_role): write Read result to resp.State so drift is detected

Fixes #1693.

## The bug

The Plugin Framework `Read` implementation for `elasticstack_elasticsearch_security_role` writes the freshly-fetched role data to `req.State` (the request-side copy, discarded after the RPC) instead of `resp.State` (the response-side copy that Terraform persists).

As a result, the framework never sees the refreshed values — it silently keeps the prior state and hides all drift on `description`, `metadata`, and other role attributes. The user-visible symptom is `terraform plan` reporting "No changes" when the role has been modified out-of-band (e.g. via Kibana UI or a direct Elasticsearch API call).

`internal/elasticsearch/security/role/read.go:51`:

```go
// before
resp.Diagnostics.Append(req.State.Set(ctx, readData)...)
```

Every other `Read` in the provider — `systemuser`, `user`, `api_key`, `rolemapping`, `kibana/*`, etc. — correctly uses `resp.State.Set(ctx, …)`. `security/role` is the only resource with this mistake. A quick grep confirms:

```
$ grep -rn "resp.State.Set\|req.State.Set" internal/elasticsearch/security/
internal/elasticsearch/security/role/read.go:51:       resp.Diagnostics.Append(req.State.Set(ctx, readData)...)   # this file, before fix
internal/elasticsearch/security/rolemapping/read.go:133:  resp.Diagnostics.Append(resp.State.Set(ctx, readData)...)  # correct
internal/elasticsearch/security/api_key/read.go:47:      resp.Diagnostics.Append(resp.State.Set(ctx, *finalModel)...) # correct
internal/elasticsearch/security/systemuser/read.go:67:   resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)       # correct
internal/elasticsearch/security/user/read.go:92:         resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)       # correct
```

## Origin

`git blame` points at commit `8e9a2401c` (2025-12-05), which is PR #1331 "Migrate elasticstack_elasticsearch_security_role resource to Plugin Framework" — the SDKv2 → Plugin Framework migration for this resource. The migration was otherwise correct; this was a single-character oversight in the generated Read method. It has silenced all drift detection for `elasticstack_elasticsearch_security_role` in `v0.13.0` and later.

## The fix

One character — `req` → `resp`:

```go
// after
[...]

Issue #2436, opened 2026-04-23
[Bug] elasticstack_fleet_integration_policy: "inconsistent values for sensitive attribute" on .inputs after apply


cowboy-samurai:
## Summary

After `terraform apply`, the provider consistently errors with:

```
Error: Provider produced inconsistent result after apply

When applying changes to elasticstack_fleet_integration_policy.pubsub,
provider "provider[\"registry.terraform.io/elastic/elasticstack\"]" produced
an unexpected new value: .inputs: inconsistent values for sensitive attribute.
This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

## Provider and Terraform version

- Provider: `elastic/elasticstack` v0.14.3 (also reproduced on v0.14.5)
- Terraform: 1.x

## Reproduction config

```hcl
resource "elasticstack_fleet_integration_policy" "pubsub" {
  name                 = "instruqt-pubsub-logs"
  namespace            = ""
  integration_name     = "gcp_pubsub"
  integration_version  = "2.31.1"
  agent_policy_id      = elasticstack_fleet_agent_policy.example.policy_id

  inputs = {
    "gcp-gcp-pubsub" = {
      enabled = true
      streams = {
        "gcp_pubsub.gcp" = {
          enabled = true
          vars = jsonencode({
            project_id        = "my-project"
            topic             = "my-topic"
            subscription_name = "my-sub"
            tags              = ["forwarded"]
          })
        }
      }
    }
  }
}
```

Apply succeeds (resource is created/updated in Kibana), but then fails with the "inconsistent values" error before Terraform can write the new state.

cowboy-samurai:
## Dev validation — lifecycle workaround for provider bug

After the pipeline apply, `terraform apply` was erroring on `elasticstack_fleet_integration_policy.pubsub` with:

```
Error: Provider produced inconsistent result after apply
.inputs: inconsistent values for sensitive attribute
```

### Root cause

The Kibana API returns extra default vars per stream (e.g., `credential_json: null`, `subscription_type: "shared"`) that aren't in our config. After `HandleReqRespSecrets` unwraps them, `MapToNormalizedType` marshals the full map into a JSON string with more keys than the plan. Since `inputs[*].streams[*].vars` is `Sensitive: true`, Terraform does a raw string comparison — not semantic JSON equality — so the mismatch triggers the error. This is a provider bug.

[...]

Issue #2403, opened 2026-04-22
[Bug]Terraform Plan Does Not Validate Elasticsearch Role Privileges Before Apply


daltonkplu:
**Describe the bug**
Terraform `plan` succeeded but `apply` failed due to invalid Elasticsearch index privileges. Due to this bug https://github.com/elastic/terraform-provider-elasticstack/issues/1835, it also deleted the role and never recreated because of the `apply` error.

Log during `plan` with no errors:
```
        + privileges               = [
              + "auto_configure",
              + "auto_create",
              + "create_doc",
              + "create_index",
              + "manage",
              + "manage_ilm",
              + "monitor",
              + "view_index_metadata",
            ]
```

Error received during apply:
```
Failed with:
│ {"error":{"root_cause":[{"type":"action_request_validation_exception","reason":"Validation
│ Failed: 1: unknown index privilege [auto_create]. a privilege must be
│ either one of the predefined fixed indices privileges
│ [all,auto_configure,create,create_doc,create_index,cross_cluster_replication,cross_cluster_replication_internal,delete,delete_index,index,maintenance,manage,manage_data_stream_lifecycle,manage_failure_store,manage_follow_index,manage_ilm,manage_leader_index,monitor,none,read,read_cross_cluster,read_failure_store,view_index_metadata,write]
│ or a pattern over one of the available index
│ actions;"}],"type":"action_request_validation_exception","reason":"Validation
│ Failed: 1: unknown index privilege [auto_create]. a privilege must be
│ either one of the predefined fixed indices privileges
│ [all,auto_configure,create,create_doc,create_index,cross_cluster_replication,cross_cluster_replication_internal,delete,delete_index,index,maintenance,manage,manage_data_stream_lifecycle,manage_failure_store,manage_follow_index,manage_ilm,manage_leader_index,monitor,none,read,read_cross_cluster,read_failure_store,view_index_metadata,write]
│ or a pattern over one of the available index actions;"},"status":400}`.
```

The role HCL contained:
```
...
 privileges = [
      "create_doc",
      "auto_create",
      "auto_configure",
      "create_index",
      "view_index_metadata",
      "monitor",
      "manage_ilm",
      "manage"
    ]
...
```

**To Reproduce**
1. TF `elasticstack_elasticsearch_security_role` configuration contained:
```
...
 privileges = [
      "create_doc",
      "auto_create",
      "auto_configure",
      "create_index",
      "view_index_metadata",
      "monitor",
[...]

Issue #2006, opened 2026-03-24
[Bug] elasticstack_kibana_security_detection_rule returns `null` when provided with `[]`


m-a-leclercq:
**Describe the bug**

The `elasticstack_kibana_security_detection_rule` resource returns `"Provider produced inconsistent result after apply"` errors when any of the following nested list attributes are set to an empty list `[]`:

- `actions`
- `exceptions_list`
- `severity_mapping`
- `risk_score_mapping`
- `related_integrations`
- `threat`
- `threat_mapping`

Additionally, nested sub-lists within `threat` (e.g. `threat[0].technique[0].subtechnique = []`) are also affected.

The provider plans the value as an empty list (`cty.ListValEmpty(...)`) but the read-back after apply returns `null`, causing Terraform to report an inconsistency. Per the Terraform plugin framework contract, the provider's `Create`/`Read` functions must return `[]` (empty list) — not `null` — when the planned value was `[]`.

**To Reproduce**

1. Use the following Terraform configuration:

```hcl
terraform {
  required_version = ">= 1.12.0"
  required_providers {
    elasticstack = {
      source  = "elastic/elasticstack"
      version = "~>0.14"
    }
  }
}

provider "elasticstack" {
  elasticsearch {
    api_key   = "REDACTED"
    endpoints = ["https://REDACTED.es.cloud.es.io:443/"]
  }
  kibana {
    api_key   = "REDACTED"
    endpoints = ["https://REDACTED.kb.cloud.es.io:443/"]
  }
}

resource "elasticstack_kibana_security_detection_rule" "repro" {
  name        = "Repro: empty nested lists return null"
  description = "Minimal reproduction of provider bug"
  type        = "query"
  rule_id     = "00000000-0000-0000-0000-000000000001"
  risk_score  = 50
  severity    = "medium"
  query       = "host.name : *"
  language    = "kuery"
  index       = ["logs-*"]
  enabled     = false

  # All of these are set to [] (empty list) in the plan,
  # but the provider returns null after apply
  actions              = []
  exceptions_list      = []
  threat               = []
  threat_mapping       = []
  severity_mapping     = []
  risk_score_mapping   = []
  related_integrations = []
}
```

2. Run `terraform apply --auto-approve`
3. Observe 7 errors of the form:

```
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to elasticstack_kibana_security_detection_rule.repro,
[...]

PR #2447, opened 2026-04-24
fix(fleet_integration): detect out-of-band version changes via InstallationInfo

### LINKED ISSUE #1585: [Bug] elasticstack_fleet_integration does not ensure the correct version is installed
To repro:

1. Install a package via the TF resource, for example `tcp` at version `1.16.0`
2. Upgrade the package via Kibana UI or devtools, for example `tcp` to `1.17.0` 
3. `terraform plan`

Expected:
Terraform should downgrade the package back to 1.16.0

Actual:
No changes required

This is partly because the Fleet API is a bit confusing, getting a package with a specific version lists the package as `installed` even if the installed version isn't the version requested. We can work around this in the provider however.

jfrantz-cw:
# fix(fleet_integration): detect out-of-band version changes via InstallationInfo

Fixes #1585.

## The bug

Per @tobio's writeup in #1585:

> 1. Install a package via the TF resource, for example `tcp` at version `1.16.0`
> 2. Upgrade the package via Kibana UI or devtools, for example `tcp` to `1.17.0`
> 3. `terraform plan`
>
> Expected: Terraform should downgrade the package back to 1.16.0
> Actual: No changes required

Fleet's `GET /epm/packages/{name}/{version}` reports `status: installed` whenever the package is installed at *any* version — the version in the path parameter is not used to decide the installed-status flag. The Read implementation in `internal/fleet/integration/read.go` trusted that flag and never noticed when the actually-installed version differed from the version recorded in state:

```go
// before
if pkg == nil || !fleetPackageInstalled(pkg) {
    resp.State.RemoveResource(ctx)
    return
}
stateModel.ID = types.StringValue(getPackageID(name, version))
```

`version` here is the stale value from state, not what Fleet says is actually installed. State is never updated, so `terraform plan` sees `state.version == config.version` and reports no diff.

## The fix

The response body does carry the actually-installed version in `pkg.InstallationInfo.Version` (see `generated/kbapi/kibana.gen.go:63053`). Consult that when populating state:

```go
// after
installedVersion := version
if pkg.InstallationInfo != nil && pkg.InstallationInfo.Version != "" {
    installedVersion = pkg.InstallationInfo.Version
}
stateModel.Version = types.StringValue(installedVersion)
stateModel.ID = types.StringValue(getPackageID(name, installedVersion))
```

With this in place:
- Refresh writes `installedVersion` into state.
- Plan diffs `state.version` (actual) against `config.version` (desired) and surfaces the drift.
- Update (which delegates to Create → `InstallPackage`) runs to reconcile on the next apply.

`InstallationInfo` is the authoritative source on recent Kibana versions. If it is absent (older servers), the fallback path preserves the previous behavior by keeping `version` from state — `fleetPackageInstalled`'s existing `Status` fallback continues to gate `RemoveResource` for the "package gone" case.

## Test updates

The existing `TestAccResourceIntegration_ExternalChange` codified the buggy behavior: step 2 installed `tcp@1.17.0` out-of-band and then asserted that `state.version` was still `"1.16.0"`. That assertion is exactly what #1585 is complaining about.

Replaced with a proper drift-detection test using `PlanOnly: true, ExpectNonEmptyPlan: true` (same pattern as `TestAccResourceIntegrationDeleted`), asserting:
- After the out-of-band upgrade, `state.version == "1.17.0"` (actual installed).
- Plan is non-empty (terraform wants to downgrade back to `1.16.0`).

The redundant step 3 (out-of-band uninstall) was dropped — `TestAccResourceIntegrationDeleted` already covers that case.

## Verification

```
$ go vet ./internal/fleet/integration/
$ go build ./internal/fleet/integration/
$ go test -count=1 -run 'TestAccResourceIntegration_ExternalChange' ./internal/fleet/integration/
```

Acceptance test requires a live Elasticsearch + Kibana (per the repo testing docs); the above ran to completion locally for compile/vet, and I relied on CI for the full acceptance run. Happy to rerun locally with a containerised stack if you want me to confirm the test result before merge.

## Scope

- Read-only behavioral change in a single resource (`elasticstack_fleet_integration`).
- No schema, no API, no model changes — the fix uses a field already generated from the Kibana OpenAPI spec.
- Existing acceptance test rewritten to assert correct behavior; new test function count unchanged.
- No CHANGELOG edit (the repo auto-generates the `[Unreleased]` section from merged PR history per CONTRIBUTING.md).

## Context

This is the second in a short series of targeted CoreWeave-authored fixes against this provider. The first is #2446 (`security_role` silent drift). The list lives behind internal planning — happy to continue if these are welcome.

## Changelog
Customer impact: fix
Summary: elasticstack_fleet_integration now detects out-of-band package upgrades and downgrades during refresh by consulting InstallationInfo.Version; terraform plan surfaces the drift instead of reporting "No changes".

> [!NOTE]
> ### Detect out-of-band integration version changes via `InstallationInfo` in Fleet resource reads
> - [read.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2447/files#diff-5319ee2508fe0fe646cb174c486e85c24dbd8e6ba98363e68deaa1c84e847ed8): `Read` now derives the installed package version from `pkg.InstallationInfo.Version` when available, updating both `stateModel.Version` and the resource ID to reflect the actually installed version rather than the requested version.
> - This causes Terraform to detect plan drift when a package is upgraded or downgraded outside of Terraform, surfacing out-of-band changes correctly.
> - [acc_test.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2447/files#diff-aaef4b65af6a17f4aea136d36cdc736d23ed916a5c5d331172eac30b25e776bd): The acceptance test `TestAccResourceIntegration_ExternalChange` is updated to upgrade `tcp` from `1.16.0` to `1.17.0` via the Fleet API and assert a non-empty plan, and is now gated on Fleet >= `8.9.0` (the first version that populates `InstallationInfo`).
> - Behavioral Change: `Read` now records the installed version instead of the requested version in state when `InstallationInfo` is present, which will trigger plan drift for any existing resources whose installed version differs from configuration.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bae5902.</sup>
> 
>

### DIFF
--- a/internal/fleet/integration/acc_test.go
+++ b/internal/fleet/integration/acc_test.go
@@ -185,10 +185,21 @@ func TestAccResourceIntegrationDeleted(t *testing.T) {
 	})
 }
 
+// TestAccResourceIntegration_ExternalChange asserts that out-of-band version
+// changes to an installed Fleet integration package are detected on the next
+// refresh, and that terraform plan surfaces the drift.
+//
+// Regression test for https://github.com/elastic/terraform-provider-elasticstack/issues/1585:
+// Fleet's GET /epm/packages/{name}/{version} returns status "installed"
+// whenever the package is installed at *any* version, so the provider
+// previously did not notice out-of-band upgrades. Read now consults
+// InstallationInfo.Version and records the actually-installed version in
+// state, so plan sees a diff between state and config.
 func TestAccResourceIntegration_ExternalChange(t *testing.T) {
 	resource.Test(t, resource.TestCase{
 		PreCheck: func() { acctest.PreCheck(t) },
 		Steps: []resource.TestStep{
+			// Step 1: apply tcp@1.16.0 via terraform.
 			{
 				ProtoV6ProviderFactories: acctest.Providers,
 				SkipFunc:                 versionutils.CheckIfVersionIsUnsupported(minVersionIntegration),
@@ -198,6 +209,11 @@ func TestAccResourceIntegration_ExternalChange(t *testing.T) {
 					resource.TestCheckResourceAttr("elasticstack_fleet_integration.test_integration", "version", "1.16.0"),
 				),
 			},
+			// Step 2: upgrade tcp to 1.17.0 via the Fleet API (out-of-band).
+			// The next refresh must record the *actually installed* version
+			// (1.17.0) in state, and the resulting plan must be non-empty
+			// because the configured version (1.16.0) no longer matches
+			// state.
 			{
 				ProtoV6ProviderFactories: acctest.Providers,
 				SkipFunc:                 versionutils.CheckIfVersionIsUnsupported(minVersionIntegration),
@@ -205,8 +221,6 @@ func TestAccResourceIntegration_ExternalChange(t *testing.T) {
 				PreConfig: func() {
 					notSupported, err := versionutils.CheckIfVersionIsUnsupported(minVersionIntegration)()
 					require.NoError(t, err)
-
-					// Skip the pre-config if the version is not supported
 					if notSupported {
 						return
 					}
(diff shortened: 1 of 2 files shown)


tobio:
> The list lives behind internal planning

@jfrantz-cw it would be great if you were able to assign yourself to any other issues your planning to look at if possible.

tobio (approved):

jfrantz-cw:
> > The list lives behind internal planning
> 
> @jfrantz-cw it would be great if you were able to assign yourself to any other issues your planning to look at if possible.

consider it done my friend

jfrantz-cw:
Tried self-assigning, hit a permissions error (`ReplaceActorsForAssignable`). Looks like external contributors can't do it directly on this repo.

Queue I'm planning to work through:

1. #2436, highest priority
2. #2403 and #1835 (plan doesn't validate ES role privileges, plus the import-flow companion)
3. #1469 (sensitive-attribute on `kibana_action_connector`, probably falls out of #2436's fix)
4. #691 and #864 (`fleet_server_host` `default=true` undeletable, 404 on modify)
5. #2006 (`kibana_security_detection_rule` returns null on `[]`)

Want me to drop a comment on each, or can you assign / grant perms?

PR #2451, opened 2026-04-24
fix(kibana_data_view): ignore Kibana-auto-populated field_attrs.count

### LINKED ISSUE #1287: [Bug] `elasticstack_kibana_data_view` forces replacement due to auto-calculated `field_attrs`
The `elasticstack_kibana_data_view` resource is unexpectedly forced into a replacement state by Terraform. This occurs after the data view has been used in Kibana, which generates field popularity statistics. The provider detects these server-side generated statistics (specifically the `count` attribute within `field_attrs`) as a change from the state file, triggering a plan to replace the entire resource.

**To Reproduce**

1.  TF configuration used:

    ```hcl
    terraform {
      required_providers {
        elasticstack = {
          source  = "elastic/elasticstack"
          version = "~> 0.11.0"
        }
      }
    }

    provider "elasticstack" {
      // Elasticsearch & Kibana connection details
    }

    resource "elasticstack_kibana_data_view" "test_view" {
      data_view = {
        title = "my-test-logs-*"
        name  = "My Test Data View"
        id    = "my-test-data-view-id"
      }
    }
    ```

2.  TF operations to execute to get the error:

      - Run `terraform apply` to create the resource.
      - In the Kibana UI, navigate to Discover and interact with the "My Test Data View" to generate field usage statistics.
      - Run `terraform plan`.

3.  See the error in the output: The plan will show that the resource needs to be replaced.

    ```diff
      # elasticstack_kibana_data_view.test_view must be replaced
    -/+ resource "elasticstack_kibana_data_view" "test_view" {
          ~ data_view = {
              - field_attrs     = {
                  - "host.hostname" = { # forces replacement
                      - count = 5 -> null
                    },
                  - "event.action" = { # forces replacement
                      - count = 12 -> null
                    },
                } -> null
                id              = "my-test-data-view-id"
                name            = "My Test Data View"
                # (3 unchanged attributes hidden)
            }
          ~ id        = "default/my-test-data-view-id" -> (known after apply)
            # (2 unchanged attributes hidden)
        }
    ```

**Expected behavior**
After the initial creation, subsequent `terraform plan` operations should show no changes to the infrastructure. The provider should ignore server-side, auto-calculated attributes that are not managed within the Terraform configuration.

**Debug output**
(To be provided after running the reproduction steps with `TF_LOG=trace` enabled)

 - OS: Linux
 - Terraform Version: v1.x
 - Provider version: v0.11.0
 - Elasticsearch Version: 8.x

**Additional context**
The root cause appears to be that the provider does not differentiate between user-managed attributes and dynamic, server-side attributes like field popularity counters. This results in a perpetual diff, making idempotent management of the resource impossible once it is actively used.

jfrantz-cw:
# fix(kibana_data_view): ignore Kibana-auto-populated `field_attrs.count`

Fixes #1287.

## The bug

Reported scenario (👍×8 on the issue, with a follow-up from LukoJy3D: *"random ID change resulted in a bunch of broken dashboards that depend on those data views"*):

1. `terraform apply` creates an `elasticstack_kibana_data_view` with no `field_attrs` configured.
2. Someone opens the data view in Kibana Discover. Kibana records field popularity statistics directly back onto the saved-object: `field_attrs.<field>.count = N`.
3. `terraform plan` now reports the whole resource must be **replaced** — not updated:

   ```text
     # elasticstack_kibana_data_view.test_view must be replaced
   -/+ resource "elasticstack_kibana_data_view" "test_view" {
         ~ data_view = {
             - field_attrs = {
                 - "host.hostname" = { # forces replacement
                     - count = 5 -> null
                   },
                 - "event.action" = { # forces replacement
                     - count = 12 -> null
                   },
               } -> null
             ...
           }
         ~ id        = "default/my-test-data-view-id" -> (known after apply)
       }
   ```

Because the `id` changes in the process, every dashboard, alerting rule, and downstream resource that referenced the prior id breaks.

## Root cause

In `internal/kibana/dataview/schema.go`, `field_attrs` carries `RequiresReplace`:

```go
"field_attrs": schema.MapNestedAttribute{
    ...
    PlanModifiers: []planmodifier.Map{
        mapplanmodifier.RequiresReplace(),
    },
},
```

That modifier is actually correct for *user-initiated* changes: the Kibana data-view PUT endpoint's `DataViewsUpdateDataViewRequestObjectInner` struct has no `fieldAttrs` field, so a true change to a custom label can only be applied by destroy+create.

The problem is that `populateFromAPI` mirrored **every** `field_attrs` entry Kibana returned — including entries Kibana wrote itself for popularity tracking — into Terraform state. Those server-only entries then looked identical to a user's config change, and the RequiresReplace path kicked in.

## The fix

Two small changes, tests cover both:

1. **Filter server-only entries on Read** (`models.go`).

   A new `filterFieldAttrs` helper drops entries where `custom_label` is nil *and* the entry isn't already present in prior state. The user-facing effect: count-only popularity entries that Kibana adds on its own are never put into state, so they can't produce a diff.

2. **`count` preserves state when config omits it** (`schema.go`).

   The nested `count` attribute gains `Computed: true` and a `int64planmodifier.UseStateForUnknown()` modifier. For fields the user *did* configure via `custom_label`, refresh-time count updates from Kibana now land in state cleanly, and subsequent plans don't diff `null → <server value>`.

`field_attrs`'s map-level `RequiresReplace` is intentionally left in place — it's the right behavior for user-initiated additions because the Update API still cannot change `field_attrs`.

## Test coverage

New table-driven unit test `TestFilterFieldAttrs` in `filter_field_attrs_test.go`, 6 subtests covering:

| Scenario | API | Prior state | Kept entries |
|---|---|---|---|
| empty API | `{}` | null | none |
| nil API | `nil` | null | none |
| server-only count-only, no prior | 2 entries, count only | null | **none (dropped)** |
| mixed custom_label + count-only, no prior | 1 w/ label, 1 count-only | null | only the one w/ label |
| server entries kept when already in prior | 1 w/ label, 1 count-only | has the count-only key | both |
| all-server entries, prior is empty set | 3 entries, count only | empty map | none |

```
$ go test -count=1 -run 'TestFilterFieldAttrs' -v ./internal/kibana/dataview/
--- PASS: TestFilterFieldAttrs (0.00s)
    --- PASS: TestFilterFieldAttrs/empty_API_returns_empty
    --- PASS: TestFilterFieldAttrs/nil_API_returns_nil
    --- PASS: TestFilterFieldAttrs/server-only_count-only_entries_dropped_when_not_in_prior_state
    --- PASS: TestFilterFieldAttrs/entries_with_custom_label_kept_even_when_also_count-populated
    --- PASS: TestFilterFieldAttrs/server_entries_kept_when_their_key_is_already_in_prior_state
    --- PASS: TestFilterFieldAttrs/all-server_entries_with_empty_prior_state_→_empty_output

$ go test -count=1 ./internal/kibana/dataview/
ok  	github.com/elastic/terraform-provider-elasticstack/internal/kibana/dataview	0.717s
```

## Why not something stronger?

Considered and rejected:

- **Drop `RequiresReplace` entirely.** The Kibana Update API genuinely can't change `field_attrs`, so user-initiated changes would silently no-op. Worse, not better.
- **Make `count` `Computed` only (not `Optional`).** Existing configs that happen to specify `count` would fail schema validation. More user pain than the asymmetric `Optional + Computed` path.
- **Strip `count` from state entirely.** Loses user-visible signal for no real benefit once the filter is in place.

## Scope

- `schema.go`: add `Computed: true` + `int64planmodifier.UseStateForUnknown()` on `field_attrs.*.count`. Add one import.
- `models.go`: add `filterFieldAttrs` helper, wire it into the existing `populateFromAPI` chain.
- `filter_field_attrs_test.go`: new file, 6-case table-driven test.
- No schema version bump (additive attribute changes only).
- No CHANGELOG edit (auto-generated).

## Changelog
Customer impact: fix
Summary: `elasticstack_kibana_data_view` no longer forces resource replacement when Kibana records field popularity statistics on a data view that was interacted with in the UI. Server-populated `field_attrs.<field>.count` entries are now filtered out of state, and the `count` attribute preserves state across refreshes when omitted from configuration.

> [!NOTE]
> ### Fix Kibana data view `field_attrs` to ignore auto-populated `count`-only entries
> - Adds `filterFieldAttrs` in [models.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2451/files#diff-04eb57a6856887354e38532ac9eafc020b82aa2e06ba94d68da7b0dbce836b8c) to strip entries where `custom_label` is nil before populating Terraform state, preventing server-managed popularity counts from leaking into state.
> - Marks `field_attrs.count` as `Computed` with `UseStateForUnknown` in [schema.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2451/files#diff-5ee27c86734848be031c141b124701b5ec7a03a803745b94f806d98281fc2ae0) so server-maintained counts preserve prior state and don't cause spurious plan diffs.
> - Behavioral Change: `field_attrs` entries that previously appeared in state with only a `count` value (no `custom_label`) will be dropped on the next refresh.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized db78fd1.</sup>
> 
>

### DIFF
--- a/internal/kibana/dataview/filter_field_attrs_test.go
+++ b/internal/kibana/dataview/filter_field_attrs_test.go
@@ -18,16 +18,10 @@
 package dataview
 
 import (
-	"context"
 	"testing"
 
 	"github.com/elastic/terraform-provider-elasticstack/generated/kbapi"
-	"github.com/elastic/terraform-provider-elasticstack/internal/utils/typeutils"
-	"github.com/hashicorp/terraform-plugin-framework/diag"
-	"github.com/hashicorp/terraform-plugin-framework/path"
-	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
 // TestFilterFieldAttrs pins the behaviour that resolves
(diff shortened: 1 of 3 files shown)


jfrantz-cw:
bumping this because it would essentially close a support ticket

tobio (commented):
The changes in this PR make an assumption that users *never* want to manage `count` but then leaves it configurable but ultimately ignored. This is kind of touched on in the PR body, but that was prior to the change to ignore the current state within the filter. 

I think we need to refine this change along one of these lines:
1. Make `count` computed only. Be clear in the schema that this isn't user-definable. I'm hesitant to do this since, it is user definable. 
2. Remove the `RequiresReplace` and wire updates to field attributes into [update field metadata](https://www.elastic.co/docs/api/doc/kibana/v8/operation/operation-updatefieldsmetadatadefault). 
3. Go back to the prior filter logic (ignore fields not explicitly managed by the user), but use _Config_ rather than State. This would clean up existing resources impacted by this bug, but still allow users to manage `count`. 

I think we actually should do 2 & 3, RequiresReplace is always a pain, and there's a viable alternative but we still want to ignore changes to non-user-managed field attributes.

tobio:
Fixed in https://github.com/elastic/terraform-provider-elasticstack/pull/2964

PR #2452, opened 2026-04-24
fix(kibana_space): preserve user-configured empty strings on update

### LINKED ISSUE #1881: Terraform provider cannot reset elasticstack_kibana_space ressource description and imageUrl field
## Readiness Checklist

- [ x] I am running the latest version
- [ x] I checked the documentation and found no answer
- [ x] I checked to make sure that this issue has not already been filed
- [ x] I am reporting the issue to the correct repository (for multi-repository projects)

## Terraform definition

When suppressing a field image or description, the Terraform provider should set it to "" for deletion
## Steps to Reproduce

```
provider "elasticstack" {
  alias = "v9"
  
  elasticsearch {
    endpoints = [ec_deployment.deployment_v9.elasticsearch.https_endpoint]
    username  = ec_deployment.deployment_v9.elasticsearch_username
    password  = ec_deployment.deployment_v9.elasticsearch_password
  }
  
  kibana {
    endpoints = [ec_deployment.deployment_v9.kibana.https_endpoint]
  }
}

resource "elasticstack_kibana_space" "v9_test_space" {
  provider = elasticstack.v9
  
  space_id          = "test_space"
  name              = "Test Space"
  description       = "this is the description"
  color             = "#a00000"
  disabled_features = ["ingestManager", "enterpriseSearch"]
  initials          = "ts"
}
```

Change the description to description  = "" and apply.

Terraform send the following API call:

```json
{
  "id": "test_space",
  "name": "Test Space",
  "disabledFeatures": ["ingestManager", "enterpriseSearch"],
  "initials": "ts",
  "color": "#a00000"
}
```
omitting the `description  = ""` this result in the field keeping the old value.

If you send the correct request with the Kibana API the description is properly removed:
```bash
curl -X PUT "https://kibana_url:443/api/spaces/space/test_space" \
  -H "Content-Type: application/json" \
  -H "kbn-xsrf: true" \
  -u "user:passwrod" \
  -d '{
    "id": "test_space",
    "name": "Test Space",
    "description": "", 
    "color": "#a00000",
    "initials": "ts",
    "disabledFeatures": ["ingestManager", "enterpriseSearch"]
  }'
```

The same applies for the field `imageUrl`

jfrantz-cw:
# fix(kibana_space): preserve user-configured empty strings on update

Fixes #1881.

## The bug (verbatim from the issue)

```hcl
resource "elasticstack_kibana_space" "v9_test_space" {
  space_id          = "test_space"
  name              = "Test Space"
  description       = "this is the description"
  color             = "#a00000"
  disabled_features = ["ingestManager", "enterpriseSearch"]
  initials          = "ts"
}
```

> Change the description to `description = ""` and apply. Terraform sends the following API call:
>
> ```json
> {
>   "id": "test_space",
>   "name": "Test Space",
>   "disabledFeatures": ["ingestManager", "enterpriseSearch"],
>   "initials": "ts",
>   "color": "#a00000"
> }
> ```
>
> omitting the `description = ""` this result in the field keeping the old value.

Also reported on `image_url`.

## Root cause

`internal/kibana/space.go` used SDKv2 `d.GetOk` to decide whether to include each optional string:

```go
var description *string
if v, ok := d.GetOk("description"); ok {
    s := v.(string)
    description = &s
}
```

[SDKv2's `GetOk`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#ResourceData.GetOk) returns `(value, ok=false)` whenever the value equals the type's zero value — for `string`, that's `""`. So an explicit `description = ""` looks identical to "attribute omitted", and the provider sends `Description: nil` in the PUT body. The `json:"description,omitempty"` tag on the generated `*string` field then strips it from the JSON altogether, so Kibana never receives a request to clear the field.

Same happened for `initials`, `color`, and `image_url`.

## The fix

Introduce a small helper that reads `d.GetRawConfig()` and distinguishes "attribute absent" from "attribute set to empty":

```go
func configuredString(rawConfig cty.Value, attr string) *string {
    if !rawConfig.IsKnown() || rawConfig.IsNull() {
        return nil
    }
    if !rawConfig.Type().IsObjectType() || !rawConfig.Type().HasAttribute(attr) {
        return nil
    }
    v := rawConfig.GetAttr(attr)
    if v.IsNull() || !v.IsKnown() {
        return nil
    }
    s := v.AsString()
    return &s
}
```

And wire it in:

```go
description := configuredString(rawConfig, "description")
initials    := configuredString(rawConfig, "initials")
color       := configuredString(rawConfig, "color")
imageURL    := configuredString(rawConfig, "image_url")
```

With this, a user's explicit `description = ""` flows through as `&""`, the `*string` field in `PutSpacesSpaceIdJSONRequestBody` is non-nil, Go's `encoding/json` includes `"description": ""` in the marshaled body despite the `omitempty` tag (pointers to zero-value strings are still non-nil, so `omitempty` keeps them), and Kibana clears the field as requested.

The `solution` attribute keeps its `GetOk` call — solution only accepts a fixed set of non-empty string values (`security`, `oblt`, `es`, `classic`), so the ambiguity doesn't apply.

## Test coverage

New unit test `TestConfiguredString` with 6 subtests:

| Scenario | Input | Expected |
|---|---|---|
| Null root object | `cty.NullVal(objType)` | `nil` |
| Attribute not in type | Object with `description`, asked for `nonexistent` | `nil` |
| Null attribute | `description = null` | `nil` |
| **Explicit empty string (the fix)** | `description = ""` | `&""` |
| Explicit value | `description = "hello world"` | `&"hello world"` |
| Cross-attribute independence | `description = "kept", image_url = ""` → ask for `image_url` | `&""` |

```
$ go test -count=1 -run 'TestConfiguredString' -v ./internal/kibana/
--- PASS: TestConfiguredString (0.00s)
    --- PASS: null_root_returns_nil
    --- PASS: attribute_missing_from_type_returns_nil
    --- PASS: null_attribute_returns_nil
    --- PASS: explicit_empty_string_returns_pointer_to_empty_string_(the_bug_fix)
    --- PASS: explicit_value_returns_pointer_to_value
    --- PASS: distinct_attributes_don't_bleed_—_image_url_reset_while_description_kept

$ go test -count=1 ./internal/kibana/
ok  	github.com/elastic/terraform-provider-elasticstack/internal/kibana	0.805s
```

## Scope

- 1 helper added (20 lines) with a clear doc-comment.
- 4 `GetOk` call-sites in `resourceSpaceUpsert` replaced with `configuredString`.
- 1 unit-test file added.
- No schema change, no CHANGELOG edit (auto-generated), no behavior change for users who set or omit these attributes normally.

## Follow-up considerations (not in this PR)

SDKv2's `GetOk`-as-presence-check is [a well-known footgun](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#getok) and almost certainly affects other SDKv2-based resources in this repo. A broader audit could be a follow-up PR if maintainers would welcome it — I intentionally kept this change scoped to the resource that has a bug report filed against it.

## Changelog
Customer impact: fix
Summary: `elasticstack_kibana_space` now correctly clears `description`, `initials`, `color`, and `image_url` when the configuration sets them to an empty string. Previously those explicit empty-string assignments were silently dropped from the outbound API request and Kibana retained the prior value.

> [!NOTE]
> ### Fix `elasticstack_kibana_space` to preserve empty strings on update
> - `d.GetOk` treats empty strings as unset, so fields like `description`, `initials`, `color`, and `image_url` were silently dropped from Kibana API requests when set to `""`.
> - Adds a [`configuredString`](https://github.com/elastic/terraform-provider-elasticstack/pull/2452/files#diff-b6fcf5d78b5b03ef1f49a05a67ab6fd3065276d3a8596be7c80fd72a3d74d59d) helper that reads attributes directly from the raw Terraform config via `d.GetRawConfig()`, returning a non-nil `*string` for explicitly configured empty strings.
> - `resourceSpaceUpsert` now uses this helper for the four affected fields, so setting them to `""` clears the corresponding Kibana Space fields instead of retaining previous values.
> - Behavioral Change: Create/Update requests will now send empty-string values to the Kibana API for these fields when explicitly configured, where previously they were omitted.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7e960d2.</sup>
> 
>


tobio (commented):
There's a _strong_ preference to migrating to the plugin framework to avoid exactly this issue. 

This PR needs an acceptance test verifying this actually fixes the underlying issue.

jfrantz-cw:
plugin framework migration as a separate pr makes sense, that's a bigger refactor. adding an acceptance test that proves an empty string update clears the field in the api, will push soon.

PR #2466, opened 2026-04-24
Auto generate skill for agent-skills-sandbox repo

tehbooom:
Auto generates a skill to put in elastic/agent-skills-sandbox repo

Will need to discuss if this fits here or if we would like to have another repo pull in the providers to generate and then push to that repo.

> [!NOTE]
> ### Add skill generator CLI for the elasticstack-terraform agent-skills-sandbox repo
> - Adds a Go CLI at [`scripts/generate-skill`](https://github.com/elastic/terraform-provider-elasticstack/pull/2466/files#diff-29e9189664c8ba8eac41ff0ad202c4811890037b51374c105c1acbecc03ea0e8) that produces a deterministic skill directory tree under `dist/skill/` by combining hand-authored assets and provider docs.
> - Scans `docs/resources` and `docs/data-sources`, parses frontmatter, and merges entities into a generated `references/index.md` with descriptions and file links.
> - Copies static assets (including [`SKILL.md`](https://github.com/elastic/terraform-provider-elasticstack/pull/2466/files#diff-b0547e30c2cea23004fe51fcb63d84215fd0a751f9b21476d75682549621e694) and reference docs) into the output tree, substituting `{{VERSION}}` with the value passed via `--provider-version`.
> - Writes a `GENERATED.md` provenance file summarising entity counts and provider version.
> - Adds `make skill-generate` and `make skill-test` targets to invoke the generator and its unit tests respectively.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6f18ed1.</sup>
> 
>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,6 +150,14 @@ workflow-test: ## Run unit tests for workflow source generation
 check-workflows: ## Check generated workflow markdown sources
 	@ go run ./scripts/compile-workflow-sources --manifest .github/workflows-src/manifest.json --check --verbose
 
+.PHONY: skill-generate
+skill-generate: ## Generate the elasticstack-terraform agent skill into dist/skill/
+	@ go run ./scripts/generate-skill -provider-version $(VERSION)
+
+.PHONY: skill-test
+skill-test: ## Run unit tests for the skill generator
+	@ go test ./scripts/generate-skill/... -count=1
+
 .PHONY: gen
 gen: docs-generate ## Generate the code and documentation
 	@ go generate ./...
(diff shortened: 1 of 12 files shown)


tobio (commented):
I'm curious what this solves over one of:
- Referencing https://registry.terraform.io/providers/elastic/elasticstack/latest/docs
- Embedding the `docs` directory (well formed, generated from the implementation so no chance of drift)

I suppose referencing HTTP docs in a skill _might_ be less effective than sourcing local content, but basing this off the generated docs feels like a better path (they're for the right audience, where the specs are more developer focused).

tehbooom:
> I'm curious what this solves over one of:
> 
> * Referencing https://registry.terraform.io/providers/elastic/elasticstack/latest/docs
> * Embedding the `docs` directory (well formed, generated from the implementation so no chance of drift)
> 
> I suppose referencing HTTP docs in a skill _might_ be less effective than sourcing local content, but basing this off the generated docs feels like a better path (they're for the right audience, where the specs are more developer focused).

You're right using the docs directory is better than the specs

PR #2582, opened 2026-04-30
fix(fleet): sync space_id from Fleet on create and read

vatsaldesai93:
## Problem

`elasticstack_fleet_integration` marks `space_id` as `Computed + RequiresReplace` but neither `Create` nor `Read` populate it from Fleet. After creating without an explicit `space_id`, state holds `null` permanently. On the next plan, any non-null config value for `space_id` triggers `RequiresReplace`, which Fleet blocks when active policies exist.

## Fix

- `Read`: populate space_id from InstallationInfo.InstalledKibanaSpaceId only when null in state, this prevents overwriting an explicitly configured value
- `Create`: same lookup immediately after install, before writing state
- Test: assert `space_id = "default"` when created without explicit config

## Changelog
Customer impact: fix
Summary: `elasticstack_fleet_integration` now syncs `space_id` from Fleet on both create and read, preventing state drift that caused unexpected forced replacements.

> [!NOTE]
> ### Sync `space_id` from Fleet API on create and read for fleet integration resource
> - On create, populates `space_id` in Terraform state from `InstallationInfo.InstalledKibanaSpaceId` when the plan value is unknown, instead of leaving it null.
> - On read, backfills `space_id` from the same field when the current state value is null.
> - Adds a shared helper `installedKibanaSpaceID` in [create.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2582/files#diff-627dc321ae80ea785f74968f1d0c20a02da9de1c7e4e2a6aeee5d5c67e54095a) used by both create and read paths.
> - Acceptance test in [acc_test.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2582/files#diff-aaef4b65af6a17f4aea136d36cdc736d23ed916a5c5d331172eac30b25e776bd) verifies `space_id` is read back as `default` on Kibana ≥ 8.10.0.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 137a72e.</sup>
> 
>


tobio:
I suspect there's more nuance required here. IIUC this will create an endless diff if a user attempts to install an integration into multiple spaces since `InstalledKibanaSpaceId` is the _first_ space the integration was installed into. 

To fix this properly, we'd need to:
* Combine `InstalledKibanaSpaceId` and `AdditionalKibanaSpaces` (or whatever the field is actually called)
* Check that the integration is installed in one of those locations
* If it's not, the resource is removed so it can be properly installed
* In create, check if the integration is already installed in another space, if it is, call [install Kibana assets](https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-fleet-epm-packages-pkgname-pkgversion-kibana-assets)

vatsaldesai93:
thanks for taking a look @tobio, still getting familiar with the code ..

the bug we hit happened after upgrading, `space_id` stayed `null` in state for resources that had no explicit space configured and since `space_id` is `Computed + RequiresReplace`, tf would generate a replacement plan, and fleet would block this replacement attempt because active policies reference the package

my intention with the added `IsNull()` and `IsUnknown()` guards is to ensure we only read `InstalledKibanaSpaceId` when `space_id` is blank in state. once set, `IsNull()` is false on all subsequent reads so `InstalledKibanaSpaceId` is never consulted again, so the endless diff wouldnt apply

is it reasonable to scope this pr to that fix and track the broader multi-space gap separately, that gap predates this change and is unaffected by it

tobio (approved):
Thanks for adding this in!

Issue #966, opened 2024-12-19
[Bug] Error 400 resource_already_exists_exception when replacing an index


BardiaN:
**Describe the bug**
Changing `mapping_coerce` on an index's mapping will trigger a replace. But the provider tries first to create the index which is not possible as it already exists.
I even tried adding `create_before_destroy = false` which is the default terraform behavior but it didn't work as expected.

```
lifecycle {
    create_before_destroy = false
  }
```

**To Reproduce**
1. TF configuration used

```terraform
resource "elasticstack_elasticsearch_index" "index" {
  name                = "index-one"

  mappings = jsonencode({
    properties = {
      name = {
        type = "text"
      }
   }
   })
  number_of_shards   = 1
  number_of_replicas = 1
  search_idle_after  = "20s"
  mapping_coerce     = false
}
```

2. `terraform apply` to create the index
3. Then change the `mapping_coerce` to true
4. `terraform apply` to apply the coerce
5. The plan will look like this :

```
Plan: 1 to add, 0 to change, 1 to destroy.
```
6. Then enter `yes` to apply.
7. Provider will try to create the resource first which causes the error:

```
elasticstack_elasticsearch_index.index: Creating...


╷
│ Error: Unable to create index: index
│ 
│   with elasticstack_elasticsearch_index.index,
│   on ./elastic-search.tf line 3, in resource "elasticstack_elasticsearch_index" "index":
│    3: resource "elasticstack_elasticsearch_index" "index" {
│ 
│ Failed with: {"error":{"root_cause":[{"type":"resource_already_exists_exception","reason":"index
│ [index/<ID>] already
│ exists","index_uuid":"<ID>","index":"index"}],"type":"resource_already_exists_exception","reason":"index
│ [index/<ID>] already
│ exists","index_uuid":"<ID>","index":"index"},"status":400}
╵
```

**Expected behavior**
The provider should respect `create_before_destroy`. In this case it means it has to destroy the index first and then create the new one.

**Versions (please complete the following information):**
 - OS: Mac OS Sequoia 15.1 (24B83)
 - Terraform v1.8.5
 - Provider version 0.11.12
 - Elasticsearch Version 7.17

**Additional context**

I also tried using alias, had the same result

tobio:
I'm struggling to reproduce this behaviour. 

Following the repro steps exactly fails at 4. since `deletion_protection` has not been set on the index. Have you updated the TF definition to set that option? 

[...]

PR #2589, opened 2026-05-01
feat(elasticsearch_index): add use_existing for adopting existing indices at create

tobio:
## Summary

Adds optional `use_existing` to `elasticstack_elasticsearch_index` ([`openspec/changes/elasticsearch-index-use-existing`](openspec/changes/elasticsearch-index-use-existing)) so create can opt in to **adopting** an index that already exists (replacement races and adopt-without-import), with strict static-setting comparison and reconciliation via the existing update helpers.

Closes / relates to https://github.com/elastic/terraform-provider-elasticstack/issues/966

## Changelog

Customer impact: enhancement

Summary: Add optional `use_existing` on `elasticstack_elasticsearch_index` to adopt an existing index at create instead of failing on duplicate.

## Behavior (high level)

- `use_existing = false` (default): unchanged create path (no pre-create Get Index).
- `use_existing = true` + static `name`: Get Index → adopt (reconcile aliases / dynamic settings / mappings) or error on static-setting mismatches without mutating the cluster; adoption warning on success.
- `use_existing = true` + date math `name`: warning, skip existence check, normal create.

## Testing

- Unit tests for `compareStaticSettings`, formatting, synthetic `populateFromAPI` path
- Acceptance tests: adopt success, mismatch error, date math, 404 fall-through, alias reconcile, template mapping superset

## OpenSpec

Change: `elasticsearch-index-use-existing` (delta spec under the change folder; not yet merged to main specs).

> [!NOTE]
> ### Add `use_existing` attribute to `elasticstack_elasticsearch_index` for adopting existing indices at create time
> - When `use_existing = true` and the index name is a static (non-date-math) name, the create path checks for an existing index via the ES API and adopts it instead of creating a new one.
> - Adoption validates that any explicitly configured static settings (e.g. `number_of_shards`) match the existing index; if they differ, an error is returned listing configured vs. actual values without mutating the cluster.
> - On successful adoption, aliases, dynamic settings, and mappings are reconciled to match the plan, and state is populated from the existing index with a warning that adoption occurred.
> - If the index does not already exist, the resource falls through to normal creation. Date-math index names emit a warning and always fall through to normal creation.
> - Acceptance tests cover fallthrough, alias reconciliation, template-injected mapping tolerance, static-setting mismatch errors, and date-math behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 35ac258.</sup>
> 
>


tobio:
Addressed review feedback in \`6f784153\`:

- **Adopt path / unknown alias & mappings:** Reconcile aliases only when \`plan.Alias\` is known; run \`updateMappings\` only when \`plan.Mappings\` is known (same \`typeutils.IsKnown\` pattern as \`toAPIModel\`).
- **Empty JSON array \`[]\` for sort order / sort field set:** \`json.Unmarshal\` of \`[]\` now returns empty configuration instead of treating the literal \`[]\` as a value.
- **Final read after adoption:** Error if \`readIndex\` returns a nil model (index disappeared between updates and read).

Also fixed **PR Changelog Check** by adding the required \`## Changelog\` section to the PR body earlier in this session.

PR #2608, opened 2026-05-01
feat(fleet): space-aware integration asset install/uninstall

tobio:
## Summary

Makes `elasticstack_fleet_integration` correctly manage space-scoped Kibana assets when `space_id` is configured on Kibana >= 8.15.0.

### Problem
Fleet treats integration package installation as global (one saved object), but Kibana assets (dashboards, visualizations, searches) are space-scoped. After installing a package in Space A, installing the same package in Space B silently no-ops because Kibana's install endpoint checks global install status and returns "already installed". Terraform sees `status: installed`, but the target space never receives assets.

### Solution
- **Read**: On Kibana >= 8.15.0, a package is only considered "installed" when it has assets in the target space (via `InstalledKibanaSpaceId` / `AdditionalSpacesInstalledKibana`). Otherwise, the resource is removed from state, triggering re-creation.
- **Create/Update**: Pre-flights with `GetPackage`. If the package is installed elsewhere, calls `POST /api/fleet/epm/packages/{pkg}/{version}/kibana_assets` scoped to the target space to install assets there. Otherwise, uses the regular install API.
- **Delete**: Pre-flights to check if the package spans multiple spaces. If multi-space, calls `DELETE /api/fleet/epm/packages/{pkg}/{version}/kibana_assets` scoped to the target space. If single-space or old Kibana, uses the global uninstall API.
- **Version gate**: Behavior is gated on Kibana >= 8.15.0. Older versions fall back to existing non-space-aware behavior without error.

### Changes
- `internal/clients/fleet/fleet.go` — adds `InstallKibanaAssets`, `DeleteKibanaAssets` helpers
- `internal/fleet/integration/read.go` — `fleetPackageInstalled` now supports strict space checking
- `internal/fleet/integration/create.go` — pre-flight logic for space-aware install path
- `internal/fleet/integration/delete.go` — multi-space detection and `DeleteKibanaAssets` fallback
- `internal/fleet/integration/resource.go` — `MinVersionSpaceAwareIntegration = 8.15.0`
- `internal/fleet/integration/acc_test.go` — multi-space install, delete, and drift tests

## Changelog
Customer impact: enhancement
Summary: Add space-aware Kibana asset management for elasticstack_fleet_integration on Kibana >= 8.15.0

### Testing
- `make build` ✅
- `make check-lint` ✅
- `make check-openspec` ✅
- Acceptance tests for multi-space scenarios added and gated on 8.15.0

> [!NOTE]
> ### Add space-aware Kibana asset install and uninstall to fleet integration resource
> - On create, the integration resource now installs Kibana assets into a specific space via `fleet.InstallKibanaAssets` when the package is already globally installed but missing from the target space; waits for both global and space-scoped install to complete.
> - On delete, if the package is installed in multiple spaces, only the target space's Kibana assets are removed via `fleet.DeleteKibanaAssets`; a full uninstall only occurs when the target space is the last remaining install.
> - On read, a package is treated as absent when it is not installed in the resource's target space, enabling drift detection for missing Kibana assets.
> - All space-aware behavior is version-gated at Kibana 9.1.0 (`MinVersionSpaceAwareIntegration`); older servers receive a warning and fall back to global behavior.
> - Adds acceptance tests for multi-space install, partial delete, and drift detection scenarios.
> - Behavioral Change: `fleet.Uninstall` now accepts an explicit `force` bool and passes it to the Fleet API instead of always passing `nil` params.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 340ceab.</sup>
> 
>


tobio:
Thanks for the automated review. Responding to both points:

On 404 handling in InstallKibanaAssets: This follows the existing Uninstall pattern where 404 is treated as "not found / already gone" for idempotent operations. The create flow already pre-flights with GetPackage before calling InstallKibanaAssets, so a genuine 404 would be unexpected. Keeping it consistent with other delete/uninstall helpers in the same file.

On version mismatch in installedElsewhere: This is pre-existing behavior — Fleet's GET endpoint reports installed regardless of the version path segment, and the existing code already had this same limitation. The Read method now records InstallationInfo.Version in state, so version mismatches are surfaced on refresh. Adding version-aware pre-flight here would be a larger change that's orthogonal to this space-aware feature.

tobio:
@github-actions re the CRITICAL issue about testdata path resolution:

The testdata directories follow the exact same pattern as all existing tests in this file (e.g. `TestAccResourceIntegration`, `TestAccResourceIntegrationDeleted`, etc.). `acctest.NamedTestCaseDirectory("create")` delegates to `config.TestNameDirectory()` from terraform-plugin-testing, which resolves `testdata/<TestFunctionName>/<name>` relative to the caller test file directory. Since `acc_test.go` is in `internal/fleet/integration/", the runtime lookup correctly finds:

- `internal/fleet/integration/testdata/TestAccResourceIntegration_MultiSpaceInstall/create/main.tf`
- `internal/fleet/integration/testdata/TestAccResourceIntegration_MultiSpaceDelete/create/main.tf`
- `internal/fleet/integration/testdata/TestAccResourceIntegration_MultiSpaceDelete/space_a_only/main.tf`
- `internal/fleet/integration/testdata/TestAccResourceIntegration_SpaceAwareDrift/create/main.tf`

`go test -c ./internal/fleet/integration/...` also compiles cleanly with no path issues. Every existing test in `acc_test.go` uses the same `NamedTestCaseDirectory` pattern and works correctly in CI.

This should be a **false positive** — the testdata is where terraform-plugin-testing expects it.

PR #2669, opened 2026-05-05
fix(ml): poll for closed state before deleting anomaly detection job

tobio:
## Summary

Fixes a race condition in the `elasticstack_elasticsearch_ml_anomaly_detection_job` delete path that causes HTTP 409 `version_conflict_engine_exception` when a job is deleted from `opened` state.

## Root cause

`CloseJob` was called immediately before `DeleteJob` without waiting for the close transition to commit in Elasticsearch's internal `.ml-config` index. Elasticsearch uses optimistic concurrency control on that index: if the close's state transition commits after `DeleteJob` reads the seqNo but before it writes, the delete fails with `version_conflict_engine_exception`.

## Fix

- **New `WaitForMLJobClosed` client helper** (`internal/clients/elasticsearch/ml_job.go`): polls `GetMLJobStats` via `asyncutils.WaitForStateTransition` until the job reports `closed` state or is not found (job gone). Bounded by the Terraform delete context timeout (default 20 min).
- **Updated delete flow** (`internal/elasticsearch/ml/anomalydetectionjob/delete.go`): calls `WaitForMLJobClosed` between `CloseJob` and `DeleteJob`. If the wait fails, logs a warning and continues. If `DeleteJob` fails, retries once with `force=true`; surfaces the retry error as a diagnostic if that also fails.
- **Spec updated** (`openspec/specs/elasticsearch-ml-anomaly-detection-job/spec.md`): REQ-021–REQ-022 updated to require polling and retry.

## Related

- OpenSpec change: `fix-ml-job-delete-close-race`
- Fixes flaky teardown in `TestAccResourceMLJobStateExplicitConnection`

## Changelog
Customer impact: fix
Summary: Poll for job closed state before deleting ML anomaly detection job to eliminate HTTP 409 version_conflict_engine_exception on teardown

> [!NOTE]
> ### Fix race condition in anomaly detection job deletion by polling for closed state
> - After closing an ML job (force=true), the delete handler now polls `GetMLJobStats` via `WaitForMLJobClosed` until the job reports `closed` or disappears, bounded by a configurable delete timeout (default 20 minutes).
> - If polling fails due to timeout, a diagnostic error is returned; otherwise the warning is logged and deletion proceeds regardless.
> - On first delete failure, the handler retries once with `force=true`; 404 responses are treated as success at every step.
> - Adds a `timeouts` block to the resource schema and model, exposing a `delete` duration attribute in the provider docs.
> - Behavioral Change: delete operations now block until the job is confirmed closed before issuing `DeleteJob`, which may increase total delete time but prevents premature deletion failures.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4dc83f1.</sup>
> 
>


tobio:
**Review feedback addressed** (commits 944d666a, 59e6ca9b, 0a018893):

- **Immediate pre-check** in `WaitForMLJobClosed` before the 2s poll loop — jobs already `closed` return without latency penalty
- **Typed `jobstate.Closed` constant** instead of string literal in the state comparison
- **404 treated as success** on both `DeleteJob` attempts (idempotent delete, mirrors `DeleteDatafeed` pattern)
- **Context expiry handled explicitly** — when the wait times out due to context cancellation, a clear diagnostic is surfaced instead of a confusing delete error
- **design.md updated** to remove the retry Non-Goal and document the rationale for the `force=true` fallback

internal/elasticsearch/ml/anomalydetectionjob/delete.go
```diff
@@ -68,11 +71,44 @@ func (r *anomalyDetectionJobResource) delete(ctx context.Context, req resource.D
 		// Continue with deletion even if close fails, as the job might already be closed
 	}
 
-	// Delete the ML job
+	// Wait for the job to reach closed state before deleting. Elasticsearch uses
+	// optimistic concurrency control on .ml-config: if CloseJob's state
+	// transition commits after DeleteJob reads the seqNo but before it writes,
+	// the delete fails with HTTP 409 version_conflict_engine_exception.
+	if err := elasticsearch.WaitForMLJobClosed(ctx, client, jobID); err != nil {
```
tobio, line 88: Since we're adding this wait we need to add a `timeouts` attribute similar to the `jobstate` resource. 

`timeouts` should only support Delete timeouts for now. 

Add the timeouts, and update the change proposal to be consistent.

tobio:
Done in 5ced8f9b — added a `timeouts` block (Delete only, default 20 min) matching the `jobstate` resource pattern. The delete context is now bounded by the configured timeout, so the `WaitForMLJobClosed` poll and both `DeleteJob` attempts respect it.

PR #2714, opened 2026-05-06
fix(ilm): delete policy even when referenced by indices

### LINKED ISSUE #1999: [Bug] Ensure integration policies can be safely deleted
> There is an interesting cross dependency when we deploy a cluster and an elastic agent sends data into the cluster with some integration managed datastreams/index templates, the destroy of the resource fails on ILM policies, indices, and some other related resources (we don't have an extensive list). We found a workaround with a local-exec bash script cleaning up everything but seems suboptimal, would be nice to have a proper dependency between all the resources even for destroy. The main issue here is that the integration itself is managed by terraform and we defined proper dependency chain between the terraform resources, but some related things are automagically created and not managed by terraform and those for most likely to prevent data deletion don't get destroyed with the integration, even when we plan to destroy a whole deployment.

tobio:
Closes #1999

When a Fleet integration is installed it creates index templates, component templates and ingest pipelines. If data streams or indices inherit from those templates and an ILM policy is attached via `elasticstack_elasticsearch_index_template_ilm_attachment`, the backing indices carry an `index.lifecycle.name` reference. On `terraform destroy`, Fleet's package uninstall succeeds but the backing indices remain; deleting the Terraform-managed ILM policy then fails with:

> Cannot delete policy [X]. It is in use by one or more indices: [.ds-...]

This PR makes ILM policy Delete scan all indices for references to the policy and clear `index.lifecycle.name` on matching indices before calling `DELETE /_ilm/policy/{name}`.

## Changes

- Adds `GetIndicesWithILMPolicy` and `ClearILMPolicyFromIndices` helpers to the ES client layer
- Updates the `Delete` handler for `elasticstack_elasticsearch_index_lifecycle` to scan-and-clear before deleting
- Adds acceptance tests confirming the fix works for both Fleet-managed and regular-index scenarios

## Changelog
Customer impact: fix
Summary: Fix ILM policy delete failures when the policy is still referenced by indices (e.g. Fleet-managed data stream backing indices)

## Acceptance tests

- `TestAccResourceILM_deleteWithReferencedIndex` — creates a policy + index with lifecycle reference, asserts successful destroy
- `TestAccResourceIntegration_destroyWithILMCrossDependency` — updated to expect success instead of error

## Verification

- [x] `go build` passes
- [x] Unit tests pass
- [x] Acceptance tests pass against local stack (docker-fleet)
- [x] Delta spec synced to main specs

> [!NOTE]
> ### Fix ILM policy deletion when indices reference the policy
> - The `Resource.Delete` handler in [delete.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2714/files#diff-6b11651d3f9eba8942947bbea942851a3350559059023ac985891d23e8072bb7) now clears `index.lifecycle.name` from any referencing indices before calling the Delete Lifecycle API, preventing failures when a policy is still in use.
> - Two new ES client helpers are added in [index.go](https://github.com/elastic/terraform-provider-elasticstack/pull/2714/files#diff-89907d8b41976616c950d51a30bd3de77069eea67eabc97c17f269ddc6ab8ec6): `GetIndicesWithILMPolicy` (queries `GET /_all/_settings/index.lifecycle.name?flat_settings=true`) and `ClearILMPolicyFromIndices` (issues `PUT /{indices}/_settings` with `index.lifecycle.name: null`).
> - A new acceptance test `TestAccResourceILM_deleteWithReferencedIndex` and an updated fleet integration test confirm end-to-end destroy behavior when an index references the policy.
> - Behavioral Change: Deleting an ILM policy now silently removes the policy association from any referencing indices as a side effect before deletion.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 06b9f69.</sup>
> 
>


tobio:
@copilot-pull-request-reviewer thanks for the review. Responses to the raised points:

1. **Side effects (delete.go:59)** — This is the intended behavior per the OpenSpec design. The scan-and-null approach intentionally targets indices directly because backing indices can outlive template changes. The alternative (reverse lookup through templates) was rejected for this reason.

2. **Permission change (delete.go:57)** — The standard Elasticsearch ILM admin role () already implies , so this is not a breaking change for standard accounts.

3. **Missing backing index assertion (acc_test.go:626)** — Addressed in 04d48194. Added a PreConfig verification that queries  and asserts at least one backing index has the lifecycle reference before the destroy step.

4. **Missing delete-level early-return test (delete.go:56)** — The short-circuit is a thin wrapper around the helper, which has unit-test coverage. Delete-level integration tests for this exact branch would require mocking the full framework resource layer, which the provider does not currently do.

5. **Request shape assertion (index_ilm_test.go:75)** — Addressed in 04d48194. Added path and query-parameter assertions validating .

6. **Performance at scale (index.go:146)** — Accepted trade-off. The request is scoped to a single setting key, keeping response sizes small. Future enhancement could add pagination if profiling shows a bottleneck.

PR #2877, opened 2026-05-11
feat(elasticsearch): custom_rules.scope for ML anomaly detection job

edsavage:
## Summary

Adds optional `scope` on `elasticstack_elasticsearch_ml_anomaly_detection_job` detector `custom_rules` blocks (map of analysis field names to ML `filter_id` and optional `filter_type`). Updates schema, API models, read/write paths, docs, and changelog. Adds plan-time `ValidateConfig` so each custom rule must have either a non-empty `scope` or at least one `conditions` entry when both are known at plan time.

**Acceptance tests** (standard `TF_ACC=1` plus Elasticsearch/Kibana env from [testing docs](https://github.com/elastic/terraform-provider-elasticstack/blob/main/dev-docs/high-level/testing.md); no extra env vars):

**`TestAccResourceAnomalyDetectionJobCustomRulesScope`** — Creates the ML filter **out of band** with the Elasticsearch **Put filter** API (`PUT _ml/filters/{filter_id}`), then applies Terraform that references that `filter_id` in `scope` (no `elasticstack_elasticsearch_ml_filter` resource). Exercises create and in-place update of `filter_type` (`include` → `exclude`) and asserts state round-trips.

**`TestAccResourceAnomalyDetectionJobCustomRulesScope_missingFilterOnCreate`** — Applies a job whose `scope` references a **never-created** filter id. Elasticsearch may still accept **Put job**; the test then calls **Open job** and asserts it **fails** with an error consistent with a missing/invalid ML filter (mirrors “filter deleted or never existed” before the job can run).

**`TestAccResourceAnomalyDetectionJobCustomRulesScope_missingFilterOnUpdate`** — Creates the job with a **valid** filter, updates config to a **non-existent** filter id, then asserts **Open job** fails the same way (covers bad or stale `filter_id` after update).\n\nSplit out from #1970 for clearer review (scope change vs ML filter resource).

## Notes

- Changelog file: single bullet under `[Unreleased]` → `### Changes`.

## Changelog

Customer impact: enhancement
Summary: Add optional `scope` on detector `custom_rules` for ML anomaly detection jobs (map analysis field names to ML `filter_id` and optional `filter_type`).


internal/elasticsearch/ml/anomalydetectionjob/acc_test.go
```diff
@@ -331,6 +378,129 @@ func TestAccResourceAnomalyDetectionJobCustomRules(t *testing.T) {
 	})
 }
 
+// TestAccResourceAnomalyDetectionJobCustomRulesScope covers detector custom_rules.scope
+// (ML filter references per analysis field), including round-trip to Elasticsearch.
+//
+// The referenced ML filter is created out-of-band via the Elasticsearch ML APIs before apply,
+// so this test does not require the elasticstack_elasticsearch_ml_filter resource.
+func TestAccResourceAnomalyDetectionJobCustomRulesScope(t *testing.T) {
+	acctest.PreCheck(t)
```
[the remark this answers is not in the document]
edsavage: Earlier review items (ValidateConfig + docs vs ml_filter) are already addressed on the branch. Latest follow-up: we intentionally do not skip scope ACC tests behind TF_ACC_ML_SCOPE_TEST—filters are created with the Elasticsearch Put filter API in the test harness; PR description matches. Added LengthAtLeast(1) on scope.filter_id. Fixed acctestconfigdirlint by giving each test its own testdata/<TestName>/… layout instead of a custom ConfigDirectory helper.

internal/elasticsearch/ml/anomalydetectionjob/schema.go
tobio, line 209:
```suggestion
									NestedObject: schema.NestedAttributeObject{
										Validators: []validator.Object{
											objectvalidator.ExactlyOneOf(
												path.MatchRoot("scope"),
												path.MatchRoot("conditions"),
											),
										},
```

internal/elasticsearch/ml/anomalydetectionjob/schema.go
```diff
@@ -240,6 +240,32 @@ func getSchema(ctx context.Context) schema.Schema {
 													},
 												},
 											},
+											"scope": schema.MapNestedAttribute{
+												MarkdownDescription: "Maps an analysis field name " +
+													"(typically matching `by_field_name`, `over_field_name`, or `partition_field_name` on the detector) " +
+													"to an ML filter reference. Each `filter_id` must identify an ML filter that already exists in the cluster " +
+													"(for example, created using the Elasticsearch ML filter APIs). " +
+													"A rule must have either a non-empty `scope` or at least one `conditions` entry.",
+												Optional: true,
```
tobio, line 257:
```suggestion
												Optional: true,
												Validators: []validator.Map{
													mapvalidator.SizeAtLeast(1),
													mapvalidator.NoNullValues(),
												},
```

internal/elasticsearch/ml/anomalydetectionjob/resource.go
```diff
@@ -55,6 +56,17 @@ func NewAnomalyDetectionJobResource() resource.Resource {
...
+// known at plan time, matching Elasticsearch validation for ML custom rules.
+func (r *anomalyDetectionJobResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
+	var config TFModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	resp.Diagnostics.Append(validateConfigCustomRules(ctx, &config)...)
+}
+
```
tobio:
```suggestion
```

It looks like we can replace this entirely with built in validators rather than defining our own.

tobio (commented):
Thanks for splitting this out. Can we simplify the validation here a little.

edsavage:
> Thanks for splitting this out. Can we simplify the validation here a little.

Thanks Toby, will do.

internal/elasticsearch/ml/anomalydetectionjob/schema.go
edsavage: Thanks for the suggestion Toby. I had to double-check this one against a live job: Elasticsearch accepts both a non-empty scope and conditions on the same custom rule (they’re not mutually exclusive). So `objectvalidator.ExactlyOneOf` would reject valid configs. For that reason I'm sticking with validation that only rejects the case where both are empty, plus docs that call out that scope and conditions can be combined.

internal/elasticsearch/ml/anomalydetectionjob/resource.go
edsavage: I've used built-in validators where they match Elasticsearch (non-empty scope when set), and kept minimal `ValidateConfig` for the part Elasticsearch documents as a joint requirement on the rule, which isn’t the same as `ExactlyOneOf` or other one-attribute validators.

tobio (approved):

PR #2882, opened 2026-05-12
fix(clients): honor framework elasticsearch_connection credentials

edsavage:
## Summary
- Build Framework `elasticsearch_connection` scoped Elasticsearch clients without overwriting non-empty connection-block credentials with `ELASTICSEARCH_*` environment variables.
- Add focused unit tests covering env-vs-config precedence for scoped Framework connections.

## Why
Acceptance tests (and many real deployments) set `ELASTICSEARCH_USERNAME`/`ELASTICSEARCH_PASSWORD` in the process environment. Framework resources that configure `elasticsearch_connection` need those explicit credentials to be used as-is rather than being replaced by the provider process env.

## Test plan
- `go test ./internal/clients/config/... ./internal/clients/...`

## Changelog
Customer impact: fix
Summary: Scoped Framework `elasticsearch_connection` Elasticsearch clients honor Terraform credentials and auth mode and no longer pick up conflicting `ELASTICSEARCH_*` environment variables.

Made with [Cursor](https://cursor.com)

PR #2883, opened 2026-05-12
test(ml): add forbidden ML filter update acceptance case

edsavage:
## Summary
- Add `TestAccResourceMLFilterUpdateForbiddenReadOnlyUser` plus testdata that creates a filter with the default provider connection, then attempts an update using `elasticsearch_connection` with a user that has no effective cluster privileges.

## Depends on
- #2882 — without the scoped Framework `elasticsearch_connection` credential handling, acceptance runs that set `ELASTICSEARCH_USERNAME`/`ELASTICSEARCH_PASSWORD` can silently use the admin credentials instead of the scoped user.

## Merge / review notes
This branch is stacked on top of the branch for #2882. Please merge **#2882 first**; after it lands on `main`, this PR should be **rebased onto `main`** (or updated) so the diff only contains the ML acceptance test + testdata.

## Test plan
- `go test -c ./internal/elasticsearch/ml/filter/...`
- `./bin/golangci-lint-custom run --enable-only=acctestconfigdirlint ./internal/elasticsearch/ml/filter/...`
- `TF_ACC=1 go test ./internal/elasticsearch/ml/filter/... -run TestAccResourceMLFilterUpdateForbiddenReadOnlyUser -count=1 -v` (requires a cluster)


edsavage:
Closing: the ML filter acceptance test and testdata now live on **#1970** (`feat/ml-filter`) as commit `749d859e`, since they depend on the ML filter resource work there.

The scoped `elasticsearch_connection` credential fix remains in **#2882**; merge **#2882** first, then rebase/merge **#1970** onto current `main` so acceptance runs pick up the client fix.

PR #2913, opened 2026-05-13
feat: optional calendars on ML anomaly detection job

edsavage:
## Changelog

Customer impact: enhancement
Summary: Add optional `calendars` on `elasticstack_elasticsearch_ml_anomaly_detection_job`, syncing ML calendar–job membership via `_ml/calendars/{calendar_id}/jobs/{job_id}` after the job exists.

## Summary

Adds an optional `calendars` argument (set of strings) to `elasticstack_elasticsearch_ml_anomaly_detection_job`. Elasticsearch does not accept calendar membership on the Put job JSON; membership is applied after the job exists using `PUT _ml/calendars/{calendar_id}/jobs/{job_id}` and removed with the corresponding `DELETE`.

## Behavior

- **Create / update:** After the job is created or updated, the provider reconciles calendar membership against the configured set (adds and removes as needed).
- **Read:** Calendar assignments are discovered by paging `GET _ml/calendars` and collecting calendars whose `job_ids` include this job.
- **Delete:** Best-effort removal from calendars recorded in state before closing/deleting the job (404 ignored on unlink).

Calendar IDs are validated at plan time to match typical ML `calendar_id` rules (lowercase alphanumerics with hyphens/underscores).

## `CHANGELOG.md`

Under **[Unreleased] → Changes**:

> Add optional `calendars` on `elasticstack_elasticsearch_ml_anomaly_detection_job` to assign the job to ML calendars (via `_ml/calendars/{calendar_id}/jobs/{job_id}`).

## Docs

Regenerated resource documentation includes the new `calendars` attribute.

## Testing

- `go test ./internal/elasticsearch/ml/anomalydetectionjob/...`
- `make build` (includes lint and docs generation)

Acceptance tests against a live cluster were not run in this environment.

PR #2933, opened 2026-05-13
feat(ml): add elasticstack_elasticsearch_ml_calendar_job resource

edsavage:
## Summary

Adds **`elasticstack_elasticsearch_ml_calendar_job`**, a small resource that assigns **one** anomaly detection job to an ML calendar using `PUT _ml/calendars/{calendar_id}/jobs/{job_id}` and removes it on destroy. This matches how Elasticsearch applies calendar membership (not via the put-job body).

### Composite `id` / import

The provider composite ID only allows a single slash between cluster UUID and the resource segment, so the segment is **`calendar_id|job_id`** (pipe separator). Import ID: **`<cluster_uuid>/<calendar_id>|<job_id>`**.

### Tests

Acceptance tests create the ML calendar via the Elasticsearch API inside each `resource.TestCase.PreCheck`, **after** `acctest.PreCheck(t)`, so no cluster setup runs until acceptance prerequisites are satisfied. That avoids depending on `elasticstack_elasticsearch_ml_calendar` (not yet on `main`).

### Related

Supersedes the approach in [#2913](https://github.com/elastic/terraform-provider-elasticstack/pull/2913) (optional `calendars` on the anomaly detection job resource). [#2913](https://github.com/elastic/terraform-provider-elasticstack/pull/2913) is closed in favour of this pattern per maintainer feedback.

## Changelog

Customer impact: enhancement
Summary: Add `elasticstack_elasticsearch_ml_calendar_job` to assign one ML anomaly detection job to a calendar; entry added under Unreleased in `CHANGELOG.md`.


internal/elasticsearch/ml/calendar_job/models.go
```diff
@@ -0,0 +1,57 @@
...
+		return types.StringUnknown()
+	}
+	if m.CalendarID.IsNull() || m.JobID.IsNull() {
+		return types.StringNull()
+	}
+	c := m.CalendarID.ValueString()
+	j := m.JobID.ValueString()
+	if c == "" || j == "" {
+		return types.StringUnknown()
+	}
```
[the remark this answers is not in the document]
edsavage: Known-empty components now return types.StringNull() instead of StringUnknown().

internal/elasticsearch/ml/calendar_job/resource.go
```diff
@@ -0,0 +1,101 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if r.Client() == nil {
+		resp.Diagnostics.AddError("Client not configured", "Provider client is not configured")
+		return
+	}
+
+	client, connDiags := r.Client().GetElasticsearchClient(ctx, providerschema.ElasticsearchConnectionNullList())
```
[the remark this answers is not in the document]
edsavage: The schema / null-list fragility point was addressed by deriving types from the FW block + tests

internal/elasticsearch/ml/calendar_job/create.go
```diff
@@ -0,0 +1,64 @@
...
+
+	if _, err := typedClient.Ml.PutCalendarJob(calendarID, jobID).Do(ctx); err != nil {
+		diags.AddError(
+			"Failed to assign ML job to calendar",
+			fmt.Sprintf("Unable to assign job %q to calendar %q: %s", jobID, calendarID, err.Error()),
+		)
+		return entitycore.WriteResult[TFModel]{Model: plan}, diags
+	}
+
+	compID, idDiags := client.ID(ctx, calendarID+"|"+jobID)
```
tobio:
```suggestion
	compID, idDiags := client.ID(ctx, calendarID+"/"+jobID)
```

You're adding support for multi-segment composite IDs in the other PR, we should be consistent.

internal/elasticsearch/ml/calendar_job/modify_plan.go
```diff
@@ -0,0 +1,49 @@
...
+package calendar_job
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+func (r *calendarJobResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
```
tobio: This will be duplicated when both the calendar PR and this PR are merged. 

Can you:
- Make this an attribute level plan modifier attached to `elasticsearch_connection`
- Move it to `utils/planmodifers`
- Either attach it everywhere in the schema helper, or add it as an override for these resources. There's a unit test that validates these connection subschemas match the helper so you'd need to update that test to handle the extra plan modifier.

internal/elasticsearch/ml/calendar_job/update.go
```diff
@@ -0,0 +1,34 @@
...
+	"github.com/elastic/terraform-provider-elasticstack/internal/entitycore"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+)
+
+func updateCalendarJobNoOp(
+	_ context.Context,
+	_ *clients.ElasticsearchScopedClient,
+	req entitycore.WriteRequest[TFModel],
+) (entitycore.WriteResult[TFModel], diag.Diagnostics) {
+	return entitycore.WriteResult[TFModel]{Model: req.Plan}, nil
```
tobio: Can we return an error diag here, this should never be called and we've broken the schema if it is suddenly part of the lifecycle.

CHANGELOG.md.tmp
tobio, line 1: 🔫

internal/elasticsearch/ml/calendar_job/create.go
edsavage: Done — composite ID now uses `calendarID+"/"+jobID` in create, read, import, and `GetResourceID()`, aligned with #1969.

internal/elasticsearch/ml/calendar_job/modify_plan.go
edsavage: Moved to `internal/utils/planmodifiers` (`ConnectionBlockRequiresReplaceOnChange` / `ModifyPlanRequiresReplaceOnConnectionChange`). Provider-schema connection blocks cannot host resource plan modifiers, so `ElasticsearchResource` and `KibanaResource` envelopes now implement `ModifyPlan` and append `RequiresReplace` for `elasticsearch_connection` / `kibana_connection` when the block changes. Removed per-resource `modify_plan.go` from calendar, calendar_event, and calendar_job.

internal/elasticsearch/ml/calendar_job/update.go
edsavage: Update now returns an error diagnostic (update is not supported; changing `calendar_id`/`job_id` or `elasticsearch_connection` should force replace).

CHANGELOG.md.tmp
edsavage: Removed `CHANGELOG.md.tmp`.

internal/elasticsearch/ml/calendar_job/schema.go
```diff
@@ -0,0 +1,95 @@
...
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+				Validators: []validator.String{
+					stringvalidator.LengthBetween(1, 64),
+					stringvalidator.RegexMatches(
+						mlCalendarOrJobPathIDRegexp,
+						jobIDAllowedCharsMessage,
+					),
+				},
```
tobio: nit: We've got essentially this pattern duplicated across a bunch of schema elements in this resource, and the other job/calendar based resources. 

We should pull this out to something like `ml.IDValidator()`

internal/elasticsearch/ml/calendar_job/update.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"context"
+
+	"github.com/elastic/terraform-provider-elasticstack/internal/clients"
+	"github.com/elastic/terraform-provider-elasticstack/internal/entitycore"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+)
+
+func updateCalendarJobNoOp(
```
tobio: Can be replaced with `entitycore.UpdateNotSupportedWriteCallback`

internal/entitycore/resource_envelope.go
```diff
@@ -228,6 +230,32 @@ func (r *ElasticsearchResource[T]) Schema(ctx context.Context, _ resource.Schema
 	resp.Schema = schema
 }
 
+// ModifyPlan implements [resource.ResourceWithModifyPlan], forcing replacement when
+// the per-resource elasticsearch_connection block changes (provider-schema blocks
+// cannot host resource plan modifiers; see [planmodifiers.ModifyPlanRequiresReplaceOnConnectionChange]).
+func (r *ElasticsearchResource[T]) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
```
tobio: So coming back to this, this isn't the correct behaviour and we should remove these plan modifiers. 

With this plan modifier in place, a replacement would be triggered if:
* The user/password changed, for example a password rotation.
* The API key changed, for example, it was rotated.
* The Elasticsearch hostname changed, for example the deployment alias in ESS was updated. 

Forcing replacement for any of these examples is incorrect and potentially causes data loss (given the ML focus on this PR, losing job history is at best far from ideal, at worse considered production data loss). 

Arguably we could reach out and grab the cluster ID and trigger a replacement if the cluster ID had changed, but that's still arguably incorrect. Snapshot/restore would create an 'identical' new cluster that users would want the provider to just work with.

tobio (approved):

Issue #2962, opened 2026-05-14
Support `data_stream_options` / failure store on `elasticstack_elasticsearch_component_template`


ottramst:
## Summary

Add a `data_stream_options { failure_store { ... } }` block to the `elasticstack_elasticsearch_component_template` resource, mirroring the block already supported on `elasticstack_elasticsearch_index_template` since v0.15.0 (#2509, originally requested in #2201).

## Motivation

Elasticsearch supports `data_stream_options` (notably `failure_store`) inside a component template's `template` body — see elastic/elasticsearch#117357 (merged in 8.18.0 / 9.0.0; the relevant API surface is `server/.../metadata/Template.java`'s `ResettableValue<DataStreamOptions.Template> dataStreamOptions`, with XContent key `data_stream_options`). That inner `Template` struct is the same shape embedded in both `ComponentTemplate` and `ComposableIndexTemplate`, so ES treats this field identically in either container.

The natural place to configure it is through Fleet's `@custom` component template extension point. For example, enabling a failure store on the managed `logs-kubernetes.container_logs` data stream requires editing the `logs-kubernetes.container_logs@custom` component template that the Kubernetes integration composes — owning the managed index template isn't desirable, and `@custom` is the supported Fleet hook. Today the only way to do this Terraform-natively would be to clone the managed index template just to add one field, which defeats the purpose of `@custom`.

## Current state

`internal/elasticsearch/index/componenttemplate/schema.go` exposes only `mappings`, `settings`, and `alias` inside the `template` block. `data_stream_options` is absent.

Good news: the wire model is already shared. `internal/models/models.go` defines `Template.DataStreamOptions *DataStreamOptions` (used by both index templates and component templates), so JSON serialization to ES is already correct. Only the Plugin Framework schema, plan/state model, and expand/flatten/validation paths in the `componenttemplate/` package need to be added.

## Proposed shape

Identical to the existing index template block:

```hcl
resource \"elasticstack_elasticsearch_component_template\" \"logs_kubernetes_custom\" {
  name = \"logs-kubernetes.container_logs@custom\"
  template {
    data_stream_options {
      failure_store {
        enabled = true
        lifecycle {
          data_retention = \"30d\"
        }
      }
    }
  }
}
```

Version-gated to `MinSupportedDataStreamOptionsVersion` (currently 9.1.0), reusing the constant already defined in `internal/elasticsearch/index/template_constants.go`.

## Implementation sketch

[...]

PR #2963, opened 2026-05-14
feat(component_template): add data_stream_options block

### LINKED ISSUE #2201: [Feature] Add data_stream_options support to elasticstack_elasticsearch_index_template
### Is your feature request related to a problem?

The `elasticstack_elasticsearch_index_template` resource's `template` block does not support `data_stream_options`, which was introduced in Elasticsearch 9.x. This prevents enabling the failure store for data streams via Terraform, requiring out-of-band API calls and creating configuration drift between Terraform state and actual cluster configuration.

### Describe the solution you'd like

Add a `data_stream_options` block inside the `template` block of `elasticstack_elasticsearch_index_template`. This should support at minimum the `failure_store.enabled` boolean.

Desired HCL syntax:

```hcl
resource "elasticstack_elasticsearch_index_template" "example" {
  name           = "my-index-template"
  index_patterns = ["my-datastream-*"]

  template {
    data_stream_options {
      failure_store {
        enabled = true
      }
    }
  }

  data_stream {}
}
```

The corresponding Elasticsearch API:

```json
PUT _index_template/my-index-template
{
  "index_patterns": ["my-datastream-*"],
  "data_stream": {},
  "template": {
    "data_stream_options": {
      "failure_store": {
        "enabled": true
      }
    }
  }
}
```

### Describe alternatives you've considered

Using `terraform_data` with `local-exec` to GET the index template from Elasticsearch, patch in `data_stream_options` via `jq`, and PUT it back. This works but breaks infrastructure-as-code principles, requires `curl`/`jq` as dependencies, and creates drift between Terraform state and actual cluster configuration.

### Additional context

- Elasticsearch documentation: https://www.elastic.co/docs/manage-data/data-store/data-streams/failure-store
- The `data_stream_options` field is part of the index template API starting in Elasticsearch 9.x
- Provider version tested: 0.14.3
- Ensure all fields within `data_stream_options` and the nested `failure_store` models are covered. Examine the API docs for a full definition of these models.

ottramst:
## Summary

Adds the `data_stream_options { failure_store { ... } }` block to `elasticstack_elasticsearch_component_template`, mirroring the block already supported on `elasticstack_elasticsearch_index_template` since v0.15.0 (#2509). Filed as #2962.

```hcl
resource "elasticstack_elasticsearch_component_template" "logs_myapp_custom" {
  name = "logs-myapp@custom"

  template {
    data_stream_options {
      failure_store {
        enabled = true
        lifecycle {
          data_retention = "30d"
        }
      }
    }
  }
}
```

Requires Elasticsearch >= 9.1.0 (gated via the existing `index.MinSupportedDataStreamOptionsVersion`).

## Motivation

Elasticsearch supports `data_stream_options` (notably `failure_store`) inside a component template's `template` body — see elastic/elasticsearch#117357. The `Template` struct that carries this field is the same shape embedded in both `ComponentTemplate` and `ComposableIndexTemplate`, so JSON serialization to ES already works for component templates. Only the Plugin Framework surface area was missing.

The primary use case is Fleet's `@custom` extension point: enabling the failure store on a managed integration's data stream (e.g. `logs-kubernetes.container_logs`) requires editing the `…@custom` component template that Fleet composes, since owning the managed index template isn't desirable.

## Implementation notes

Following Option A from #2962 — extracted a shared `internal/elasticsearch/index/datastreamoptions/` package so the block's schema builders, models, expand/flatten helpers, validation, and version gating live in one place. Both `template/` and `componenttemplate/` import it; the index template package is refactored to delegate (no behaviour change). Happy to switch to the duplication approach (Option B) if maintainers prefer a smaller, more localized diff.

### Files

- `internal/elasticsearch/index/datastreamoptions/` (new): shared package — schema block, models + `*AttrTypes()` helpers, `Expand`/`Flatten`, `ValidateRequiresFailureStore`, `EnforceMinServerVersion`.
- `internal/elasticsearch/index/template/`: refactored to use the shared package; data-source schema uses the shared description constants.
- `internal/elasticsearch/index/componenttemplate/`: wires the block into schema/models, expand/flatten, and create/update version-gating; adds a `ValidateConfig` method that delegates to the shared rule.
- `internal/elasticsearch/index/componenttemplate/state_upgrade.go`: `schemaVersion` bumped 1 → 2 with a v1 → v2 upgrader that injects null `data_stream_options` into existing template objects. The v0 → v1 upgrader is also updated to fill the new key so v0 state migrates directly to v2 (the framework dispatches to a single upgrader keyed by prior version — no chaining).
- `docs/resources/elasticsearch_component_template.md`: regenerated via `make docs-generate`.
- `examples/resources/elasticstack_elasticsearch_component_template/resource.tf`: adds a `failure_store` example.

### Tests

- New acceptance test `TestAccResourceComponentTemplateDataStreamOptions` (create / update_enabled / update_lifecycle / ImportStateVerify), mirroring `TestAccResourceIndexTemplateDataStreamOptions`.
- New state-upgrade tests for v1 → v2 (`injects_null_data_stream_options`, `template_null_noop`, `invalid_json`).
- Existing index template tests adapted for the shared-package names (no behaviour change).

## Changelog
Customer impact: enhancement
Summary: Add `template.data_stream_options` block to `elasticstack_elasticsearch_component_template` to configure the failure store on data streams composed from this component. Requires Elasticsearch >= 9.1.0.

## Test plan

Verified locally against an Elasticsearch 9.4.0 container.

- [x] `make build` — passes
- [x] `make golangci-lint` — 0 issues
- [x] `go test ./internal/elasticsearch/index/...` (unit tests) — passes
- [x] `TF_ACC=1 go test -run 'TestAccResourceComponentTemplateDataStreamOptions|TestAccResourceIndexTemplateDataStreamOptions' ./internal/elasticsearch/index/...` — both pass (~5s each)
- [x] `make docs-generate` — output committed

## Related

- Issue: #2962 (this PR)
- Index template equivalent: #2509 (closed #2201)
- Upstream ES feature: elastic/elasticsearch#117357 (8.18.0 / 9.0.0)


internal/elasticsearch/index/componenttemplate/schema.go
```diff
@@ -49,15 +50,16 @@ func aliasAttrTypes() map[string]attr.Type {
...
+		"mappings":            esindex.MappingsType{},
+		"settings":            customtypes.IndexSettingsType{},
+		"data_stream_options": types.ObjectType{AttrTypes: datastreamoptions.AttrTypes()},
 	}
 }
 
 // getSchema returns the Plugin Framework schema for elasticstack_elasticsearch_component_template.
 // The elasticsearch_connection block is NOT included here; the envelope injects it.
-const schemaVersion int64 = 1
+const schemaVersion int64 = 2
```
tobio:
```suggestion
const schemaVersion int64 = 1
```

We don't need to bump schema versions here, the framework will add the new field on the next apply.

internal/elasticsearch/index/datastreamoptions/validate.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// ValidateRequiresFailureStore enforces the rule that when template.data_stream_options
+// is set, the failure_store block must also be set. Shared between index templates and
+// component templates: both use the same template.data_stream_options.failure_store path.
+func ValidateRequiresFailureStore(ctx context.Context, config tfsdk.Config) diag.Diagnostics {
```
tobio: We've split data stream options out into a shared schema, but the code still expects specific upstream structure (`template.data_stream_options...`). 

This validation should be defined on the schema directly, in reality it looks like we can just have `objectvalidator.AlsoRequires(path.MatchRelative().AtName("failure_store"))` on the `data_stream_options` block. 

Expand/Flatten should also likely take the `data_stream_options` itself rather than the parent, but thats less clear cut.

internal/elasticsearch/index/componenttemplate/schema.go
ottramst: Reverted to `schemaVersion = 1` and dropped the v1→v2 upgrader and its tests. The existing v0→v1 tests cover decoding v0 state against the v1 schema with the new optional block (via `requireUpgradedStateDecodes`), so we still get a regression gate for free. I kept the `data_stream_options: nil` line in `ensureTemplateObjectKeysForV1` alongside the existing alias/mappings/settings ones — same defensive pattern the upgrader already uses for the other optional template keys, doesn't hurt to be explicit.

internal/elasticsearch/index/datastreamoptions/validate.go
ottramst: Switched to `objectvalidator.AlsoRequires(path.MatchRelative().AtName("failure_store"))` on the block itself — same pattern used in `ilm/schema.go` for the frozen phase. That let me delete `datastreamoptions/validate.go`, both `ValidateConfig` methods (in `componenttemplate/` and `template/`), and the now-unused error message constants. The shared package no longer leaks parent-path assumptions.

On Expand/Flatten: I think they already do what you're describing — `Expand(obj types.Object)` takes the `data_stream_options` object directly (see `datastreamoptions/expand.go:30`) and `Flatten` returns one. Callers in both packages pass the child object, not the template parent. Let me know if you were seeing something different.

tobio (approved):
LGTM, thanks for closing this out

PR #2968, opened 2026-05-14
fix(component_template): use MappingsType to normalize implicit type:object

### LINKED ISSUE #2967: [Bug] component_template: inconsistent result after apply with nested object mappings
**Describe the bug**

`elasticstack_elasticsearch_component_template` produces a "Provider produced inconsistent result after apply" error on first apply when `template.mappings` contains nested object mappings without an explicit `"type": "object"`.

Elasticsearch implicitly returns `"type": "object"` for any property that has sub-`properties`, but the component template resource does not normalize this on read. The planned value omits it, the state after create includes it, and Terraform detects the mismatch.

The `index` and `index_template` resources handle this correctly via `MappingsType`/`MappingsValue`, which strips the implicit `"type": "object"` during normalization. The component template resource was not wired up to use this type when it was migrated to Plugin Framework.

**To Reproduce**

1. Apply the following configuration:

```hcl
resource "elasticstack_elasticsearch_component_template" "test" {
  name = "test-mapping-bug"

  template {
    mappings = jsonencode({
      properties = {
        event = {
          properties = {
            size = { type = "long" }
          }
        }
      }
    })
  }
}
```

2. Run `terraform apply`
3. Observe error:

```

.template.mappings: was
cty.StringVal("{\"properties\":{\"event\":{\"properties\":{\"size\":{\"type\":\"long\"}}}}}"),
but now
cty.StringVal("{\"properties\":{\"event\":{\"properties\":{\"size\":{\"type\":\"long\"}},\"type\":\"object\"}}}").
```

4. Run `terraform apply` again. It succeeds because the state now includes `"type": "object"`.

**Expected behavior**

`terraform apply` should succeed on the first run. The provider should normalize the API response to strip implicit `"type": "object"`, matching the behavior of the `index` and `index_template` resources.

- OS: Linux
- Terraform Version: 1.14.3
- Provider version: 0.15.0
- Elasticsearch Version: 9.2.6

**Additional context**

The fix is to use `esindex.MappingsType{}` instead of `jsontypes.NormalizedType{}` in the component template schema, model, and flatten logic. This is the same custom type already used by `index` and `index_template`.

Workaround: retry `terraform apply` (second run succeeds), or explicitly include `"type": "object"` in all nested mapping definitions.

DanielChanJA:
## Changelog
Customer impact: fix
Summary: Fix "Provider produced inconsistent result after apply" for component templates with nested object mappings.

## Detailed changes

The component template resource used `jsontypes.NormalizedType` for its mappings field. This type does not strip the implicit `"type":"object"` that Elasticsearch injects on read for properties with sub-properties. On first apply, the planned value lacks `"type":"object"` but the state returned after creation includes it, causing Terraform to reject the result.

The `index` and `index_template` resources already use `MappingsType`/`MappingsValue`, which normalizes this on both construction and semantic equality comparison. The component template was not wired up to use this type when it was migrated to Plugin Framework.

This PR switches the component template's mappings field to use `esindex.MappingsType{}` in the schema and `templateAttrTypes()`, `esindex.MappingsValue` in the model, and `esindex.NewMappingsValue()`/`esindex.NewMappingsNull()` in the flatten logic.

Closes #2967

### DIFF
--- a/internal/elasticsearch/index/componenttemplate/acc_test.go
+++ b/internal/elasticsearch/index/componenttemplate/acc_test.go
@@ -53,6 +53,25 @@ func TestAccResourceComponentTemplate(t *testing.T) {
 	})
 }
 
+func TestAccResourceComponentTemplateNestedObjectMappings(t *testing.T) {
+	templateName := sdkacctest.RandStringFromCharSet(10, sdkacctest.CharSetAlphaNum)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { acctest.PreCheck(t) },
+		CheckDestroy: checkResourceComponentTemplateDestroy,
+		Steps: []resource.TestStep{
+			{
+				ProtoV6ProviderFactories: acctest.Providers,
+				ConfigDirectory:          acctest.NamedTestCaseDirectory("create"),
+				ConfigVariables:          config.Variables{"name": config.StringVariable(templateName)},
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("elasticstack_elasticsearch_component_template.test", "name", templateName),
+				),
+			},
+		},
+	})
+}
+
 func TestAccResourceComponentTemplateAliasDetails(t *testing.T) {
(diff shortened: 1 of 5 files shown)


tobio (approved):
👌

PR #3572, opened 2026-06-03
fix(security_exception_item): honour append-only-immutable comments semantics

### LINKED ISSUE #3549: security_exception_item: comments are append-only; provider needs to honor that on both Read (preserve ids) and Plan (require replacement for non-append edits)

mrohland:
## Changelog
Customer impact: fix
Summary: Stop `elasticstack_kibana_security_exception_item` from duplicating comments on every Update, and surface unsupported edits (text edit / remove / reorder) as a clean resource replacement instead of silent drift or HTTP 400.

## Detailed changes
Fixes #3549. Supersedes #3551 (closed).

The Kibana `PUT /api/exception_lists/items` API treats the `comments` field as **append-only and immutable**, verified empirically against a 9.x cluster:

| PUT operation | HTTP | Effect on Kibana storage |
|---|---|---|
| No-op (same ids + same texts) | 200 | no change |
| Append (entry with no `id` at the END) | 200 | new comment added |
| Omit existing entries | 200 | **silently ignored** — all kept |
| Edit text via same id | 200 | **silently ignored** — text preserved |
| Reorder / insert in middle / prepend | 400 | `"item \"comments\" are append only"` |

`created_by` is unconditionally stamped from the authenticated API-key owner; the Create/Update request structs in the OpenAPI spec do not contain a `created_by` field at all. There is no client path to preserve original analyst attribution across a recreate.

### Coordinated fixes

Three changes that only work as a set:

1. `commentsToUpdateAPI` (in `models.go`) now passes `comment.ID` through to the update payload when the planned value is Known.

2. The `comment.id` schema attribute gets a `stringplanmodifier.UseStateForUnknown()` plan modifier. Without this, the framework marks the id Unknown on every plan that touches the parent `comments` list, and (1) has nothing to send.

3. A new custom `commentsAppendOnlyRequireReplacePlanModifier` list-attribute plan modifier on `comments`. It compares state to plan and triggers `resp.RequiresReplace = true` whenever the diff is not pure append-at-end (edit / remove / reorder of existing comments). The accompanying warning makes clear that `created_by` will be re-stamped on every comment of the recreated item.

### Result

| YAML edit | Plan | Apply |
|---|---|---|
| No change | no-op | no API call |
| Append at end | `~ update in-place` | PUT, Kibana appends, original ids preserved |
| Edit existing text | **`-/+ destroy and recreate`** + warning | item destroyed, recreated with new texts |
| Remove a comment | **`-/+ destroy and recreate`** + warning | item destroyed, recreated without that comment |
| Reorder / prepend | **`-/+ destroy and recreate`** + warning | item destroyed, recreated in desired order |

No more silent duplication, no more `produced inconsistent result after apply`, no more opaque 400s. The cost of a recreate (re-stamped `created_by`) is loud and explicit in the plan output.

## Test plan
- [x] 9 unit tests in `comments_plan_modifiers_test.go` cover every branch of the new plan modifier (create, destroy, no-op, append, edit, shrink, prepend, reorder existing, UseStateForUnknown placeholder).
- [x] `TestAccResourceExceptionItem_CommentsAppendOnly` (6 steps) covers the live integration: create, no-op (`plancheck.ResourceActionNoop`), append in-place (`plancheck.ResourceActionUpdate` + id preservation), edit-replace, shrink-replace, prepend-replace (`plancheck.ResourceActionDestroyBeforeCreate` for the last three).
- [ ] Reviewer to run the acceptance test against a live Kibana cluster.
- [x] Existing `TestAccResourceExceptionItem_Comments` still compiles (`go test -short` passes).


tobio:
Given the behaviour in Kibana, I'm a little skeptical of the `comments` attribute as it stands. It feels relatively impractical to force a replacement ~to simply add a comment~ because comments have been added out of band. 

WDYT about adding a sibling attribute which is instead a subset (ordered subset?) of the existing comments? I'm not asking you to implement that right now, and I think this PR stands on it's own, but you're obviously using this resource so have an actual opinion.

PR #3749, opened 2026-06-13
fix(kibana): normalize .slack_api connector config to avoid inconsistent apply

### LINKED ISSUE #3748: [Bug] `.slack_api` action connector fails with "Provider produced inconsistent result after apply" when allowedChannels omit id
**Describe the bug**
Creating an elasticstack_kibana_action_connector of type .slack_api whose config lists allowedChannels with only name fields fails after the Kibana API call succeeds with a "Provider produced inconsistent result after apply" error. Kibana echoes the channels back with an empty id for each channel, and the provider also stamps its internal __tf_provider_context key onto the value read back into state. The planned config value and the post-apply read value therefore differ, and the provider does not normalize them to be equal for the .slack_api type.

**To Reproduce**
1. TF configuration used 
```hcl
resource "elasticstack_kibana_action_connector" "slack-api-connector" {
  name              = "slack-api-connector"
  connector_type_id = ".slack_api"
  config = jsonencode({
    allowedChannels = [
      { name = "#testing" },
      { name = "#test-alerts" },
    ]
  })
  secrets = jsonencode({
    token = "xxx-..."
  })
}
```
2. TF operations to execute to get the error : `terraform apply`
3. See the error in the output:
```

When applying changes to elasticstack_kibana_action_connector.slack-api-connector,
provider "registry.terraform.io/elastic/elasticstack" produced an unexpected new value:
.config: was
  cty.StringVal("{\"allowedChannels\":[{\"name\":\"#testing\"},{\"name\":\"#test-alerts\"}]}"),
but now
  cty.StringVal("{\"__tf_provider_context\":\".slack_api\",\"allowedChannels\":[{\"id\":\"\",\"name\":\"#testing\"},{\"id\":\"\",\"name\":\"#test-alerts\"}]}").
```

**Expected behavior**
The connector is created and a subsequent terraform plan reports no changes. The provider-internal __tf_provider_context key and the Kibana-injected empty id fields are normalized away so the planned and applied config values compare as equal.

**Actual Behavior**
Apply errors out with "Provider produced inconsistent result after apply" because the planned config (channels with only name) does not match the value read back after apply (channels with an extra empty id, plus the internal __tf_provider_context key).

 - OS: MacOs
 - Terraform Version 1.6.6
 - Provider version 0.16.0
 - Elasticsearch Version 9.4.2

**Root Cause**
- `kbapi.SlackApiConfig.AllowedChannels[].Id` is serialized as `json:"id"`
  *without* `omitempty`. When the provider remarshals the Kibana read response
  through this struct (`ConnectorResponseToModel` →
  `connectorConfigHandlers[".slack_api"].remarshalConfig`), each channel that
  the user supplied with only a `name` gains an empty `"id":""`.
- The `config` attribute uses a custom `ConfigType`/`ConfigValue`
  (`customtypes.JSONWithContextualDefaultsValue`) whose `StringSemanticEquals`
  strips `__tf_provider_context` and then runs both the planned and the applied
  value through `kibanaoapi.ConnectorConfigWithDefaults(connectorTypeID, ...)`
  before comparing.
- `ConnectorConfigWithDefaults` only invokes a connector type's `defaults`
  function. `.slack_api` had **no** `defaults` function (only `remarshalConfig`,
  which is used at read time), so the planned value was returned unchanged and
  never gained the empty `id` fields. The applied value (already remarshaled at
  read) kept its `id` fields, so semantic equality returned false and Terraform
  reported an inconsistent result.
- Connector types that already have a `defaults` function do not hit this
  because their `defaults` function remarshals through the typed struct, so both
  sides of the comparison are normalized identically.

**Proposed Fix:**
Wire the typed remarshal in as the `.slack_api` `defaults` function in
`internal/clients/kibanaoapi/connector_defaults.go`:

```go
".slack_api": {
    defaults:        remarshalConfig[kbapi.SlackApiConfig],
    remarshalConfig: remarshalConfig[kbapi.SlackApiConfig],
},
```

This makes `ConnectorConfigWithDefaults(".slack_api", planned)` normalize the
planned value the same way the read path normalizes the applied value, so both
carry the empty `id` fields during semantic equality and compare as equal. The
fix is scoped to `.slack_api` and does not change any other connector type.

aydasraf:
Fixes #3748

## What

Creating an `elasticstack_kibana_action_connector` of type `.slack_api` whose `config` lists `allowedChannels` with only `name` fields fails after the Kibana API call succeeds:

```

.config: was
  cty.StringVal("{\"allowedChannels\":[{\"name\":\"#kar_testing\"},{\"name\":\"#test-prod-alerts\"}]}"),
but now
  cty.StringVal("{\"__tf_provider_context\":\".slack_api\",\"allowedChannels\":[{\"id\":\"\",\"name\":\"#kar_testing\"},{\"id\":\"\",\"name\":\"#test-prod-alerts\"}]}").
```

## Root cause

- `kbapi.SlackApiConfig.AllowedChannels[].Id` is serialized as `json:"id"` **without** `omitempty`. When the provider remarshals the Kibana read response through this struct (`ConnectorResponseToModel` → `connectorConfigHandlers[".slack_api"].remarshalConfig`), each channel the user supplied with only a `name` gains an empty `"id":""`.
- The `config` attribute's `StringSemanticEquals` strips `__tf_provider_context` and then runs both the planned and the applied value through `kibanaoapi.ConnectorConfigWithDefaults(connectorTypeID, ...)`, which only invokes a connector type's `defaults` function.
- `.slack_api` had **no** `defaults` function (only `remarshalConfig`, used at read time), so the planned value was never normalized the same way as the read-back value. The two diverged (`id:""` only on the applied side) and Terraform reported an inconsistent result.
- Connector types that already have a `defaults` function don't hit this, because their `defaults` remarshals through the typed struct so both sides normalize identically.

## Fix

Wire the typed remarshal in as the `.slack_api` `defaults` function in `internal/clients/kibanaoapi/connector_defaults.go`, so the planned value is normalized identically to the read response. Scoped to `.slack_api`; no other connector type is affected.

## Tests

- Unit regression: `TestConfigValue_StringSemanticEquals_SlackAPI` (fails without the fix, passes with it).
- Acceptance regression: `TestAccResourceKibanaConnectorSlackAPI` — creates the connector and asserts a subsequent plan is empty and `__tf_provider_context` does not leak into state. Verified green against a local Kibana 9.4 stack.

## Changelog
Customer impact: fix
Summary: Fix "Provider produced inconsistent result after apply" for `.slack_api` action connectors by normalizing the planned `config` value the same way the read response is normalized.


kavealimohammadi:
I am encountering the same issue when provisioning a webhook connector configured with OAuth2 authentication. The connector is created successfully and appears to be functional, but terraform apply still returns an error. As a result, there seems to be a discrepancy between the actual resource state and the response reported during the apply process.

Is there any update on this issue?

kavealimohammadi:
Will this change also affect other connector types that might be experiencing a similar error?
We have an issue by webhook type connectors

aydasraf:
> Will this change also affect other connector types that might be experiencing a similar error? We have an issue by webhook type connectors

Hello @kavealimohammadi, I don't work with Elastic; I'm only contributing a fix for my specific case. This fix only addresses the Slack API issue, as I have personally investigated and tested it.

tobio (approved):

PR #4073, opened 2026-07-03
fix(fleet): fall back to Uninstall when Fleet rejects DeleteKibanaAssets in install space

### LINKED ISSUE #4045: [9.5.0-SNAPSHOT] Fleet integration destroy: DeleteKibanaAssets 400 on install space requires package uninstall fallback
## Summary

`TestAccResourceIntegrationSkipDestroy` fails on the 9.5.0-SNAPSHOT matrix during post-test destroy. Fleet 9.5 rejects deleting kibana assets from the space where a package was installed, requiring a full package uninstall instead.

## Failure

```
acc_test.go:445: Error running post-test destroy, there may be dangling resources: exit status 1
Error: Unexpected status code from server: got HTTP 400
{"statusCode":400,"error":"Bad Request","message":"Impossible to delete
kibana assets from the space where the package was installed, you must
uninstall the package."}
```

Reproduced locally (the 400 was reproduced directly against the Fleet API; the full acceptance test passes locally only because the local stack doesn't trigger the multi-space code path that CI does).

## Root cause

Fleet 9.5 changed package asset lifecycle: deleting kibana assets from the space where a package was installed is rejected unless the package itself is uninstalled. The provider's destroy path for `elasticstack_fleet_integration` (`internal/fleet/integration/delete.go`) attempts to remove those assets via `fleet.DeleteKibanaAssets` when `isInstalledInMultipleSpaces` is true, hitting the 400.

### Validated locally

- `DELETE /api/fleet/epm/packages/tcp/1.16.0/kibana_assets?force=true` on the install (default) space → **HTTP 400** with the exact CI message: *"Impossible to delete kibana assets from the space where the package was installed, you must uninstall the package."*
- `DELETE /api/fleet/epm/packages/tcp/1.16.0?force=true` (full uninstall) → **HTTP 200** and clears the assets.

So falling back to `Uninstall` when `DeleteKibanaAssets` returns this 400 is sound.

### Code path

`internal/fleet/integration/delete.go` calls `fleet.DeleteKibanaAssets` (in `internal/clients/fleet/packages.go:152`) when `isInstalledInMultipleSpaces(pkg, spaceID)` is true. In CI, the global-Fleet-asset inflation (visible in CI docker logs: "Global Fleet assets changed … Reinstalling 2 packages") makes this branch trigger for the primary install space, where Fleet now rejects the asset-only delete.

`fleet.Uninstall` (`internal/clients/fleet/packages.go:98`) already exists and routes to `DELETE /api/fleet/epm/packages/{name}/{version}`.

## Classification

**Functional Stack change** (package-asset deletion semantics). The provider's destroy path is a **latent issue** in that it doesn't route through package uninstall when assets are install-space-managed and Fleet rejects the asset-only delete.

## Options considered

- **F1 (provider, route to uninstall):** On destroy, when Fleet returns this 400 from `DeleteKibanaAssets`, fall back to / call the package `Uninstall` endpoint for the affected package (which clears the install-space assets). Aligns with the new Fleet 9.5 contract. **Validated** via direct API testing (400 → uninstall → 200).
- **F2 (test skip):** Version-gate `TestAccResourceIntegrationSkipDestroy` to skip on `>=9.5.0-SNAPSHOT` until the provider destroy path is fixed. Unblocks CI but defers the real bug.
- **F3 (provider, suppress the 400):** Suppress/ignore this specific 400 during destroy (treat assets-already-managed as success). Avoids the dangling-resource error but leaves the package installed — conflicts with the test's intent and leaves real dangling resources.

## Selected option

**F1** — on destroy, when Fleet returns this 400 from `DeleteKibanaAssets`, fall back to `Uninstall` for the affected package.

## Proposed change

In `internal/fleet/integration/delete.go` (or `internal/clients/fleet/packages.go` `DeleteKibanaAssets`), detect the "Impossible to delete kibana assets from the space where the package was installed" 400 response and fall back to `fleet.Uninstall(ctx, fleetClient, name, version, spaceID, force)`, which clears the install-space assets.

## Reproduction

Direct API repro of the 400 and the 200 fallback:
```bash
EP=http://localhost:15701
# install tcp
curl -s -u elastic:password -X POST "$EP/api/fleet/epm/packages/tcp/1.16.0" -H 'kbn-xsrf: true' -H 'Content-Type: application/json' -d '{"force":true}'
# DeleteKibanaAssets on install space -> 400
curl -s -o - -w "\nHTTP %{http_code}\n" -u elastic:password -X DELETE "$EP/api/fleet/epm/packages/tcp/1.16.0/kibana_assets?force=true" -H 'kbn-xsrf: true'
# Uninstall (full) -> 200
curl -s -o - -w "\nHTTP %{http_code}\n" -u elastic:password -X DELETE "$EP/api/fleet/epm/packages/tcp/1.16.0?force=true" -H 'kbn-xsrf: true'
```

Acceptance test (passes locally only when the multi-space branch isn't hit; CI hits it):
```bash
export ELASTICSEARCH_ENDPOINTS=http://localhost:19200
export ELASTICSEARCH_USERNAME=elastic ELASTICSEARCH_PASSWORD=password
export KIBANA_ENDPOINT=http://localhost:15701
export KIBANA_USERNAME=elastic KIBANA_PASSWORD=password
export TF_ELASTICSTACK_INCLUDE_EXPERIMENTAL=true TF_ACC=1
go test -v -run '^TestAccResourceIntegrationSkipDestroy$' ./internal/fleet/integration
```

tobio:
## Summary

Fixes https://github.com/elastic/terraform-provider-elasticstack/issues/4045

Fleet 9.5 changed the semantics of `DELETE /api/fleet/epm/packages/{name}/{version}/kibana_assets`: it now returns HTTP 400 (`"Impossible to delete kibana assets from the space where the package was installed, you must uninstall the package."`) when called from the package's install space. The `elasticstack_fleet_integration` destroy path called this endpoint whenever the package was installed in multiple spaces, surfacing the 400 as a Terraform error and breaking `TestAccResourceIntegrationSkipDestroy` on the Fleet 9.5.0-SNAPSHOT CI matrix.

- When `fleet.DeleteKibanaAssets` returns this specific 400 (matched by a stable, whitespace-normalized message substring, restricted to error-severity diagnostics), the resource now falls back to `fleet.Uninstall` (full package uninstall) instead of surfacing the 400.
- A `DEBUG` log entry is emitted before the fallback with the package name, version, and space ID for auditability.
- Any other error from `DeleteKibanaAssets` is still surfaced unchanged, and `Uninstall` is never called on the success path.
- No schema changes; all other destroy paths (`skip_destroy`, non-space-aware, single-space) are unchanged.

## Changelog
Customer impact: fix
Summary: Fall back to full package uninstall when Fleet 9.5 rejects DeleteKibanaAssets in the install space during integration destroy.

## Test plan

- [x] Unit tests in `internal/fleet/integration/delete_test.go` cover: install-space 400 → fallback to `Uninstall` (with `force` propagation and no leakage of the original 400 into diagnostics), a different 400 → surfaced as error with no fallback, HTTP 200 → no fallback, and a fallback where `Uninstall` itself fails (only the `Uninstall` error is surfaced).
- [x] `make build`
- [x] `make check-lint`
- [x] `make check-openspec`
- [x] `go test ./internal/fleet/integration/... ./internal/clients/fleet/...`
- Acceptance test `TestAccResourceIntegrationSkipDestroy` requires a Fleet 9.5.0-SNAPSHOT stack to reproduce the original failure; this worktree's local stack runs 9.4.0, so this is validated by CI on the 9.5.0-SNAPSHOT matrix.


tobio:
## Critical fix: gated the fallback on `force`

A reviewer caught a serious issue in the original approach (commit `e3481ae`, approved by verify-openspec at `64c7319`): `fleet.Uninstall` removes a package **globally**, from every space it's installed in — not just the target space. Confirmed via Fleet/Kibana engineering discussion ([elastic/kibana#186620](https://github.com/elastic/kibana/pull/186620), [#172963](https://github.com/elastic/kibana/issues/172963)): *"uninstall will uninstall the package and his assets in all space."*

This meant the original fallback had a data-loss bug: if a package is managed across multiple spaces via separate `elasticstack_fleet_integration` resources (e.g. `space_id = "A"` and `space_id = "B"`), destroying the resource for whichever space happens to be the package's **install space** would silently uninstall the package from every other space too — including spaces still declared and managed by other Terraform resources.

**Fix** (commit `b09b5e5`): gate the fallback on the existing `force` attribute (no schema change):
- `force = true` → proceed with the fallback exactly as before (global `Uninstall`, DEBUG log first). This preserves the original CI fix, since `TestAccResourceIntegrationSkipDestroy`'s fixtures already set `force = true`.
- `force = false` (default) → do **not** call `Uninstall`; instead surface a distinct, actionable error explaining the situation and directing the caller to destroy the other space's resource(s) first or set `force = true`.

Also un-archived the change (the earlier verify-openspec run had approved and archived it before this was found) and updated `proposal.md`/`design.md`/the delta spec/`tasks.md` to document the corrected design (see Decision 4 in `design.md`).

Re-requesting `verify-openspec` review on the corrected design.

PR #4160, opened 2026-07-15
fix: skip data stream lifecycle delete on Elastic serverless

### LINKED ISSUE #1289: [Bug] Data stream lifecycles on Elastic Cloud serverless can't be deleted
**Describe the bug**
Deleting the resource `resource elasticstack_elasticsearch_data_stream_lifecycle` is not possible with the Terraform provider. Nonethless, it is possible in the Kibana UI that uses the following API endpoint: `api/index_management/delete_data_streams`.

**To Reproduce**

```terraform
resource "elasticstack_elasticsearch_index_template" "main" {
  name     = "test"
  priority = 500

  index_patterns = ["test*"]

  data_stream {}

  template { }
}

resource "elasticstack_elasticsearch_data_stream" "main" {
  name = "test"

  depends_on = [
    elasticstack_elasticsearch_index_template.main
  ]
}

resource "elasticstack_elasticsearch_data_stream_lifecycle" "main" {
  name               = "test"
  data_retention = null

  depends_on = [
    elasticstack_elasticsearch_data_stream.main,
  ]
```

The error:

```
Failed with: {"error":{"root_cause":[{"type":"api_not_available_exception","reason":"Request for uri
│ [/_data_stream/test/_lifecycle?expand_wildcards=open] with method [DELETE] exists but is not available when running
│ in serverless mode"}],"type":"api_not_available_exception","reason":"Request for uri
│ [/_data_stream/test/_lifecycle?expand_wildcards=open] with method [DELETE] exists but is not available when running
│ in serverless mode"},"status":410}
````

**Expected behavior**
The resource `elasticstack_elasticsearch_data_stream_lifecycle` is correctly deleted. In the same manner that this can be done in the Kibana UI (`api/index_management/delete_data_streams`).

If this is not possible due to API limitations, it should be documented that use of this resource requires a manual removal from the Terraform state when destroying the infrastructure (`tf state rmelasticstack_elasticsearch_data_stream_lifecycle.main`)?

mvanhorn:
## Summary

`terraform destroy` on an `elasticstack_elasticsearch_data_stream_lifecycle` resource now completes on Elastic Cloud Serverless instead of getting stuck.

## Why this matters

Deleting the resource on serverless calls `DELETE /_data_stream/{name}/_lifecycle`, which Elasticsearch answers with HTTP 410 `api_not_available_exception` ("not available when running in serverless mode"). The error is uncatchable, so the destroy never finishes and the resource is wedged in state. @tobio confirmed in #1289 that this is a real bug, and clarified that removing the lifecycle is distinct from deleting the data stream's data.

On serverless, data stream lifecycle and retention are managed by Elastic, so there is nothing for the provider to delete.

## Changes

- Guard `DeleteDataStreamLifecycle` (`internal/clients/elasticsearch/datastream.go`) with an `IsServerless` check before issuing the DELETE, mirroring the existing pattern in the index create path. On serverless it skips the request and returns a warning diagnostic explaining that Elastic manages retention automatically, so Terraform removes the resource from state cleanly. Stateful clusters keep the existing behavior, including the 404-on-missing-lifecycle handling.

## Testing

`go test ./internal/clients/elasticsearch/`. Added unit tests (`datastream_test.go`) using an httptest mock server, so no live Elasticsearch is needed:

- serverless (`build_flavor: serverless`): the DELETE is not sent, no error, one warning diagnostic,
- stateful (`build_flavor: default`): the DELETE is sent and succeeds with no diagnostics,
- a failing serverless check (info endpoint 500) surfaces the error and does not send the DELETE.

Fixes #1289

## Changelog
Customer impact: fix
Summary: Skip the data stream lifecycle DELETE on Elastic Cloud Serverless so destroying `elasticstack_elasticsearch_data_stream_lifecycle` completes instead of failing on the HTTP 410 `api_not_available_exception`.


tobio (approved):
LGTM, ty!

mvanhorn:
Thanks for the ty and merge, @tobio - serverless targets skip the lifecycle delete cleanly now.

PR #4352, opened 2026-07-30
Fix: allow manage_own_api_key to delete elasticstack_elasticsearch_security_api_key

### LINKED ISSUE #4351: [Bug] elasticstack_elasticsearch_security_api_key delete requires manage_api_key even when the key is owned by the calling user
**Describe the bug**
Destroying (or replacing) an `elasticstack_elasticsearch_security_api_key` resource fails with a 403 unless the Elasticsearch user configured on `elasticsearch_connection` has the broad `manage_api_key` (or `manage_security`) cluster privilege. The `manage_own_api_key` privilege, which is documented as sufficient to create/manage API keys owned by the calling user, is not accepted.

This happens because `DeleteAPIKey` (`internal/clients/elasticsearch/security_api_key.go`) always calls the Invalidate API Key endpoint with an explicit `ids` list and never sets `"owner": true`:

```go
_, err := typedClient.Security.InvalidateApiKey().Request(&invalidateapikey.Request{
    Ids: []string{id},
}).Do(ctx)
```

Per the [Invalidate API Key API docs](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-security-invalidate-api-key):

> To use this API, you must have at least the `manage_security`, `manage_api_key`, or `manage_own_api_key` cluster privileges. The `manage_security` privilege allows deleting any API key, including both REST and cross cluster API keys. The `manage_api_key` privilege allows deleting any REST API key, but not cross cluster API keys. The `manage_own_api_key` only allows deleting REST API keys that are owned by the user. In addition, with the `manage_own_api_key` privilege, an invalidation request must be issued in one of the three formats:
> 1. Set the parameter `owner=true`.
> 2. Or, set both `username` and `realm_name` to match the user's identity.
> 3. Or, if the request is issued by an API key, that is to say an API key invalidates itself, specify its ID in the `ids` field.

Because the provider always sends `ids` alone without `owner: true` (and the calling credential is normally a user/service account, not an API key invalidating itself), none of the three `manage_own_api_key` request formats are satisfied, so Elasticsearch falls back to requiring `manage_api_key`.

**To Reproduce**
1. TF configuration used: an `elasticstack_elasticsearch_security_api_key` resource, with `elasticsearch_connection` pointed at a user/service account whose role only grants `manage_own_api_key` (plus whatever cluster privileges are needed to create the key, e.g. `monitor`).
2. TF operations to execute to get the error: `terraform apply` followed by `terraform destroy` (or any change that requires replacing the resource).
3. See the error in the output:
```
Error: status: 403, failed: [security_exception], reason: action
[cluster:admin/xpack/security/api_key/invalidate] is unauthorized for user
[build_queue_writer] with effective roles
[build_queue_writer,manage_own_api_keys,monitor_cluster], this action is
granted by the cluster privileges [manage_api_key,manage_security,all]
```

**Expected behavior**
When the resource's API key is invalidated, the provider should be able to authorize the request with `manage_own_api_key` alone (the same privilege that already covers create), by sending the invalidate request with `owner: true` (request format 1 above) rather than only `ids`.

 - Provider: `elasticstack_elasticsearch_security_api_key` resource
 - Elasticsearch Version: any version supporting the Invalidate API Key API's `owner` flag (added in 6.7.0)

**Additional context**
Affected code: `internal/clients/elasticsearch/security_api_key.go` (`DeleteAPIKey`), called from `internal/elasticsearch/security/apikey/resource/delete.go` and `internal/elasticsearch/security/apikey/ephemeral/resource.go`.

I'll open a PR that adds an optional `owner` attribute (default `true`) to the resource schema and threads it through to the Invalidate API Key request's `owner` flag.

aklakina:
Fixes #4351

## Changelog
Customer impact: fix
Summary: Add an optional `owner` attribute (default `true`) to `elasticstack_elasticsearch_security_api_key` so deleting/invalidating the key can be authorized with the `manage_own_api_key` cluster privilege instead of always requiring `manage_api_key`.

## Detailed changes

`DeleteAPIKey` (`internal/clients/elasticsearch/security_api_key.go`) always invalidated API keys by sending an id-scoped Invalidate API Key request (`ids: [...]`) without setting `owner: true`.

Per the [Invalidate API Key API docs](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-security-invalidate-api-key), the `manage_own_api_key` cluster privilege only authorizes an invalidate request when it is issued in one of three formats:
1. `owner=true` is set.
2. Both `username` and `realm_name` are set and match the caller's identity.
3. The request is issued by an API key invalidating itself, identified via `ids`.

The provider always sent `ids` alone (format 3 requires the *caller* to be authenticating as that same API key, which isn't the case for a Terraform-managed connection), so none of the three formats were satisfied and Elasticsearch required the broader `manage_api_key`/`manage_security` privilege instead, even for a key actually owned by the calling user.

Changes:
- `DeleteAPIKey` now accepts an `owner bool` parameter and sets it on the `invalidateapikey.Request` (satisfying format 1 above, which is compatible with also specifying `ids`).
- Added a new `owner` attribute (`Optional`, `Computed`, default `true`) to the `elasticstack_elasticsearch_security_api_key` resource schema, threaded through to `DeleteAPIKey` via a new `TfModel.OwnerValue()` helper (defaults to `true` when null/unknown).
- The ephemeral `security_api_key` resource does not expose an `owner` attribute; its `Close` callback always invalidates with `owner: true` (an ephemeral key is always owned by the invoking connection).
- Updated existing unit tests (`ephemeral/resource_test.go`, `ephemeral/acc_test.go`) for the new `DeleteAPIKey` signature, and added `TestTfModelOwnerValue` covering the null/unknown/true/false cases.
- Updated `docs/resources/elasticsearch_security_api_key.md` with the new `owner` attribute description.

### Limitations / follow-up
- `owner: false` (i.e. deleting keys not owned by the caller) still requires `manage_api_key`/`manage_security`; this PR doesn't change that, it only makes the common "delete my own key" case work with the narrower `manage_own_api_key` privilege.
- No behavior change for users already granting `manage_api_key`/`manage_security`; the new attribute defaults to `true` and simply adds the `owner` flag to the existing `ids`-based invalidate request.


internal/elasticsearch/security/apikey/models.go
```diff
@@ -86,6 +87,16 @@ func (model TfModel) GetElasticsearchConnection() types.List {
...
+// OwnerValue returns the effective value of the `owner` attribute, defaulting
+// to true (the safe default that only requires `manage_own_api_key`) when the
+// value is null or unknown.
+func (model TfModel) OwnerValue() bool {
+	if !typeutils.IsKnown(model.Owner) || model.Owner.IsNull() {
+		return true
+	}
+	return model.Owner.ValueBool()
+}
+
```
tobio:
```suggestion
```

This is just duplicating the schema default.

internal/elasticsearch/security/apikey/resource/schema.go
```diff
@@ -130,6 +132,15 @@ func getSchema(version int64) schema.Schema {
 					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
+			attrOwner: schema.BoolAttribute{
+				Description: apikey.OwnerDescription,
+				Optional:    true,
+				Computed:    true,
+				Default:     booldefault.StaticBool(true),
```
tobio:
```suggestion
				Default:     booldefault.StaticBool(false),
```

As it is, this is a breaking change since `owner=true` narrows the scope of the invalidate request to only those keys actually owned by the current user. Keys owned by a different user a ignored, and left active whilst being removed from Terraform state. 

From the example in the API docs:

```
// Run `DELETE /_security/api_key` to invalidate the API keys identified by ID // if they are owned by the currently authenticated user.
{
  "ids" : ["VuaCfGcBCdbkQm-e5aOx"],
  "owner" : "true"
}
```

Is this a 'good' default in the greenfield scenario? Absolutely. Unfortunately it introduces a silent failure. 

To move ahead here, I think we'd want to:
1. Switch the default to `false`. It's the only option for a non-breaking change.
2. Validate the response from the DELETE call, i.e ensure the `id` is in either `invalidated_api_keys` or `previously_invalidated_api_keys`. This is an existing issue, but exposed by this change since it wasn't possible to filter the keys previously. 
3. Most likely thread `owner` through to the read path. An API key owned by a different user would be treated as non-existent which more closely matches the schema intent.

aklakina: But there can not be a state where the api key is created for a different user as per right now the provider does not have this functionality. Correct me if I'm wrong, but without this functionality the default true is a better approach and will allow later for this functionality to appear.

aklakina: Please resolve this comment if you find my answer correct.

tobio: > But there can not be a state where the api key is created for a different user as per right now 

This isn't true unfortunately. There's nothing preventing a team from creating an API key under one user, then deleting the key with another user. As long as everyone has the right access, it all works.

aklakina: Hmm. To me this is a very weird edge case and maybe even a bug in the resource and provider configuration but I can't really reason it. It just does not feel right that one can do this. Anyway, I see your point, I'll set the default to false.

tobio:
@aklakina sorry for the delay on this one. I haven't gotten a chance to focus on it over the last week. 

> Hmm. To me this is a very weird edge case and maybe even a bug in the resource and provider configuration but I can't really reason it. It just does not feel right that one can do this. Anyway, I see your point, I'll set the default to false.

You're right here, it's a pretty niche case, and avoiding a breaking change here (`owner=true` by default) creates a worse UX for many other provider users. 

In practice, the mental model is that the provider owns the keys, and when the user applies a change which would destroy a key they expect it to be destroyed. Taking a step back from this PR, I think the best approach might be to just embed that behaviour into the provider itself, essentially attempt to invalidate the key with `owner=true`, if that fails then try with `owner=false`, and only report an error if that request also fails. I've been meaning to verify that pattern works and look at moving towards that approach. No need for another attribute, things Just Work™️.

This PR is great, and I don't expect you to keep up with us shifting the goal posts a couple of times. I'm planning on verifying the above plan this week, but if you wanted to take a look down that path then that would also be really well received. 

So to be clear, this hasn't just been dropped on the floor, I'm keen to get this behaviour fixed but your feedback is on-point and we don't want to introduce a new attribute unnecessarily, especially when it's less good for many users.

aklakina:
@tobio 

While I agree that the things just work mentality is important and general users don't need to know what the owner attribute is, advanced and niche cases could cause confusion when the same operation needs different permission because of underlying logic. Are you sure you don't want to open this up? Of course the error messages could be made clear in these cases, which also somewhat solves this issue.

As for you not expecting to keep up with moving the goal post, I'm completely fine with working more on this.

tobio:
> advanced and niche cases could cause confusion when the same operation needs different permission because of underlying logic. Are you sure you don't want to open this up?

Good question, that's essentially what I've been wanting to look into. AFAICT the argument for retaining an `owner` style attribute is to take full control over the provider behaviour (e.g restrict the provider to only reading/invalidating owned API keys). 

That's valid, but it's a different use-case from the reported bug. IMO there's two paths here:

1. The bug fix
What I described above, we attempt to invalidate with both `owner=true` and `owner=false` and just make things work. 

2. The bug fix, and control
Behaviour from 1, with an _optional_ `restrict_to_owned` attribute (not tied to that name, but more IMO something more explicit that just `owner`). That new attribute limits read/delete to `owner=true` and so avoids accidentally deleting a shared key in the case of multiple privileged users applying the same TF stack.  

I'd err towards just the bug fix purely for the sake of simplicity, the shared key use case is quite niche. I'm happy to accept a PR implementing 2 though. WDYT? Am I missing some other benefit to exposing this?

aklakina:
I've dug into the code a bit more about this and this bug, where the `owner=true` blocks the delete, appears only when someone changes the user in the global provider configuration and the resource itself does not have `elastic_connection` block defined. If `elastic_connection` is defined, then the delete function uses the credentials from the state, not the configuration, as far as I understood the code.

I would say that in such case, the `owner=true` flag is also necessary in the requests for read because otherwise the resource itself has no idea that the api token being created shall belong to a different user. This ofc leaves the original api token alive, which could be a reason why not to do this.

For me this suggests that instead of `restrict_to_owned` should be `allow_unowned` or something similar and when the refresh sees that the api key is gone in the response, the error message shall clearly state that if the provider configuration has changed, then set this to true for first run.

What do you think about this approach?

Edit: typo

tobio:
> I would say that in such case, the owner=true flag is also necessary in the requests for read because otherwise the resource itself has no idea that the api token being created shall belong to a different user. This ofc leaves the original api token alive, which could be a reason why not to do this.

+1 to consistently setting the `owner` flag on read/delete. 

 > For me this suggests that instead of restrict_to_owned should be allow_unowned or something similar and when the refresh sees that the api key is gone in the response, the error message shall clearly state that if the provider configuration has changed, then set this to true for first run.
 
I don't follow why we'd need to invert the existing behaviour here. `allow_unowned` either defaults to true (and so adds noise to the plan) or introduces a breaking change in provider behaviour. What's the problem being solved with `allow_unowned`?

aklakina:
So then this part:

> I've dug into the code a bit more about this and this bug, where the `owner=true` blocks the delete, appears only when someone changes the user in the global provider configuration and the resource itself does not have `elastic_connection` block defined. If `elastic_connection` is defined, then the delete function uses the credentials from the state, not the configuration, as far as I understood the code.

Is incorrect? If it is not incorrect then this would not be a breaking change. The problem being solved with it is when an api_key resource is created, it always does so in the context of the provider user unless `elastic_connection` overwrites this. This means users expect the key to be created in the name of the provider user. If they change the provider configuration but the resource itself does not filter to owned true, the resource has no context awareness about that change and the api key will be marked as no change required while in reality the configuration changed the user who shall own the api key. This would be solved by wiring in the `owned=true` property to the read path and making it default.

Is there any other case where setting the default to true can cause complications?

tobio:
> This means users expect the key to be created in the name of the provider user. If they change the provider configuration but the resource itself does not filter to owned true, the resource has no context awareness about that change and the api key will be marked as no change required while in reality the configuration changed the user who shall own the api key.

This isn't the behaviour today though, existing configurations managing a key not owned by the current user would then suddenly be recreating that key without users expecting it. Setting this as the default, rather than an explicit opt-in breaks for any team using multiple individual users backed by the same role to manage a single Terraform workspace. 

 > This means users expect the key to be created in the name of the provider user
 > the configuration changed the user who shall own the api key
 
We've previously made no claim about this, whilst it logically follows in some contexts, it's not an absolute. Given the existing attributes, the only real access guarantee is that the referenced key exists and is readable by the current user. Enforcing ownership is what makes this break existing customer workflows, in a potentially pretty destructive way (i.e an in-use API key is invalidated and data is lost).

aklakina:
> This isn't the behaviour today though, existing configurations managing a key not owned by the current user would then suddenly be recreating that key without users expecting it.

But such state could only ever exist if someone would manually import an api_key that was created manually. Otherwise, this provider can never produce an api_key in the name of a different user. But this state is also misleading in my opinion because the resource now only checks if the api_key exists and does not care about the owner of the key.

> Enforcing ownership is what makes this break existing customer workflows, in a potentially pretty destructive way (i.e an in-use API key is invalidated and data is lost).

Also the keys would not get recreated because the refresh would not find the original one so it would only say that it will create the api_key resource because it disappeared but the original key would not be touched anymore.

> We've previously made no claim about this, whilst it logically follows in some contexts, it's not an absolute. Given the existing attributes, the only real access guarantee is that the referenced key exists and is readable by the current user.

In case of resource import, this is true, but what about resource creation?

tobio:
> But such state could only ever exist if someone would manually import an api_key that was created manually

It's possible to manage a single Terraform workspace with different Elasticsearch users, which would create this situation. For example, I can create the initial set of resources, share the TF modules/state, and have a second team member apply another set of changes. As long as we both have appropriate permissions, there's no issue. There's nothing in the provider that enforces key ownership, and adding it is a breaking change that's not really justified.

aklakina:
Hmm, yes. I did not in fact consoder a scenario where the team members use their own users to run the same workspace. Alright so back to the core question, would you be fine with the 2nd implementation?

tobio:
> 2. The bug fix, and control
> Behaviour from 1, with an optional restrict_to_owned attribute (not tied to that name, but more IMO something more explicit that just owner). That new attribute limits read/delete to owner=true and so avoids accidentally deleting a shared key in the case of multiple privileged users applying the same TF stack.

Just so I'm clear, this being the 2nd implementation? If so, yep this sounds like the right path.

aklakina:
Yes, that's what I meant.

Okay, I'm on vacation now, will implement the changes next week.

aklakina:
Pushed the implementation we agreed on (bug fix + control):

- `DeleteAPIKey` now always attempts the invalidate scoped to the current user (`owner=true`) first, so the common case works with just `manage_own_api_key`. If that doesn't invalidate the key (e.g. it's owned by someone else), it automatically falls back to an unscoped invalidate (`owner=false`, requiring `manage_api_key`). No attribute needed for this to "just work".
- Renamed the `owner` attribute to `restrict_to_owned` (default `false`). When set to `true`, the fallback is disabled — the key is only ever read/deleted if owned by the calling user, otherwise treated as non-existent, so a shared key is never touched.

Updated unit tests to cover the new fallback behavior (including that `restrict_to_owned=true` makes only a single scoped request, no fallback) and updated docs. Let me know if this matches what you had in mind.

tobio:
@aklakina it looks like [this](https://github.com/elastic/terraform-provider-elasticstack/commit/8c9d09ba92a418a233b852dc2dd6a903c2692a03) commit has a different email as the author which isn't associated with your GH account and so is failing the CLA check. 

Are you able to rewrite the author of this commit and force push the change so we can tick that box?

aklakina:
Hi @tobio

Can you check if my downmerge conflict resolution is correct?

aklakina:
Noticed the `Go Lint (golangci-lint)` job fails with a version mismatch error: `.custom-gcl.yml` pins `version: v2.13.1`, but `.github/workflows/provider.yml`'s `golangci-lint-action` step pins `version: v2.12.2`, and the action hard-fails when these don't match exactly. This is present on `main` as well, so it's a pre-existing repo issue unrelated to this PR's changes. I've bumped the workflow's version to `v2.13.1` to align it with `.custom-gcl.yml` and unblock this check.

PR #4460, opened 2026-08-07
Propose OpenSpec change for Kibana Alerting v2 rule resource

dominiqueclarke:
## Changelog
Customer impact: none

## Summary
- Design-only OpenSpec change for a new experimental Terraform resource `elasticstack_kibana_rule` covering Kibana Alerting v2 rules ([rna-program#673](https://github.com/elastic/rna-program/issues/673)).
- Adds proposal, design, tasks, and delta specs under `openspec/changes/add-kibana-alerting-v2-rule-resource/` — no provider implementation in this PR.
- Records contingent Alerting v2 API asks, PUT-only write path, typed query schema, and experimental registration behind `TF_ELASTICSTACK_INCLUDE_EXPERIMENTAL`.

## Test plan
- [ ] Review naming (`elasticstack_kibana_rule`) and experimental release path
- [ ] Review PUT-only CRUD decision and read-before-write mitigation in `design.md`
- [ ] Review delta requirements in `specs/kibana-rule/spec.md`
- [ ] Confirm `openspec validate add-kibana-alerting-v2-rule-resource --type change` stays green
- [ ] No implementation / `make build` required for this PR


openspec/changes/add-kibana-alerting-v2-rule-resource/specs/kibana-rule/spec.md
```diff
@@ -0,0 +1,585 @@
...
+  kind       = <required, string>            # "alert" | "signal"; RequiresReplace
+  time_field = <optional, computed, string>  # default "@timestamp"; 1..128 chars
+  enabled    = <optional, computed, bool>    # default true
+
+  metadata {                                  # required, single nested block
+    name         = <required, string>         # 1..256
+    description  = <optional, string>         # <=1024
+    owner        = <optional, string>         # <=256
+    tags         = <optional, set(string)>    # non-empty when set
+    builder_type = <optional, string>         # <=64
```
dominiqueclarke, line 38: Do we want terraform consumers to be able to set builder type, which requires a very specific ESQL shape to work properly with the UI, I think not.

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
```diff
@@ -0,0 +1,113 @@
...
+4. **`ForceNew` on the discriminator would present destroy-and-recreate as if it were a migration.** No conversion exists. A practitioner who flips `engine` would see a plan that looks like an upgrade and get a deleted rule plus a brand-new one with different history and a different ID.
+
+There is a fifth, structural cost: a discriminated resource forces every attribute to `Optional` so that both variants can validate, which moves all the required/forbidden logic out of the schema and into hand-written `ValidateConfig` code. That is precisely the failure mode this proposal is trying to escape from with `params`.
+
+### Decision 2 — naming
+
+| Object | Proposed Terraform type |
+|--------|-------------------------|
+| v2 rule | `elasticstack_kibana_rule` |
+| v1 rule (rename, future) | `elasticstack_kibana_alerting_rule_classic` |
```
dominiqueclarke: A rename is not strictly necessary, it's presented here just for thought.

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
```diff
@@ -0,0 +1,113 @@
...
+
+**Do not reuse `elasticstack_kibana_alerting_rule` for v2.** Renaming v1 vacates the name, but reclaiming it is not safe. A practitioner who upgrades the provider without editing configuration would have Terraform read existing v1 state against the v2 schema. Renaming v1 is a supported *address* move — `terraform-plugin-framework` v1.19.0 (in `go.mod`) provides `resource.ResourceWithMoveState`, and the repo pins Terraform 1.15.8, well past the 1.8 floor for cross-type `moved` blocks — so v1 state can be carried to the new address with a `moved` block and no destroy. Reusing the vacated name is a *schema* substitution and has no such mechanism.
```
dominiqueclarke: The inability to reuse `elasticsearch_kibana_alerting_rule` is important here, but please don't let any of the other renaming bits be too distracting. Details on rename of alerting v1 are provided for thought.

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
```diff
@@ -0,0 +1,113 @@
...
+
+**3. Two different fields are called "version".** `metadata.version` is an integer configuration counter incremented on every mutation (surfaced on rule events as `rule.version`). Top-level `version` is the saved object optimistic-concurrency token, a string. Both appear in the same response object. The Terraform schema has to expose both, and no naming makes that unconfusing. **Ask: rename one — `metadata.revision` for the counter, or `occ_token`/`sequence_token` for the OCC string.**
+
+**4. Privilege names embed `alerting-v2`.** The API privileges for rules are `read-alerting-v2-rules` and `write-alerting-v2-rules`. These are user-authored strings in `elasticstack_kibana_security_role` and `elasticstack_elasticsearch_security_role` configurations. If classic alerting is retired and these are renamed, every Terraform-managed role definition that grants them breaks. **Ask: confirm the names are permanent, or settle them before the API is public.**
```
dominiqueclarke: The alerting-rules privilege key isn't taken currently, we should consider if we want to claim the name now instead of alerting-v2-rules.

openspec/changes/add-kibana-alerting-v2-rule-resource/design.md
```diff
@@ -0,0 +1,183 @@
+## Context
+
+See `proposal.md` — Why, for motivation, and the naming and release-path decisions. This document covers how the resource maps onto the API and why each mapping choice was made. Requirements are in `specs/kibana-rule/spec.md`; this file does not restate them.
+
+**Source of truth.** Alerting v2 is not yet in Kibana's published OAS bundle on `main`. Every schema claim here is taken from the zod schemas in `x-pack/platform/packages/shared/response-ops/alerting-v2-schemas/src/` (principally `rule_data_schema.ts`, `common.ts`, `constants.ts`) and the route and client code in `x-pack/platform/plugins/shared/alerting_v2/server/`, and cross-checked against the OAS bundle on the branch of [kibana#279519](https://github.com/elastic/kibana/pull/279519). Where a claim comes from server behaviour rather than the schema — for example how `PUT` treats `enabled` — it is attributed to `lib/rules_client/rules_client.ts`.
```
dominiqueclarke: All schemas will continue to be cross checked against main and OAS docs until this PR is merged, we'll be mindful of any inprogress changes.

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
```diff
@@ -0,0 +1,113 @@
...
+- Model the ES|QL `query` discriminated union, `schedule`, `state_transition`, `grouping`, and `artifacts` as **typed Terraform attributes**, not a JSON blob.
+- Register the resource in `experimentalResources()` behind `TF_ELASTICSTACK_INCLUDE_EXPERIMENTAL`, as a technical preview with no generated registry documentation. Graduation is [rna-program#678](https://github.com/elastic/rna-program/issues/678) and is designed to need no schema, API, or state change at that point.
+- Record a set of **contingent API changes** to request from the Alerting v2 team (see below). One of them — writable `enabled` — materially changes the resource's apply semantics, so it is called out as a decision the design depends on rather than a wish list.
+
+**Not in this change:** any Go code, `generated/kbapi` regeneration, an action-policy resource, or the rename of the v1 resource (the last is argued in Decision 2; delivered separately).
+
+### Decision 1 — a separate resource type, not a discriminator on the existing one
```
tobio, line 19: 👍

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
```diff
@@ -0,0 +1,113 @@
...
+
+| Object | Proposed Terraform type |
+|--------|-------------------------|
+| v2 rule | `elasticstack_kibana_rule` |
+| v1 rule (rename, future) | `elasticstack_kibana_alerting_rule_classic` |
+
+The tracking issues currently say `elasticstack_kibana_alerting_rule_v2`, marked "name TBD". Three arguments against carrying the version into the type name:
+
+- **`_v2` names age badly.** The version suffix is meaningful only while both engines exist. Once classic alerting is retired, the provider is left with a resource permanently called `_v2` and no `_v1`, and renaming it then is a second breaking change.
+- **Dropping the `alerting_` prefix has precedent.** `elasticstack_kibana_maintenance_window` is an alerting-framework object and does not carry the prefix.
```
tobio: This isn't really an accurate characterisation. Maintenance windows are a top level API resource ([`/api/maintenance_window`](https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-maintenance-window)) and TF reflects that. 

Dropping `alerting` is pretty difficult to justify given all the different _rules_ within Kibana. That's covered below, but IMO that just highlights the confusion and inconsistency that having an unqualified rule resource would create.  

> Once classic alerting is retired, the provider is left with a resource permanently called `_v2` and no `_v1`, and renaming it then is a second breaking change.

There's not really a _second_ breaking change here. It would be the first breaking change for the users of the v2 alerting rule. 

Looking through other popular providers (e.g AWS), the `v2` naming is pretty common.

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
tobio: > and the repo pins Terraform 1.15.8, well past the 1.8 floor for cross-type `moved` blocks — so v1 state can be carried to the new address with a `moved` block and no destroy.

There's a bit of conflation happening here. The repo pins a Terraform version for our internal CI, that doesn't correspond to customer usage though. Without `moved` support, the rename becomes a pretty painful breaking change for customers and one that's pretty difficult to justify IMO.

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
dominiqueclarke: We don't really want to refer to the system as v2 externally, but that's really good to know. It'll be an important discussion point.

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
dominiqueclarke: Of course, makes sense. I'll stress that much further.

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
dominiqueclarke: I think it's also important to note that although there are other "rules" in KIbana, the rule resource we're proposing is for the platform level rule, not an observability or security flavor of rule.

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
```diff
@@ -0,0 +1,91 @@
...
+
+**2. Response casing is inconsistent.** `ruleResponseSchema` mixes snake_case configuration fields (`time_field`, `recovery_strategy`, `no_data_strategy`, `state_transition`) with camelCase server fields (`createdBy`, `createdAt`, `updatedBy`, `updatedAt`). The find API compounds it: the request takes `per_page`, the response returns `perPage`. This flows straight into the generated Go types and into every mapping function. **Ask: pick one convention for the v2 surface — snake_case, matching the request bodies.**
+
+**3. Two different fields are called "version".** `metadata.version` is an integer configuration counter incremented on every mutation (surfaced on rule events as `rule.version`). Top-level `version` is the saved object optimistic-concurrency token, a string. Both appear in the same response object. The Terraform schema has to expose both, and no naming makes that unconfusing. **Ask: rename one — `metadata.revision` for the counter, or `occ_token`/`sequence_token` for the OCC string.**
```
dominiqueclarke, line 67: What use do consumers have for the top level `version`? Is it useful to return from the response body at all? `metadata.version` is what is used to correlate rule version info.

openspec/changes/add-kibana-alerting-v2-rule-resource/tasks.md
```diff
@@ -0,0 +1,77 @@
...
+
+## 1. Upstream prerequisites and open questions
+
+- [ ] 1.1 File the three contingent API requests from `proposal.md` against the Alerting v2 team, linking each to this change. Prioritise **3 (two fields named "version")** before graduation ([#678](https://github.com/elastic/rna-program/issues/678)) — it is the only contingent ask that would force a Terraform schema rename — and **1 (writable `enabled`)** for apply-semantics correctness (not a schema change).
+- [ ] 1.2 Track [kibana#279519](https://github.com/elastic/kibana/pull/279519) to merge (Check OAS Snapshot is currently red). Implementation of section 3 onwards is blocked on it; nothing in sections 1–2 is.
+- [ ] 1.3 Resolve `design.md` Open Question 1: determine the first Kibana version shipping `/api/alerting/v2/rules` as a public route, and confirm serverless behaviour. Record the value in the delta spec as a compatibility requirement.
+- [ ] 1.4 Raise `design.md` Open Question 2 (server-side v1 → v2 migration vs. the classic Terraform resource) with the Alerting v2 team and open a follow-up issue for whatever the classic resource needs to do about it.
```
dominiqueclarke, line 8: If we ever have a migration path for v1 to v2, we need to make sure we factor in how to handle terraform management rule resources, that's what I'm trying to say here.

openspec/changes/add-kibana-alerting-v2-rule-resource/design.md
```diff
@@ -0,0 +1,193 @@
...
+
+`created_at` and `created_by` do get `UseStateForUnknown`: `upsertRule` explicitly carries `createdBy` and `createdAt` forward from the existing object on replace.
+
+The `version` OCC token is exposed as a read-only attribute rather than hidden. It is useful for out-of-band tooling and for debugging conflicts, and hiding a field that appears in the API response tends to generate questions. It is never sent back; `PUT` does not accept it.
+
+`proposal.md` contingent change 3 asks for one of the two "version" fields to be renamed. If that happens the Terraform attribute names change with it, which is a schema change — this is the strongest reason to land that request before graduation.
```
dominiqueclarke, line 149: I think I don't really understand what top level version provides, for a consumer, and we should probably just not return it. Unless I'm missing something.

openspec/changes/add-kibana-alerting-v2-rule-resource/design.md
```diff
@@ -0,0 +1,193 @@
...
+`createRuleDataSchema` has no `enabled`. `RulesClient.createRule` hard-codes `enabled: true`; `upsertRule` copies `existingAttrs.enabled` forward. So:
+
+- create with `enabled = true` → one call;
+- create with `enabled = false` → `PUT`, then `POST .../_disable`;
+- update with no change to `enabled` → one call, because `PUT` preserves it;
+- update that toggles `enabled` → `PUT`, then `_enable` or `_disable`.
+
+The reconciliation mirrors `reconcileRuleEnabled` in `internal/clients/kibanaoapi/alerting_rule.go`, and runs before the authoritative read so state reflects the final server state.
+
+The window between the two calls is the problem, and it is worse here than in v1. A create with `enabled = false` briefly produces a running rule that can fire notifications before the disable lands. That is the substance of contingent change 1.
```
dominiqueclarke, line 162: "A create with `enabled = false` briefly produces a running rule that can fire notifications before the disable lands."
Because the create api creates with enabled true by default.

openspec/changes/add-kibana-alerting-v2-rule-resource/design.md
```diff
@@ -0,0 +1,193 @@
...
+
+If `enabled` becomes writable in the `PUT` body, the change is confined to the client wrapper: send the field, delete the follow-up. Spec REQ-011 is written so its scenarios hold either way, and `enabled` stays `Optional + Computed` with a `true` default in both worlds — so this is not a schema change and does not block graduation.
+
+### D6. Unwrap the zod-nullable union in codegen, not in every mapper
+
+`.nullable()` in zod v4 serialises to `anyOf: [T, {nullable: true}]`. `oapi-codegen` reads any `anyOf` as a union and emits a `json.RawMessage` wrapper with typed accessors, so `RuleResponse.CreatedBy` is a union rather than `*string`. Counting the pattern across the v2 components and paths in the [kibana#279519](https://github.com/elastic/kibana/pull/279519) bundle gives exactly 37 occurrences.
```
dominiqueclarke: Comment from Tina regarding this https://github.com/elastic/kibana/pull/279519#discussion_r3618605876

openspec/changes/add-kibana-alerting-v2-rule-resource/design.md
```diff
@@ -0,0 +1,258 @@
...
+
+Either way this is a codegen-and-client-wrapper concern. The Terraform schema is unaffected: `created_by` and `updated_by` are nullable computed strings and `state_transition` is an optional block regardless of how the Go type is shaped.
+
+### D7. Space handling by request editor, not by path rewriting
+
+The v2 rule saved object is `namespaceType: 'multiple-isolated'`, so rules are space-scoped and the composite `<space_id>/<rule_id>` identity applies.
+
+`transform_schema.go` has a `spaceIdPaths` allowlist that rewrites selected paths into `/s/{spaceId}` variants, but the classic alerting resource does not use it — it passes `kibanautil.SpaceAwarePathRequestEditor(spaceID)` to the generated client instead, rewriting the URL per request. This design does the same. It keeps the v2 paths out of the transform allowlist, so a `kbapi` regeneration that pulls in new v2 endpoints does not need a matching transform change, and it matches the resource this one is most likely to be read alongside.
```
dominiqueclarke, line 170: @tobio What is the preferred pattern here?

tobio: Using the request editor is the right approach here. I've been meaning to kick off a refactor to remove the `spaceIdPaths` approach for a while now.

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
tobio: The API resources are `v2` though? Is that going to change before release? I don't mind if it's `v2` or something else (e.g just plain `kibana_rule` if the product strongly believes in that naming). 

I'm mostly just trying to make sure we make a fully informed decision and the justification for what can be a confusing naming scheme feels weak at the moment.

openspec/changes/add-kibana-alerting-v2-rule-resource/design.md
dominiqueclarke: Spoke to Tina, a core fix is on the way.

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
```diff
@@ -0,0 +1,91 @@
...
+
+**1. `enabled` is not writable on create or update.** `createRuleDataSchema` has no `enabled` field; `ruleResponseSchema` does. `RulesClient.createRule` hard-codes `enabled: true`, and `upsertRule` copies the existing value forward. The only way to change it is `POST /rules/{id}/_enable` or `_disable`.
+
+For Terraform this means every apply that sets `enabled = false` is **two calls that are not atomic**: create/replace the rule (which starts it running), then disable it. A failure in between leaves a rule enabled that the configuration says should be disabled — and a briefly live rule can fire real notifications.
+
+The provider already carries the scar tissue from this pattern twice: `internal/clients/kibanaoapi/alerting_rule.go` has a `reconcileRuleEnabled` follow-up for v1, and `elasticstack_kibana_security_enable_rule` is an entire resource type that exists only because the detection rule API made the same choice. **Ask: accept `enabled` in the `PUT /rules/{id}` body (and in `PATCH`).** The `_enable`/`_disable` routes can stay for UI use.
```
dominiqueclarke:
```suggestion
The provider already provides workarounds for this pattern twice: `internal/clients/kibanaoapi/alerting_rule.go` has a `reconcileRuleEnabled` follow-up for v1, and `elasticstack_kibana_security_enable_rule` is an entire resource type that exists only because the detection rule API made the same choice. **Ask: accept `enabled` in the `PUT /rules/{id}` body (and in `PATCH`).** The `_enable`/`_disable` routes can stay for UI use.
```

mikecote (approved):

openspec/changes/add-kibana-alerting-v2-rule-resource/proposal.md
dominiqueclarke: It might change before release, we'll see. We'll definitely discuss with our product management and tech leads to come to on the naming. Though we still have `v2` in the api path, we've been working during this tech preview phase on decentering the v2 label as much as possible, with the intention of referring to v1 rules as Classic and v2 rules as just Rules. @mikecote @jasonrhodes @tiamliu

mikecote (commented):
A couple of notes after reviewing the proposal, looks good overall! Some of these are drive by comments we can channel accordingly:
- +1 to having the APIs use snake case ([HTTP API guidelines](https://www.elastic.co/docs/extend/kibana/contributing/api-design/guidelines-for-http-api-design-in-kibana))
- -1 to the `elasticstack_kibana_rule` naming, I'd prefer keeping `alerting` in it to describe what type of `rule` this is (how we name it internally and externally as well). It's ok to use `v2` in the naming as it's a second version / new resource type, and part of the API path.
- It makes me think we may eventually want an API endpoint per builder type as we start passing the various per-type container fields (something to keep in mind). Makes it easier to document, use and validate cc @cnasikas @darnautov 
- It would be nice if the API can just use one `version` and derive the other at the server layer.. If one is the rule version and the other is the ES document version, they are fairly much the same and hard to explain/distinguish to an API user, would be good if we can just use one.

openspec/changes/add-kibana-alerting-v2-rule-resource/specs/kibana-rule/spec.md
dominiqueclarke: @mikecote to your point maybe we disallow this for now, then support it via per builder type apis, as mentioned.

mikecote: I think that would make sense, as long as we have it sorted by GA. cc @cnasikas who has more context as I recall we've had issues with OAS, terraform and such by having a single API for all rule types in v1 and I'm sure we don't want to repeat issues in v2.

Issue #1408, opened 2025-10-30
[Feature] Configure artifacts on Kibana Alerting rule


keiransteele-phocas:
**Is your feature request related to a problem? Please describe.**
When creating a Kibana Alerting rule using elasticstack_kibana_alerting_rule we are unable to set artifacts such as dashboards or the investigation guide

We plan to create a large number of alert rules for new Kubernetes environments and our existing AWS infrastructure, creating and managing the alert rules manually across dev and prod clusters and is not optimal and we are going to use the Elasticstack provider to configure them alongside our infrastructure. Creation of the rules would be improved if we can add links to dashboards and an investigation guide alongside the alert definition.

**Describe the solution you'd like**
Add the ability to set artifacts when using elasticstack_kibana_alerting_rule

**Describe alternatives you've considered**
I have considered using the API directly but we would prefer to use Terraform alongside our other infrastructure deployment code.

**Additional context**
https://www.elastic.co/docs/api/doc/kibana/operation/operation-post-alerting-rule-id#operation-post-alerting-rule-id-body-application-json-artifacts

-----------------------

Proposed schema additions:

```
artifacts = {
  dashboards = [{
    id = "<dashboard_id>"
  }]
  investigation_guide = {
    content = "<guide>"
    content_path = "/path/to/guide"
    checksum = "<computed checksum of path>"
  }
}
```

Only one of `content` or `content_path` can be configured, with the checksum computed similarly to the custom fleet integration checksum.

tobio:
Blocked on https://github.com/elastic/kibana/pull/252210

PR #4489, opened 2026-08-10
feat(kibana_alerting_rule): add artifacts.investigation_guide support

kroustou:
Refs #1408.

## Changelog
Customer impact: enhancement
Summary: Add `artifacts.investigation_guide` support to `elasticstack_kibana_alerting_rule` (inline `content` or file-based `content_path` with drift-detecting checksum). Requires
 Elastic Stack 9.1.0+.

   ## What

   Adds an optional `artifacts.investigation_guide` to `elasticstack_kibana_alerting_rule`, mapping to the Kibana alerting rule API `artifacts.investigation_guide.blob` (available from
 Elastic Stack 9.1.0). Practitioners can co-locate a runbook / investigation guide with the alert rule as code.

   The guide can be provided two ways:

   ​```hcl
   resource "elasticstack_kibana_alerting_rule" "example" {
     # ...
     artifacts = {
       investigation_guide = {
         content = "## Runbook\nInvestigate ..."        # inline, OR
         # content_path = "${path.module}/guide.md"     # from a local file
       }
     }
   }
   ​```

   - `content` — inline Markdown, sent directly as `blob`.
   - `content_path` — path to a local file; its contents are sent as `blob`, and the provider computes a SHA-256 `checksum` at plan time to detect out-of-band changes (mirrors the
 `elasticstack_fleet_custom_integration` pattern).
   - Exactly one of `content` / `content_path` must be set.

   ## Implementation notes

   - `artifacts` / `investigation_guide` are modelled as `SingleNestedAttribute` rather than `SingleNestedBlock`. An omitted `SingleNestedBlock` still materialises as a non-null object,
 which made the nested `ExactlyOneOf(content, content_path)` validator fire on every rule with no artifacts. A nested attribute is genuinely null when omitted (same pattern as the
 existing optional `flapping` attribute).
   - A version gate in `GetVersionRequirements` requires Elastic Stack >= 9.1.0 when `artifacts` is set. 9.1.0 was confirmed against the Kibana OpenAPI (field first appears at 9.1.0,
 absent in 9.0.0).
   - Read path preserves the practitioner-chosen source; Create/Update record the concrete checksum before the read-after-write refresh (the API returns no checksum).
   - When `artifacts` is omitted, the key is left out of the create/update body so Kibana preserves any existing server-side artifacts.

   ## Scope

   Investigation guide only. The `artifacts.dashboards` portion of the OpenSpec change remains deferred.

   ## Testing

   - Unit tests: body construction, response mapping, model conversion, checksum computation.
   - Acceptance test `TestAccResourceAlertingRuleInvestigationGuide` (gated >= 9.1.0): inline create + update, `content_path` create, and out-of-band file edit -> non-empty update plan.
   - Verified end-to-end against local Elasticsearch + Kibana 9.1.3; existing `TestAccResourceAlertingRule` also passes.
   - `docs/resources/kibana_alerting_rule.md` regenerated via tfplugindocs.


internal/kibana/alertingrule/plan_modifier.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// On create (no prior state) or when the file changed, mark checksum unknown
+	// so the plan shows a change and Create/Update recomputes the concrete value.
+	if priorChecksum == "" || newChecksum != priorChecksum {
+		resp.Diagnostics.Append(setInvestigationGuideChecksumUnknown(ctx, &plan)...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+		resp.Diagnostics.Append(resp.Plan.Set(ctx, &plan)...)
+	}
```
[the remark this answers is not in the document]
tobio: The current code should be correct. Invalidating `id` is a deliberate choice in custom integrations since the ID  is derived from the integration itself and would change.

tobio (commented):
## Review — `artifacts.investigation_guide`

Adversarial review (3 fresh-context reviewers) converged on one blocker plus a consistent set of doc/test gaps. The write-side `>= 9.1.0` gate is correct (verified against the published Kibana v9.1.0 OpenAPI request bodies); the gap is read-back: Kibana only **returns** `artifacts` from 9.5.0 (elastic/kibana#247279, `backport:skip`).

Inline comments below cover 7 points:
1. `artifacts`/`investigation_guide` → `Optional`+`Computed`+`UseStateForUnknown` (fixes `inconsistent result after apply` on removal)
2. 9.5.0-gated import acceptance test
3. Document the read-back version limitation (import, drift detection)
4. Rename `block*` → `attr*` constants
5. Unit tests for `ModifyPlan` and `GetVersionRequirements`
6. `UseStateForUnknown` on `checksum`
7. Blob-normalisation coverage in the acc suite (9.5.0+)

Review feedback only — no files changed.

internal/kibana/alertingrule/schema.go
```diff
@@ -200,6 +202,38 @@ func getSchema(_ context.Context) schema.Schema {
 					},
 				},
 			},
+			blockArtifacts: schema.SingleNestedAttribute{
+				MarkdownDescription: artifactsDescription,
+				Optional:            true,
```
tobio: **Point 1 (1 of 2): `artifacts` → `Optional`+`Computed` + `UseStateForUnknown`.**

Removing `artifacts` from config triggers `Provider produced inconsistent result after apply` on stacks where GET returns artifacts (9.5.0+): the update omits `artifacts` (Kibana preserves the guide), but read-after-write's `priorIG == nil` inline branch (`models.go:291-297`) re-populates `m.Artifacts` from the preserved blob — state differs from the null plan. The sibling `flapping` attribute is `Optional`+`Computed` with `UseStateForUnknown()` precisely to tolerate this (see the `no_flapping` acc step). Apply the same to `artifacts` here and to `investigation_guide` below. (The key is also renamed to `attrArtifacts` — paired with the constants rename in Point 4.)

```suggestion
			attrArtifacts: schema.SingleNestedAttribute{
				MarkdownDescription: artifactsDescription,
				Optional:            true,
				Computed:            true,
				PlanModifiers: []planmodifier.Object{
					objectplanmodifier.UseStateForUnknown(),
				},
```

internal/kibana/alertingrule/schema.go
```diff
@@ -200,6 +202,38 @@ func getSchema(_ context.Context) schema.Schema {
...
+				Attributes: map[string]schema.Attribute{
+					blockInvestigationGuide: schema.SingleNestedAttribute{
+						MarkdownDescription: investigationGuideDescription,
+						Optional:            true,
```
tobio: **Point 1 (2 of 2): `investigation_guide` → `Optional`+`Computed` + `UseStateForUnknown`.**

Same reasoning as the `artifacts` block above — the nested object must also be `Computed` so the read path can repopulate it without an apply-time inconsistency. (`UseStateForUnknown` only fills *unknowns* from state, not nulls, so `toAPIModel`/`artifactsToAPI` still omit `artifacts` when the planned value is null — Kibana's preserve-on-omit semantics hold.)

```suggestion
				attrInvestigationGuide: schema.SingleNestedAttribute{
					MarkdownDescription: investigationGuideDescription,
					Optional:            true,
					Computed:            true,
					PlanModifiers: []planmodifier.Object{
						objectplanmodifier.UseStateForUnknown(),
					},
```

internal/kibana/alertingrule/schema.go
```diff
@@ -200,6 +202,38 @@ func getSchema(_ context.Context) schema.Schema {
...
+							attrContentPath: schema.StringAttribute{
+								MarkdownDescription: "Path to a local file whose contents are used as the investigation guide. " +
+									"The provider computes a SHA-256 `checksum` of the file at plan time to detect external changes. " +
+									"Mutually exclusive with `content`.",
+								Optional: true,
+							},
+							attrChecksum: schema.StringAttribute{
+								MarkdownDescription: "SHA-256 checksum of the file at `content_path`, used to detect drift. Computed; not user-settable.",
+								Computed:            true,
+							},
```
tobio, line 246: **Point 6: add `UseStateForUnknown` to `checksum`.**

Without it, every unrelated update plan shows `checksum: (known after apply)` noise that resolves to the same digest (file) or null (inline). `ModifyPlan` runs after attribute plan modifiers and still overrides it to unknown on file change, so this is safe and removes the churn.

```suggestion
							attrChecksum: schema.StringAttribute{
								MarkdownDescription: "SHA-256 checksum of the file at `content_path`, used to detect drift. Computed; not user-settable.",
								Computed:            true,
								PlanModifiers: []planmodifier.String{
									stringplanmodifier.UseStateForUnknown(),
								},
							},
```

internal/kibana/alertingrule/constants.go
```diff
@@ -36,6 +36,12 @@ const (
 	attrThrottle      = "throttle"
 	blockFrequency    = "frequency"
 	blockAlertsFilter = "alerts_filter"
+
+	blockArtifacts          = "artifacts"
+	blockInvestigationGuide = "investigation_guide"
```
tobio: **Point 4 (1 of 3): rename `block*` → `attr*`.**

These declare `SingleNestedAttribute`s, not blocks — the `block` prefix misleads readers into block-syntax HCL. Rename to match the `attrContent`/`attrContentPath`/`attrChecksum` siblings. The two `getSchema` map-key usages are renamed in the Point 1 suggestions above; the two `initAttrTypes` references are renamed below.

```suggestion
	attrArtifacts          = "artifacts"
	attrInvestigationGuide = "investigation_guide"
```

internal/kibana/alertingrule/schema.go
```diff
@@ -328,6 +362,12 @@ func initAttrTypes() {
 
 	flapAttr := s.Attributes["flapping"].(schema.SingleNestedAttribute)
 	cachedFlappingTypes = flapAttr.GetType().(attr.TypeWithAttributeTypes).AttributeTypes()
+
+	artifactsAttr := s.Attributes[blockArtifacts].(schema.SingleNestedAttribute)
```
tobio: **Point 4 (2 of 3):** update the `initAttrTypes` reference.

```suggestion
	artifactsAttr := s.Attributes[attrArtifacts].(schema.SingleNestedAttribute)
```

internal/kibana/alertingrule/schema.go
```diff
@@ -328,6 +362,12 @@ func initAttrTypes() {
...
+	cachedArtifactsTypes = artifactsAttr.GetType().(attr.TypeWithAttributeTypes).AttributeTypes()
+
+	investigationGuideAttr := artifactsAttr.Attributes[blockInvestigationGuide].(schema.SingleNestedAttribute)
```
tobio: **Point 4 (3 of 3):** update the `initAttrTypes` reference.

```suggestion
	investigationGuideAttr := artifactsAttr.Attributes[attrInvestigationGuide].(schema.SingleNestedAttribute)
```

tobio (commented):
## Review — `artifacts.investigation_guide` (adversarial, 3 reviewers)

The write-side `>= 9.1.0` gate is correct (verified against the published Kibana v9.1.0 OpenAPI request bodies); the gap is read-back — Kibana only **returns** `artifacts` from 9.5.0 (elastic/kibana#247279, `backport:skip`).

Inline suggestions below cover the 7 requested points. **Points 1, 6, 4, 3, 2, 7 are click-to-accept `suggestion` blocks.** Point 5 (new unit tests needing a new file / imports) is provided as code blocks since it isn't inline-acceptable.

1. `artifacts`/`investigation_guide` → `Optional`+`Computed`+`UseStateForUnknown` (fixes `inconsistent result after apply` on removal)
2. 9.5.0-gated import acceptance test
3. Document the read-back version limitation (import, drift detection)
4. Rename `block*` → `attr*` constants + references
5. Unit tests for `ModifyPlan` and `GetVersionRequirements`
6. `UseStateForUnknown` on `checksum`
7. Blob-normalisation coverage in the acc suite (9.5.0+)

_(The earlier empty review above has been superseded — its inline comments were removed in favour of these suggestions.)_

internal/kibana/alertingrule/descriptions/artifacts.md
```diff
@@ -0,0 +1 @@
+Linked assets for the alerting rule. Currently supports attaching an investigation guide so that runbooks and context are co-located with the alert configuration as code.
```
tobio: **Point 3 (1 of 2): document the read-back version limitation.**

`GetVersionRequirements` enforces `>= 9.1.0` and the sibling `flapping` description documents its version floors, but this description records none. The write/read asymmetry is invisible to users: create/update accept `artifacts` from 9.1.0, but Kibana only **returns** `artifacts` from 9.5.0 (elastic/kibana#247279, `backport:skip`). On 9.1.0–9.4.x the provider is write-only — import won't populate `artifacts` and out-of-band edits are invisible to refresh. Then regenerate `docs/resources/kibana_alerting_rule.md`.

```suggestion
Linked assets for the alerting rule. Currently supports attaching an investigation guide so that runbooks and context are co-located with the alert configuration as code. Requires Elastic Stack 9.1 or higher to write; on stacks older than 9.5.0 the Kibana GET API does not return artifacts (elastic/kibana#247279), so `terraform import` will not populate this block and external changes to the guide made outside Terraform will not be detected by refresh. Omitting this block does not clear artifacts already stored on the rule in Kibana.
```

internal/kibana/alertingrule/descriptions/investigation_guide.md
```diff
@@ -0,0 +1 @@
+An investigation guide attached to the rule. Provide the guide either inline via `content`, or from a local file via `content_path` (in which case the provider tracks a SHA-256 `checksum` of the file to detect external changes). Exactly one of `content` or `content_path` must be set.
```
tobio: **Point 3 (2 of 2):** mirror a short version note in the investigation guide description.

```suggestion
An investigation guide attached to the rule. Provide the guide either inline via `content`, or from a local file via `content_path` (in which case the provider tracks a SHA-256 `checksum` of the file to detect external changes). Exactly one of `content` or `content_path` must be set. Requires Elastic Stack 9.1 or higher to write; inline-`content` round-trips from the API and `terraform import` require 9.5.0+ (elastic/kibana#247279).
```

internal/kibana/alertingrule/acc_test.go
```diff
@@ -727,6 +728,96 @@ func TestAccResourceAlertingRuleFlapping(t *testing.T) {
...
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("elasticstack_kibana_alerting_rule.test_rule", plancheck.ResourceActionUpdate),
+					},
+				},
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet("elasticstack_kibana_alerting_rule.test_rule", "artifacts.investigation_guide.checksum"),
+				),
+			},
+		},
```
tobio, line 890: **Point 2: add a 9.5.0-gated import acceptance test.**

The current inline-`content` assertions pass on 9.1.0–9.4.x purely from state preservation (`populateArtifactsFromAPI` early-returns when `!apiHasIG`, `models.go:277-284`), so they're green regardless of whether Kibana stored the blob. Import starts from an empty model, so an import step is the assertion that actually exercises the elastic/kibana#247279 GET round-trip. Insert as the last step, gated at 9.5.0 (existing write steps stay at 9.1.0).

```suggestion
			},
			// Import: only meaningful where GET returns artifacts (Kibana 9.5.0+, elastic/kibana#247279).
			// The inline-content assertions above pass on 9.1–9.4 from state preservation alone;
			// import is the only flow that starts from an empty model and proves the GET round-trip.
			{
				ProtoV6ProviderFactories: acctest.Providers,
				SkipFunc:                 versionutils.CheckIfVersionIsUnsupported(version.Must(version.NewSemver("9.5.0"))),
				ResourceName:             "elasticstack_kibana_alerting_rule.test_rule",
				ImportState:              true,
				ImportStateVerify:        true,
				ImportStateVerifyIgnore:  []string{"notify_when", "last_execution_date", "last_execution_status"},
				ConfigDirectory:          acctest.NamedTestCaseDirectory("content_create"),
				ConfigVariables: config.Variables{
					"name":    config.StringVariable(ruleName),
					"rule_id": config.StringVariable(ruleID),
					"content": config.StringVariable("# Runbook\nUpdated inline guide."),
				},
			},
		},
```

internal/kibana/alertingrule/acc_test.go
```diff
@@ -727,6 +728,96 @@ func TestAccResourceAlertingRuleFlapping(t *testing.T) {
...
+					"name":    config.StringVariable(ruleName),
+					"rule_id": config.StringVariable(ruleID),
+					"content": config.StringVariable("# Runbook\nInitial inline guide."),
+				},
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("elasticstack_kibana_alerting_rule.test_rule", "artifacts.investigation_guide.content", "# Runbook\nInitial inline guide."),
+					resource.TestCheckNoResourceAttr("elasticstack_kibana_alerting_rule.test_rule", "artifacts.investigation_guide.content_path"),
+				),
+			},
+			// Inline content: update the text.
```
tobio, line 799: **Point 7: include blob-normalisation coverage (9.5.0+).**

Whether Kibana modifies the investigation-guide blob (trims whitespace, normalises line endings) before storing is unverified and can only be settled against a real 9.5 stack. Since 9.5 is available in CI, add a step that writes CRLF + trailing-newline content and asserts the read-back `content` equals what was sent. If it fails, capture the returned blob and decide the normalisation policy before merge — don't weaken to `Contains`.

```suggestion
			},
			// Blob normalisation probe (Kibana 9.5.0+): confirms Kibana stores/returns the
			// investigation-guide blob verbatim. If this fails, normalise in the read path
			// (or treat the blob as opaque) to avoid perpetual drift for inline-content users.
			{
				ProtoV6ProviderFactories: acctest.Providers,
				SkipFunc:                 versionutils.CheckIfVersionIsUnsupported(version.Must(version.NewSemver("9.5.0"))),
				ConfigDirectory:          acctest.NamedTestCaseDirectory("content_create"),
				ConfigVariables: config.Variables{
					"name":    config.StringVariable(ruleName),
					"rule_id": config.StringVariable(ruleID),
					"content": config.StringVariable("# Runbook\r\nStep 1.\n"),
				},
				Check: resource.ComposeTestCheckFunc(
					resource.TestCheckResourceAttr(
						"elasticstack_kibana_alerting_rule.test_rule",
						"artifacts.investigation_guide.content",
						"# Runbook\r\nStep 1.\n",
					),
				),
			},
			// Inline content: update the text.
```

internal/kibana/alertingrule/plan_modifier.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+)
+
+// ModifyPlan detects external changes to a file referenced by
+// artifacts.investigation_guide.content_path. When the file's SHA-256 differs
+// from the checksum recorded in state (or the resource is being created), the
+// computed checksum is marked unknown so Terraform produces a non-empty plan.
+func (r *Resource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
```
tobio, line 35: **Point 5a: add unit tests for `ModifyPlan`.**

`ModifyPlan` has zero unit coverage; the "unknown `content_path` skips checksum computation" behavior is only guarded by `typeutils.IsKnown(planIG.ContentPath)` at `plan_modifier.go:56` with no test. Provided as a code block (not a suggestion) because it belongs in a new `plan_modifier_test.go` (new file + imports — not click-acceptable inline). Suggested cases:

```go
func TestModifyPlan_contentPathFileChangedMarksChecksumUnknown(t *testing.T) {
	// state has checksum X; write a file whose SHA-256 != X; run ModifyPlan;
	// assert plan.artifacts.investigation_guide.checksum becomes unknown.
}

func TestModifyPlan_contentPathUnknownSkipsChecksum(t *testing.T) {
	// plan.investigation_guide.content_path is types.StringUnknown();
	// assert ModifyPlan does not read any file and leaves checksum unchanged.
}

func TestModifyPlan_inlineContentIsNoOp(t *testing.T) {
	// content set, content_path null -> ModifyPlan returns without marking checksum unknown.
}

func TestModifyPlan_destroyIsNoOp(t *testing.T) {
	// req.Plan.Raw.IsNull() -> early return, no diagnostics.
}
```

internal/kibana/alertingrule/models.go
```diff
@@ -398,6 +502,14 @@ func (m alertingRuleModel) GetVersionRequirements(ctx context.Context) ([]entity
 		}
 	}
 
+	// 9.1.0 when Artifacts (investigation guide) is set
```
tobio, line 505: **Point 5b: add unit tests for `GetVersionRequirements` (artifacts gate).**

The 9.1.0 artifacts gate has no test coverage — grep for `GetVersionRequirements`/`artifactsMinSupportedVersion` across `internal/kibana/alertingrule/*_test.go` returns nothing. Provided as a code block (not a suggestion) because it needs `github.com/hashicorp/go-version` + `entitycore` imports not present in `models_artifacts_test.go`. This locks in the 9.1.0 write-side gate independent of the 9.5.0 read-back gap in the `artifacts.md` note.

```go
func TestGetVersionRequirements_artifactsSetRequires91(t *testing.T) {
	m := alertingRuleModel{Artifacts: types.ObjectValueMust(getArtifactsAttrTypes(), artifactsModel{
		InvestigationGuide: types.ObjectValueMust(getInvestigationGuideAttrTypes(), investigationGuideModel{
			Content: types.StringValue("x"),
		}),
	})}
	reqs, diags := m.GetVersionRequirements(context.Background())
	require.False(t, diags.HasError())
	require.Len(t, reqs, 1)
	require.True(t, reqs[0].MinVersion.Equal(version.Must(version.NewVersion("9.1.0"))))
}

func TestGetVersionRequirements_noArtifactsNoRequirement(t *testing.T) {
	for _, art := range []types.Object{types.ObjectNull(getArtifactsAttrTypes())} {
		m := alertingRuleModel{Artifacts: art}
		reqs, _ := m.GetVersionRequirements(context.Background())
		require.Empty(t, reqs)
	}
}
```

kroustou:
Thanks, this is thorough. Addressed points 1, 2, 3, 4, 5a, 5b, 6, 7 in faab559b:
  - artifacts and investigation_guide are now Optional+Computed+UseStateForUnknown; checksum got UseStateForUnknown too.
  - Renamed the block* constants to attr*.
  - Documented the 9.1.0-write / 9.5.0-read-back asymmetry (elastic/kibana#247279) in both descriptions and regenerated docs.
  - Extracted a pure contentPathChecksumChanged helper and added unit tests for the drift cases plus GetVersionRequirements.
  - Added the 9.5.0-gated import step (placed after the inline steps, since a content_path prior state wouldn't match an import that returns the blob as content) and the CRLF normalization probe.

Ran the full test against a real Kibana 9.5.0: everything passes, and the normalization probe confirms Kibana returns the blob verbatim (no read-path normalization required).

tobio (approved):
Thanks for this!

PR #4520, opened 2026-08-12
feat(kibana_alerting_rule): link dashboards via artifacts.dashboards

kroustou:
Refs #1408. Follow-up to #4489 (which added `artifacts.investigation_guide`).

## Changelog

Customer impact: enhancement
Summary: Add `artifacts.dashboards` support to `elasticstack_kibana_alerting_rule`, letting practitioners link Kibana dashboards to a rule (alongside or instead of an investigation guide). Requires Elastic Stack 9.1.0+.

## What

Completes the deferred dashboards portion of the alerting rule artifacts work. Dashboards are linked by their saved-object id and map to the Kibana API `artifacts.dashboards[].id`:

```hcl
resource "elasticstack_kibana_alerting_rule" "example" {
  # ...
  artifacts = {
    dashboards = [
      { id = "dashboard-a" },
      { id = "dashboard-b" },
    ]
    # investigation_guide = { content = "..." }   # optional, still supported
  }
}
```

At least one of `investigation_guide` or `dashboards` must be set.

## Implementation notes

- `dashboards` is a `ListNestedAttribute` (each element has a required `id`), `Optional`+`Computed` with `UseStateForUnknown()`, consistent with the attribute-based `artifacts`.
- The `artifacts` object validator is relaxed from `AlsoRequires(investigation_guide)` to a `ValidateConfig` "at least one of investigation_guide or dashboards" check. It rejects `artifacts = {}` but only when both nested attributes are known-null, so an unknown dashboards list (e.g. built from a variable/for-expression or another resource) does not false-positive.
- The artifacts rebuild sites (`populateArtifactsFromAPI`, `artifactsToAPI`, `applyInvestigationGuideChecksum`, and the plan modifier`s checksum invalidation) were refactored onto a shared `buildArtifactsObject` helper so touching one field never drops the other.
- Read path preserves prior dashboards when Kibana omits artifacts (pre-9.5.0 GET does not return them, elastic/kibana#247279). On 9.5.0+ dashboards round-trip and `terraform import` populates them.
- On the request, `investigation_guide` is omitted when only dashboards are configured (and vice versa).

## Testing

- Unit tests: dashboards request-body construction and response mapping; model conversion (dashboards-only and combined with a guide); dashboards preservation when the API omits artifacts; empty-`artifacts` validation.
- Acceptance test `TestAccResourceAlertingRuleArtifactsDashboards`: create with two dashboards, update the list (add/remove), and a 9.5.0-gated import step proving the GET round-trip.
- Verified end-to-end against a real Elasticsearch + Kibana 9.5.0 stack: the dashboards test, the existing `TestAccResourceAlertingRuleInvestigationGuide`, and `TestAccResourceAlertingRule` all pass. Kibana accepts arbitrary dashboard ids on the rule.
- Regenerated `docs/resources/kibana_alerting_rule.md`; `make check-openspec` passes; custom golangci-lint clean.

## OpenSpec

Updates the existing `kibana-alerting-rule-artifacts` change: marks the deferred dashboards tasks done and aligns the delta spec (REQ-045/047) with the shipped attribute-based schema and the at-least-one rule.


tobio (commented):
Automated review pass — flagging two correctness issues and a simplification opportunity for the validator logic.

internal/kibana/alertingrule/models.go
```diff
@@ -628,34 +727,55 @@ func (m alertingRuleModel) toAPIModel(ctx context.Context) (models.AlertingRule,
...
+			if err != nil {
+				diags.AddError(
+					"Cannot read investigation guide file",
+					fmt.Sprintf("Failed to read content_path %q: %s", ig.ContentPath.ValueString(), err),
+				)
+				return nil, diags
+			}
+			blob = string(data)
+		}
+		if blob != "" {
```
tobio: **Bug:** explicit `content = ""` silently drops the investigation guide from the request instead of sending an empty blob.

```go
if blob != "" {
    out.InvestigationGuide = &models.AlertingRuleInvestigationGuide{Blob: blob}
}
```

`content = ""` is a valid, non-null config value and passes `validateArtifactsNotEmpty` (since `investigation_guide` is non-null), but this guard means `out.InvestigationGuide` stays `nil`. If no `dashboards` are configured either, `artifactsToAPI` returns `nil` entirely and the whole `artifacts` key is omitted from the create/update body — a silent no-op rather than actually writing the empty guide.

Worse, on the following read-after-write, if the server still has the old (or no) guide content, Terraform can see a value that doesn't match the planned `content = ""` and fail with "provider produced inconsistent result after apply."

Repro:
```go
m.Artifacts = artifactsObjectWith(ctx, t,
    mustIGObj(ctx, t, investigationGuideModel{Content: types.StringValue(""), ContentPath: types.StringNull(), Checksum: types.StringNull()}),
    dashboardsList(ctx, t),
)
rule, _ := m.toAPIModel(ctx)
// rule.Artifacts == nil
```

Suggest dropping the `blob != ""` check and setting `out.InvestigationGuide` whenever `ig != nil`, matching the pre-refactor behavior.

internal/kibana/alertingrule/models.go
```diff
@@ -244,65 +250,158 @@ func (m alertingRuleModel) investigationGuideFrom(ctx context.Context) (*investi
...
+		dashList, d := types.ListValueFrom(ctx, getDashboardsElementType(), dashboards)
+		diags.Append(d...)
+		if diags.HasError() {
+			return diags
+		}
+		am.Dashboards = dashList
+	} else if prior != nil && typeutils.IsKnown(prior.Dashboards) {
+		// API omitted dashboards but some were configured: preserve prior value.
+		am.Dashboards = prior.Dashboards
 	}
```
tobio: **Bug:** on Kibana 9.5.0+ (where GET *does* return `artifacts`), if a rule has both an investigation guide and dashboards configured, and the guide is later deleted out-of-band (e.g. via the Kibana UI) while dashboards remain, refresh won't detect it — the stale guide from prior state is kept instead of being nulled out. Same issue in reverse for dashboards being removed while the guide remains (mirrored block below, `} else if prior != nil && typeutils.IsKnown(prior.Dashboards) {`).

This "preserve prior on omission" fallback is meant for pre-9.5.0 stacks where the *entire* `artifacts` key is omitted from the response (already handled correctly by the `!apiHasIG && !apiHasDashboards` branch above this). But it also fires when `artifacts` *is* present and genuinely contains only one sub-field — there's no way here to distinguish "whole API contract doesn't return this" from "this field is genuinely gone now."

Repro: seed state with a guide + 1 dashboard, then call `populateArtifactsFromAPI` with a response containing only `Dashboards` (guide externally removed) — the returned model still has the old guide content.

This is a narrower edge case than the `content = ""` issue above (requires 9.5.0+, a previously-stored sub-field, and out-of-band deletion of just that field), but worth a fix or at least a tracked follow-up — e.g. only apply the "preserve prior" fallback when the whole `artifacts` key came back nil, not per-field.

internal/kibana/alertingrule/schema.go
```diff
@@ -206,13 +208,29 @@ func getSchema(_ context.Context) schema.Schema {
 				MarkdownDescription: artifactsDescription,
 				Optional:            true,
 				Computed:            true,
-				Validators: []validator.Object{
-					objectvalidator.AlsoRequires(path.MatchRelative().AtName(attrInvestigationGuide)),
-				},
+				// At least one of investigation_guide / dashboards must be set when
+				// artifacts is present; this is enforced in ValidateConfig because
+				// object validators cannot express "at least one nested attribute".
```
tobio:
```go
// At least one of investigation_guide / dashboards must be set when
// artifacts is present; this is enforced in ValidateConfig because
// object validators cannot express "at least one nested attribute".
```

This isn't quite right — `listvalidator`/`objectvalidator.AtLeastOneOf` already exists and handles this declaratively, including the same unknown-value deferral this comment gives as the reason to go custom (it returns early with no error if any referenced path is unknown). This file already uses the identical idiom one level down for `content`/`content_path`:

```go
stringvalidator.ExactlyOneOf(
    path.MatchRelative().AtParent().AtName(attrContent),
    path.MatchRelative().AtParent().AtName(attrContentPath),
),
```

The same pattern here would be:

```go
attrDashboards: schema.ListNestedAttribute{
    Validators: []validator.List{
        listvalidator.AtLeastOneOf(
            path.MatchRelative().AtParent().AtName(attrInvestigationGuide),
            path.MatchRelative().AtParent().AtName(attrDashboards),
        ),
        listvalidator.SizeAtLeast(1), // rejects `dashboards = []`; null/unknown are skipped
    },
    ...
},
```

`listvalidator.SizeAtLeast(1)` covers the "reject an explicitly empty `dashboards = []`" case `validateArtifactsNotEmpty` was added for — it skips null and unknown values and only errors on a known, too-short list, so it doesn't reintroduce false positives for computed/unknown dashboards. Combined, these two declarative validators reproduce everything `validateArtifactsNotEmpty` does by hand, and that custom `ValidateConfig` function (and its dedicated test cases) could likely be dropped.

internal/kibana/alertingrule/models.go
kroustou: Good catch. Dropped the `blob != ""` guard so a configured guide is always sent, including `content = ""`. Added a unit test for the empty-blob write path.

internal/kibana/alertingrule/models.go
kroustou: Fixed. We only preserve prior artifacts when GET omits the whole `artifacts` key (pre-9.5.0). If the API returns artifacts, omitted siblings are cleared so an out-of-band delete of just the guide or just dashboards shows up on refresh. `ConvertResponseToModel` now keeps a non-nil artifacts object whenever the key is present, so `{}` is distinct from “key omitted.”

internal/kibana/alertingrule/schema.go
kroustou: Agreed — that comment was wrong. Replaced the custom `ValidateConfig` check with `listvalidator.AtLeastOneOf` / `objectvalidator.AtLeastOneOf` and `listvalidator.SizeAtLeast(1)` on `dashboards`. Same unknown-value deferral, and `dashboards = []` is still rejected.

docs/resources/kibana_alerting_rule.md
```diff
@@ -63,7 +63,7 @@ resource "elasticstack_kibana_alerting_rule" "example" {
...
+- `artifacts` (Attributes) Linked assets for the alerting rule: an investigation guide and/or references to related dashboards, so that runbooks and context are co-located with the alert configuration as code. At least one of `investigation_guide` or `dashboards` must be set. Requires Elastic Stack 9.1 or higher to write; on stacks older than 9.5.0 the Kibana GET API does not return artifacts (elastic/kibana#247279), so `terraform import` will not populate this block and external changes to artifacts made outside Terraform will not be detected by refresh. Omitting this block does not clear artifacts already stored on the rule in Kibana. (see [below for nested schema](#nestedatt--artifacts))
```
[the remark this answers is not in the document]
kroustou: Updated — artifacts is described as an attribute throughout.

internal/kibana/alertingrule/validate.go
```diff
@@ -443,6 +444,31 @@ const frequencyExclusivityDetail = "Rule-level notify_when and throttle cannot b
...
+	// An unknown value (e.g. a dashboards list built from a variable/for-expression
+	// or a value sourced from another resource) is treated as possibly-set so we do
+	// not emit a false positive during validation.
+	if am.InvestigationGuide.IsNull() && am.Dashboards.IsNull() {
+		diags.AddAttributeError(
+			path.Root("artifacts"),
+			"Empty artifacts",
+			"artifacts must set at least one of investigation_guide or dashboards.",
+		)
+	}
```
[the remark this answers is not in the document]
kroustou: Addressed — empty dashboards = [] is rejected via SizeAtLeast(1).

internal/kibana/alertingrule/descriptions/artifacts.md
```diff
@@ -1 +1 @@
...
+Linked assets for the alerting rule: an investigation guide and/or references to related dashboards, so that runbooks and context are co-located with the alert configuration as code. At least one of `investigation_guide` or `dashboards` must be set. Requires Elastic Stack 9.1 or higher to write; on stacks older than 9.5.0 the Kibana GET API does not return artifacts (elastic/kibana#247279), so `terraform import` will not populate this block and external changes to artifacts made outside Terraform will not be detected by refresh. Omitting this block does not clear artifacts already stored on the rule in Kibana.
```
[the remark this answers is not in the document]
kroustou: Updated the description to call it an attribute.

tobio (approved):
Thanks for this!

PR #4581, opened 2026-08-17
fix(elasticsearch-index): compare dynamic templates semantically

### LINKED ISSUE #4519: [Bug] index_mappings fails with inconsistent result after applying existing index mappings
**Describe the bug**
elasticstack_elasticsearch_index_mappings fails with Provider produced inconsistent result after apply when updating the mapping of an existing index that already contains dynamic_templates and nested object fields with a custom analyzer. The PUT to Elasticsearch itself succeeds (confirmed via GET <index>/_mapping after the failed apply — the mapping was actually applied), but Terraform still reports the apply as an error because the .mappings value returned after apply does not match the planned value byte-for-byte.

**To Reproduce**
1. TF configuration used:
hcl
resource "elasticstack_elasticsearch_index_mappings" "example" {
  index    = "my-index-20260812"
  mappings = jsonencode({
    dynamic = true
    dynamic_templates = [
      {
        text_ja_example = {
          mapping    = { type = "text", analyzer = "kuro" }
          path_match = "hoge.example_field.freetext"
        }
      }
    ]

1. (The target index already exists and already has a superset of this mapping applied via a matching index template.)
2. terraform apply
3. Error in output:
When applying changes to elasticstack_elasticsearch_index_mappings.example, provider "provider[\"registry.terraform.io/elastic/elasticstack\"]" produced an unexpected new value: .mappings: was cty.StringVal("{\"dynamic\":true,\"dynamic_templates\":[{\"text_ja_example\":{\"mapping\":{\"analyzer\":\"kuro\",\"type\":\"text\"},\"path_match\":\"hoge.example_field.freetext\"}}], ...

**Expected behavior**
terraform apply completes successfully, and the stored state's .mappings reflects a value semantically equal to the plan (per the existing MappingsValue/StringSemanticEquals normalization already implemented for this resource), without raising the provider-consistency error.

**Debug output**
Unable to provide TF_LOG=trace output for this report — the mapping payload contains internal field/schema names that we cannot disclose publicly. 

**Screenshots**
NA

- OS: Linux (HCP Terraform remote execution)
- Terraform Version: 1.15.8
- Provider version: v0.16.3
- Elasticsearch Version: 8.14.0

**Additional context**
- elasticstack_elasticsearch_index_mappings was added recently in #3121 (v0.16.0).
- The provider already has normalization logic for this exact class of drift in internal/elasticsearch/index/mappings_value.go (MappingsValue.StringSemanticEquals, normalizeMappings), but it does not appear to cover whatever difference Elasticsearch introduces in our case (possibly related to dynamic_templates ordering or additional keys Elasticsearch injects that aren't yet normalized).
- A related but distinct bug in the sibling elasticstack_elasticsearch_index resource's mapping-update decision logic was just fixed in #4507 (merged 2026-08-12, not yet released as of v0.16.3) — this suggests the mapping-comparison code path for both resources is under active development.
- As a workaround, we replaced this resource with a terraform_data + local-exec provisioner that PUTs the mapping directly via curl, bypassing the provider's comparison entirely.

shojiiii:
## Changelog
Customer impact: fix
Summary: Prevent false inconsistent-result errors when index mappings use template-injected dynamic templates.

## Summary

  Fixes false `Provider produced inconsistent result after apply` errors for `elasticstack_elasticsearch_index_mappings` when
  Elasticsearch returns template-injected `dynamic_templates`.

  - Compare `dynamic_templates` by template name instead of JSON array order.
  - Allow additional API-returned templates while requiring every user-declared template to be present with an equivalent definition.
  - Preserve update behavior for missing or changed user-owned templates.
  - Add unit and acceptance regression coverage.

  ## Root cause

  `MappingsSemanticallyEqual` handled `properties` as a user-owned subset of API state, but compared `dynamic_templates` as an exact
  array. Elasticsearch can return additional templates or a different order, making an otherwise compatible mapping fail Terraform's
  post-apply consistency check.

  ## Validation

  - `go test ./internal/elasticsearch/index`
  - `TF_ACC=1 go test -v -run '^TestAccResourceIndexMappings_dynamicTemplatesFromIndexTemplate$' ./internal/elasticsearch/index/
  indexmappings`
  - `make build`

  Fixes #4519

### DIFF
--- a/internal/elasticsearch/index/indexmappings/acc_test.go
+++ b/internal/elasticsearch/index/indexmappings/acc_test.go
@@ -193,6 +193,55 @@ func TestAccResourceIndexMappings_allTopLevelKeys(t *testing.T) {
 	})
 }
 
+func TestAccResourceIndexMappings_dynamicTemplatesFromIndexTemplate(t *testing.T) {
+	indexName := sdkacctest.RandStringFromCharSet(22, sdkacctest.CharSetAlphaNum)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:     func() { acctest.PreCheck(t) },
+		CheckDestroy: checkResourceIndexMappingsDestroy,
+		Steps: []resource.TestStep{
+			{
+				ProtoV6ProviderFactories: acctest.Providers,
+				ConfigDirectory:          acctest.NamedTestCaseDirectory("apply"),
+				ConfigVariables: config.Variables{
+					"index_name": config.StringVariable(indexName),
+				},
+				Check: resource.ComposeTestCheckFunc(
+					checkStateMappingsDynamicTemplates(1),
+				),
+			},
+			{
+				ProtoV6ProviderFactories: acctest.Providers,
+				ConfigDirectory:          acctest.NamedTestCaseDirectory("apply"),
+				ConfigVariables: config.Variables{
+					"index_name": config.StringVariable(indexName),
+				},
+				PreConfig: func() {
+					setDynamicTemplatesOutOfBand(t, indexName, []any{
+						map[string]any{
+							"template_default": map[string]any{
+								"match_mapping_type": "string",
+								"mapping":            map[string]any{"type": "keyword"},
+							},
+						},
+						map[string]any{
+							"text_ja_example": map[string]any{
+								"path_match": "hoge.example_field.freetext",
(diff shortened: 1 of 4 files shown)


tobio (approved):
Thanks for taking the time to get this fixed up!

PR #4615, opened 2026-08-20
Propose fleet-managed-integration-ga OpenSpec change

rubenruizdegauna:
## Summary

OpenSpec change proposal (no implementation yet) to promote `elasticstack_fleet_managed_integration` out of the provider's experimental gate.

The Kibana Fleet `managed_integrations` API is generally available since Elastic Stack 9.5 — exactly the resource's existing minimum-version floor — so the experimental gate now only hides the resource (no docs page, opt-in env var, "may change" wording) without protecting anyone: the stable resource can never reach a stack where the API is still in preview.

## What the change covers

- Move the resource from `experimentalResources()` to the stable `resources()` list (move, not copy — `Resources()` appends the experimental slice, so a duplicate entry would register the type twice).
- Rewrite the experimental sentence in the schema `MarkdownDescription` (keeping the Kibana 9.5.0 minimum and the ECH/Serverless-only constraint) and drop the `(experimental API)` suffix from the version-gate diagnostic.
- Retarget the experimental-gate tests to `elasticstack_kibana_tag` so the gate mechanism keeps failure-detecting coverage, and rename the test file to `plugin_framework_registration_test.go`.
- Generate the previously missing `docs/resources/fleet_managed_integration.md` page and remove the obsolete env-var instruction from the examples README.
- Delta spec: requirement **Resource type and registration** is replaced by **Resource type and stable registration** (REMOVED + ADDED, since a scenario is dropped and the requirement is renamed). The Kibana 9.5.0 version-gate requirement is deliberately untouched.

`openspec validate --all` passes.

## Changelog
Customer impact: none

This PR only adds an OpenSpec change proposal under `openspec/changes/`; the implementation (and its enhancement changelog entry) will come in a follow-up PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/openspec/changes/fleet-managed-integration-ga/.openspec.yaml
+++ b/openspec/changes/fleet-managed-integration-ga/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-08-20
--- a/openspec/changes/fleet-managed-integration-ga/design.md
+++ b/openspec/changes/fleet-managed-integration-ga/design.md
@@ -1,0 +1,48 @@
+## Context
+
+See `proposal.md` — Why. Requirements are in `specs/fleet-managed-integration/spec.md`.
+
+Three mechanics shape the approach:
+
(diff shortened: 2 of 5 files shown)


tobio (approved):
The tests in `plugin_framework_experimental_test.go` aren't really valuable. Just deleting those tests entirely would also be ok rather than the palaver in Task 3. 

I'll open a follow up issue to make those tests actually relevant if not though.

rubenruizdegauna:
> The tests in `plugin_framework_experimental_test.go` aren't really valuable. Just deleting those tests entirely would also be ok rather than the palaver in Task 3.
> 
> I'll open a follow up issue to make those tests actually relevant if not though.

Thanks for the super quick review @tobio !  I'll take a look to those tests

PR #4988, opened 2026-09-21
feat(fleet): add elasticstack_fleet_space_settings resource

### LINKED ISSUE #4980: [Feature] Add elasticstack_fleet_space_settings resource

Fleet's per-space settings cannot be managed with the provider. In a setup with one Kibana
space per team, `allowed_namespace_prefixes` restricts which data stream namespaces a space
may write to. It is a security control, so having it as the only manually maintained part
of an otherwise declarative Fleet setup means no review and no drift detection on it.

`GET` / `PUT /api/fleet/space_settings`

- the endpoint takes no space parameter in the body — the space comes from the request
  path (`/s/<space>/api/fleet/space_settings`)
- requires the `fleet-settings-all` route privilege
- request body: `allowed_namespace_prefixes` (array of string, max 10 elements per write)
- response: `allowed_namespace_prefixes` (max 100 elements) and `managed_by`

API reference:
https://www.elastic.co/docs/api/doc/kibana/v9/operation/operation-put-fleet-space-settings

A resource that selects the space explicitly, since the API expresses it through the path:

```hcl
resource "elasticstack_fleet_space_settings" "team_a" {
  space_id                   = "team-a"
  allowed_namespace_prefixes = ["team_a"]
}
```

`managed_by` as a read-only attribute. Destroy would presumably write an empty list rather
than delete anything. A matching data source would help for reading the state of spaces
that are not managed by Terraform.

- Applying it by hand per space. Does not scale to one space per team and leaves the
  setting outside review.
- Wrapping the API call in `terraform_data` with a local-exec provisioner. Works, but
  produces no state and no drift detection, which defeats the purpose for a security
  control.

**Additional context**

This became more relevant once Fleet space awareness turned into the default for new
deployments in 9.1 (https://www.elastic.co/docs/deploy-manage/manage-spaces-fleet).
Setups using one space per team now have a per-space Fleet configuration surface, and this
appears to be the only part of it without provider coverage.

gianvetter:
## Changelog
Customer impact: enhancement
Summary: Add the `elasticstack_fleet_space_settings` resource to manage Fleet's per-space settings (`allowed_namespace_prefixes`). Requires Elastic Stack 9.1.0 or newer.

## Summary

Adds a resource for Fleet's per-space settings (`GET`/`PUT /api/fleet/space_settings`), so the namespace restriction of a Kibana space can be reviewed as code and drift-detected.

Closes #4980

- `space_id` (required, replaces on change), `allowed_namespace_prefixes` (required list, max 10 unique entries, empty list allowed), computed `managed_by`; `id` equals `space_id`
- Import by space ID (including `default`)
- Destroy resets the prefixes to an empty list (this lifts the restriction); the Kibana space is kept
- Requires Elastic Stack 9.1.0+; optional `kibana_connection` override like other Fleet resources
- Includes the OpenSpec change (`openspec/changes/add-fleet-space-settings-resource`: proposal, spec, design, tasks)

## Notes for reviewers

- Kibana 9.4 accepts settings for a space that does not exist, so a mistyped `space_id` silently creates orphan settings. The resource neither creates nor verifies the space; this is documented in the resource description and the spec/design.
- Only one `elasticstack_fleet_space_settings` resource should exist per space; multiple resources targeting the same space overwrite each other.
- The manual `[Unreleased]` entry in `CHANGELOG.md` has no PR link yet; drop it if the changelog is generated from this PR description.

## Testing

- Unit tests for the client wrapper, schema validation, model mapping, read/delete paths, version gate, scoped connection and error surfacing
- Acceptance tests (create/update, import, default-space import, drift, destroy reset, `kibana_connection`) pass against a local Elastic Stack 9.4.0
- `make build`, `make golangci-lint` and `openspec validate --strict` pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/fleet/spacesettings/schema.go
```diff
@@ -0,0 +1,72 @@
...
+				},
+			},
+			"space_id": schema.StringAttribute{
+				Description: "The ID of the space whose Fleet settings are managed. Changing it forces a new resource.",
+				Required:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"allowed_namespace_prefixes": schema.ListAttribute{
```
tobio, line 56:
```suggestion
			"allowed_namespace_prefixes": schema.SetAttribute{
```

This should almost certainly be a set rather than an ordered list.

provider/provider_test.go
```diff
@@ -47,6 +49,20 @@ func TestProvider(t *testing.T) {
 	}
 }
 
+func TestProviderRegistersFleetSpaceSettingsResource(t *testing.T) {
```
tobio, line 52: Can we drop this test, it's not meaningfully adding anything.

